PR #1, opened 2024-01-18
Remove "white" flavour of hero image cover page for CDC

pcraig3:
This PR removes the "white" theme of the hero image title page. You can still have the white theme for your NOFO, but the cover page will look the same either way. 

Note that it doesn't do anything with the HRSA or ACF theme files, those both need to be updated to make sure the cover page actually looks good in this new format.

| before | after |
|--------|-------|
| A "white" theme that I improvised       |   Now uses the blue theme that's in the mocks     |
|   Screenshot 2024-01-18 at 2 15 48 PM    |    Screenshot 2024-01-18 at 2 15 08 PM |


admoorgit:
Looks good to me! I SHALL HIT BUTTON

PR #2, opened 2024-01-19
Fix Icons + adjust header sizing + new CDC logo

pcraig3:
There are lots of individual changes in this PR, so I will put the list here which I grabbed from the CHANGELOG.

There's a lot in here, but arguably the biggest thing I changed was replacing the previous SVGs with new ones. I had been using one SVG for different themes and locations and then using CSS to recolour them, but this resulted in super pixelated icons in the actual PDF. In the end, I created different svg files for different colours and then rejigged how icons are returned. Result: icons in the final rendered PDF looks way better.

### Added

- New combined CDC and HHS logo
  - Dropped the footer logo
- Callout box for the in-PDF nav instructions
  - Add PDF svg icon
- Alternate callout-box style
- Icons are revamped
  - Now all svgs are in the colour they will end up being displayed as

### Changed

- Fixed the border above tables, reduced spacing underneath
- Use NOFO title for PDF title
- Change body heading sizes to match typescale
  - Larger for h3, h4, h5
- CDC white theme uses vibrant blue everywhere in place of dark blue
- Even in the white theme, the "hero" cover page uses blue background and white text
- Use 50% width for section 4 (other than tables)
  - Same with "Before you begin" page
- Tables with 3 cols are "large", and have grey top bar
  - Before it was only for tables with captions
- Clean up header nav colours
- Standardize heading colours
- Underline page numbers in table of contents
- Increase space after lists
- Reduce footer padding: move it on top to separate footer from page content
- Prefer filling in column over balancing them

### Fixed

- Show arrow svgs on bolded li elements in table cells
- Contacts and support section is not a step
- Fix: Headings have less top spacing if they follow another heading


admoorgit:
YAY! Looks good.

PR #11, opened 2024-10-24
Markdown attribute ids should not be broken links

pcraig3:
## Summary

The NOFO Builder allows users to use curly braces in markdown to add attributes to elements using [attribute lists](https://python-markdown.github.io/extensions/attr_list/), like so:

```
This is a paragraph.
{ .class-for-the-paragraph #id-for-the-paragraph }
```

However, our 'broken link' checking function didn't take these into account, so manually-added ids were showing up as 'broken'. 

This PR fixes that and adds a test for it.

### Why are we doing this?

It's a bug 🐞 

### How to test this

- Find a NOFO
- Manually add an attribute ID 
- Manually add a link to that ID
- It should not show up as a 'broken link' in the alert box


pcraig3:
Merging this in because it is a bug and should therefore just be merged quickly.

PR #26, opened 2024-10-30
Light blue table rows + fix asterisks

pcraig3:
## Summary

This PR does 3 things:

1. Adds in a new background colour for table cells
2. Fixes CSS for the "Adam 👀" status (self-explanatory)
3. Fixes the asterisk problem we are seeing in table cells

#### 1. Adds new background colour for table cells
This PR adds in a new background colour for tables so that we can have secondary header-like rows. 

| adding a new background colour |
|-------|
|    Screenshot 2024-10-30 at 11 44 28 AM   |
| Adds the new light blue colour |

#### 3. Fixes the asterisk problem we are seeing in table cells

Sometimes we see asterisks cause text to be italicized inside of tables, when what we want is a literal asterisk. 

Here's an example.

| bad | good |
|--------|-------|
| italics are NOT intentional | literal asterisks showing up (*) |
| Screenshot 2024-10-30 at 5 21 26 PM  |   Screenshot 2024-10-30 at 5 21 18 PM |

What I realized after monkeying around with them was:

Sometimes we have regular asterisks and it's no problem:

```
* this is a list item
* this is a list item with an asterisk*
* this is a list item also
```

That works.

Other times, we have markdown tables with an asterisk.

```
| before | after |
|--------|-------|
| before* | after* |
```

That works.

Other times, we have HTML tables with HTML inside of the cells.

```
<table><tr><th>Header 1</th><th>Header 2</th></tr><tr><td><ul><li>Item 1*</li><li>Item 2</li></ul></td><td>Cell 2</td></tr></table>
```

That works.

FINALLY, we can have markdown tables with HTML in the table cells.

```
| before | after |
|--------|-------|
| <ul><li>Item 1*</li><li>Item 2</li></ul> | Cell 2 |
```

THAT is the one that causes problems for us.

So, if we intervene in the markdown conversion, specifically for lists in table cells, then we can convert only those astersiks into HTML entities.

The result is that we don't have these broad rulesets that cause weird
side effects, and everything will be solved from now on.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ The format is based on Keep a Changelog, and this project adheres to Semantic Ve
 
 ### Fixed
 
+- Convert literal asterisks to `&ast;` inside of HTML lists in table cells
 - More left padding on callout box lists that are NOT in the right hand column
 - More specific CSS selector for application table divs
 - Application list checkbox cells that are not sublists have bottom borders
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ The format is based on Keep a Changelog, and this project adheres to Semantic Ve
 - More left padding on callout box lists that are NOT in the right hand column
 - More specific CSS selector for application table divs
 - Application list checkbox cells that are not sublists have bottom borders
+- "Adam 👀" status needs yellow highlight background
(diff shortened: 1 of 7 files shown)


pcraig3:
Let's get this in!

PR #27, opened 2024-11-02
Add 3 audit events + Fix audit logs

pcraig3:
## Summary

This PR does 4 things:
1. Creates 3 new audit events
2. Fixes the old audit events that were in the wrong format
3. Fix an alt text bug
4. Remove "Default" style map

#### 1. Creates 3 new audit events

This PR creates 3 audit events for NOFO lifecycle stuff:

1. When a NOFO is imported
2. When a NOFO is reimported
3. When a NOFO is printed

Looking through our audit logs currently, it's very hard to understand just from the 'Update' audits when a NOFO has been re-imported, and it's not super easy to tell when they have been imported, so there is a gap here. 

Now we can specifically track imports, re-imports, and print actions.

#### 2. Fixes the old audit events in the wrong format

There is a migration that will fix the previous 'print' audit events that were saved as strings rather than JSON. Added a one-time migration that fixes them all. 🤞


pcraig3:
Going to merge this one as it is a bit sensitive with the DB migration.

PR #28, opened 2024-11-04
Change name of cover, icon style, and review status

pcraig3:
This PR updates various string names for NOFO attributes.

- `NOFO.cover`: 
  -  "Hero (large) image" -> "Full coverage with image"
  -  "Small image" -> "Standard image"
  -  "No image, large text" -> "Text only"
- `NOFO.status`
  - "Adam 👀" -> "Ready for QA"
- `NOFO. icon_style`
  - "(Outlined)" -> "(Standard)"

The only actual id attribute that changed is "status", which also meant updating some CSS (tightly coupled!).


pcraig3:
Merging because Adam is off for a couple days!

PR #29, opened 2024-11-05
Admin external link checker

pcraig3:
## Summary

This PR adds an external link checker for superadmin users. 

Sometimes when we use the external link checker, we see that some URL statuses fail even though it is possible to visit them in the browser. When this happens, it is generally not obvious what the issue is.

When I've investigated this in the past, I have been adding code manually to print the entire response and/or save to a file, which has worked but is not a good way to do things. This PR creates an actual page for us to do this in a nicer way, available only to superadmins. I don't expect we will need to use this very often but it is also pretty lightweight and will save me some time.

| page with successful URL | page with bad URL  |
|--------|--------|
|   Shows title, status code, and response if needed.     |   Shows the Error we are getting from requests.  No status code because this is an exception.   |
| Screenshot 2024-11-05 at 10 16 06 AM | Screenshot 2024-11-05 at 10 16 20 AM |

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ The format is based on Keep a Changelog, and this project adheres to Semantic Ve
 
 ### Added
 
+- Add new view to verify individual external URL responses
 - Add button to copy list of broken links on nofo_edit page
 - Add new coaches and designers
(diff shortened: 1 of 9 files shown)


pcraig3:
Merging in!

Issue #85, opened 2024-12-04
Pull data on print counts


pcraig3:
We have been tracking "print" counts in audit events since October, although the early data doesn't differentiate between "test" and "live" prints. That only came a few weeks later. 

No matter! We have some data, and that's something.

Let's pull some data to find out:

- how many times NOFOs are printed
- total time between importing a NOFO and final print in "live" mode.

PR #86, opened 2024-12-04
Get print counts

pcraig3:
## Summary

This PR creates a new command we can use to pull data about print counts, and runs a migration to clean up our audit events.

1. Pull data about print counts
2. Run a migration to clean up some old data

### 1. Pull data on print counts

We have been tracking "print" counts in audit events since October, although the early data doesn't differentiate between "test" and "live" prints. That only came a few weeks later. No matter! We have some data, and that's something.

Let's pull some data to find out:

- how many times NOFOs are printed
- total time between importing a NOFO and final print in "live" mode.

### 2. Run a migration to convert early print audit events to JSON

I created an audit event function that adds a CRUD event every time a NOFO is printed. 

Early on, these were strings, like this: `PRINT: NOFO id #351 at 2024-10-08 14:30:23.072831+00:00`

This is not valid JSON, which initially I wasn't aware that I should be doing. 

I added a [migration on Nov 2](https://github.com/HHS/simpler-grants-pdf-builder/commit/305ae39329b9a24d43fae6425a4a4550b652508c) to fix some of these, but because I changed the string format, it didn't catch all of them.

This is a cleanup migration that finds the earliest version of these string audit events and converts them all.

Unfortunately, we don't know if these early audit events were "test" or "live", so we are defaulting them all to "live".


pcraig3:
This PR closes #85

Issue #25, opened 2024-10-30
Better error message for long headings


pcraig3:
When you import a Word document with a really long (paragraph-length) block tagged as a heading, it throws up an exception related to the line length, which originates in the [in the model definition for a`Section`](https://github.com/HHS/simpler-grants-pdf-builder/blob/main/bloom_nofos/nofos/models.py#L315).

<figure>
  <img src="https://github.com/user-attachments/assets/e1343c2d-ddfa-4d32-9c8c-a80ed4f8efbb"  />
  <figcaption><em>This paragraph in the orange box is styled as a Heading 3</em></figcaption>
</figure>

<br>
<br>

This is actually desired behaviour, since we don't ever see headings that are this large. In (probably) every case, these are paragraph elements that have been mistagged as headings. 

The problem is that the error message we return to the user doesn't explain the issue very well. It talks about the character length but not about what's the root issue.

![Image](https://github.com/user-attachments/assets/44ebff59-5255-4f9c-b5d4-566e74627cb4)

This story is about adding an error message that communicates to users what is happening. 

What is happening is that they have mistagged paragraph text as heading text. We should tell them some version of this ('this document can't be imported because'), and also indicate where in the document this issue is (either referring to the text itself or the previous heading). 

Right now the 400-level error being thrown is a general catch-all error route. Check out this commit for the origin of this: https://github.com/HHS/simpler-grants-pdf-builder/commit/4981aa47f5f3956994c8a8fa7d6a0386dfe6c66d#diff-9438ebabdf1c1d43f10f6eafb9bdcb6459283b95943993031f9660187da51b0fR10

pcraig3:
Let me know if you need an example document for testing this.

pcraig3:
This is closed since #97 went in.

PR #97, opened 2024-12-10
Heading length error messaging

jtmst:
Original issue: https://github.com/HHS/simpler-grants-pdf-builder/issues/25

This adds more specific error messaging for cases where validation errors are thrown for long headings. New text:

Screenshots: 
image

QA Steps:
1. Navigate to `http://localhost:8000/nofos/import`
2. Import a NOFO that has a heading that exceedes the character limit of 511
3. Verify that error message message renders as expected with a status code of 422

### DIFF
--- a/bloom_nofos/bloom_nofos/templates/400.html
+++ b/bloom_nofos/bloom_nofos/templates/400.html
@@ -1,12 +1,12 @@
 {% extends 'base.html' %}
 
 {% block title %}
-	{% if status %}{{ status }}{% else %}400{% endif %} bruv, oops — Bad request
+	{% if status %}{{ status }}{% else %}400{% endif %} - Bad Request
 {% endblock %}
 
 
 {% block content %}
-	<h1 class="font-heading-xl margin-y-0">{% if status %}{{ status }}{% else %}400{% endif %} bruv, oops</h1>
+	<h1 class="font-heading-xl margin-y-0">{% if status %}{{ status }}{% else %}400{% endif %} Error</h1>
 	{% if error_message %}
 		<p><strong>{{ error_message }}</strong></p>
 	{% elif error_message_html %}
--- a/bloom_nofos/bloom_nofos/templates/404.html
+++ b/bloom_nofos/bloom_nofos/templates/404.html
@@ -1,14 +1,14 @@
 {% extends 'base.html' %}
 
 {% block title %}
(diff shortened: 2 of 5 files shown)


pcraig3 (changes requested):
This looks good!

I have 2 requested changes for you.

### make the app less fun

The first is that I am using cool uk slang words when errors come up, but we should 100% definitely stop doing this:
- [400: bruv](https://github.com/HHS/simpler-grants-pdf-builder/blob/main/bloom_nofos/bloom_nofos/templates/400.html#L4)
- [404: dawg](https://github.com/HHS/simpler-grants-pdf-builder/blob/main/bloom_nofos/bloom_nofos/templates/404.html#L4)
- [500: fam](https://github.com/HHS/simpler-grants-pdf-builder/blob/main/bloom_nofos/bloom_nofos/templates/500.html#L4)

Easiest thing is just remove those words (and the "dang/sorry", etc) and keep things professional. 

I know that this is not part of the issue description, but it should be a quick fix to modify those templates.

### better signposting for this specific issue

I feel like we should just white-glove this issue. 

When you import a NOFO that fails to validate, it actually gets partially created. The subsection that has the title that is too long fails to be created but the one beforehand _is_ created. 

At this point we know 3 things:
- the NOFO 
- the last compliant subsection
- the user who uploaded it

So let's:

- link to the NOFO (the latest one we have)
- pull in the title of the last subsection in the NOFO (also the most recently created subsection, most likely)
- change the group to be the user's group

By default, all NOFOs are given the "bloom" group, but we should actually override that here and add the user's group. The reason for this is that if a HRSA user runs into this error, their NOFO is not actually a HRSA NOFO yet, so they can't see it. But they should be able to take a look even though it is busted.

So we should add another line that says something like:

```
Your document ("title of nofo" + link) contains a heading that is too long. Headings have a character limit of 511 characters.

This usually means that a paragraph has been incorrectly styled as a heading. 
The last valid heading was "name of heading" (no link)."
```

Also, I know this is bad practice, but can you leave a TODO in the codebase that says something like `# TODO: dynamically pull in heading character limit from models.py _or_ error message`.

We should do it, but since we are going to be updating the schema soon, we can wait until then.

pcraig3 (approved):
Looks great! Way better than before. 

Feel free to rebase and merge!

Screenshot 2024-12-11 at 4 01 27 PM

PR #112, opened 2024-12-11
Add script to find lengths of section.name and subsection.name values

pcraig3:
## Summary

I don't know if we really need this for much but it's not really hurting anything.

Added a new Django command to pull down sections and subsections from non-archived NOFOs and to export them to a spreadsheet. 

It lives in the folder with our other commands. 

Run it with: 

```
poetry run python manage.py section_lengths --all
```

Or 

```
poetry run python manage.py section_lengths --nofo-id 1
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ The format is based on Keep a Changelog, and this project adheres to Semantic Ve
 ### Added
 
 - Add new command for downloading NOFOs with published times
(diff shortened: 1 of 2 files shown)


jtmst (approved):

jtmst (commented):
The section/subsection loops could be a concern if we had millions of them, but thats not an issue at the moment

Issue #99, opened 2024-12-10
Spike out admin API with django ninja


pcraig3:
I created an JSON import/export story (#24) which works as well as it needs to.  The problem I was having is that people were finding NOFO issues in prod that I wasn’t able to replicate locally unless I had the original Word doc. By adding this import/export route, it's now possible to export a NOFO as pure data and then import it into another environment. 

However, it’s really not an API.

It doesn’t announce itself as such and it doesn’t use a POST request. Also, it includes a bunch of validation rules in the view, which we should really keep in `models.py`. 

Admittedly this is my putting my thumb on the scale here, but I have a preference for [Django Ninja](https://django-ninja.dev/). I’ve used it on past projects and I found it to be pretty easy to set up and use.

However, I am not sure that it will be able to do everything we need to yet. 

The major factors we need to investigate are:

- models.py validation is preserved in routes
    - Meaningful error messages are returned
- Admin-only routes (probably this means a bearer token, rather than Django login creds) 
- General observations on ease of use/maintainability

This initial spike should be centred around replicating the import/export functionality (see #44). 

That is, as an admin user (can be someone with a bearer token), I should be able to export a NOFO as JSON and then POST a JSON request that imports a NOFO as well.

Issue #101, opened 2024-12-10
Admin API routes


pcraig3:
This is the “good copy” of the admin routes we want:

1. Admin GET route: superusers can use this to export NOFO data by id
Admin POST route: superusers can use this to import new NOFOs given a compliant JSON file

This may be completed by the spike (#99) or we may find we need a new solution, or we just need some extra work to ‘prodify’ the work during the spike.

This story also means we should retire the import/export views we have right now (#44) — ALTHOUGH, it might be nice to keep that import screen around and have it do the POST request for us. If that’s a no-go though, no problem, we can remove it. This is a pro user feature anyway.

pcraig3:
Decided to kill the import screen for reasons of dogfood and simplicity.

Issue #100, opened 2024-12-10
Follow up: Spike out admin API


pcraig3:
This is a placeholder ticket that follows #99 

If the spike was successful, we can continue. If not, we fill out this ticket — either we try a new API lib or we roll our own. After all, there are only 3 endpoints here.

Issue #75, opened 2024-12-02
Define initial schema for NOFO validation


pcraig3:
We will try to import pretty much any Word Doc and assume it's a NOFO.

Early on, this was super helpful because we didn't know what these documents looked like. 

However, we now have stronger opinions about what makes a NOFO (they have at least 7 sections, for example), so we should introduce some validation with meaningful validation messaging to give people some feedback when they upload a document that is invalid. 

Currently, uploading a 'bad' document will either:

- fail, usually with a cryptic error message
- succeed, but then you don't get what you were expecting

pcraig3:
Talked to @admoorgit about this this week. 

Here's what we landed on:

- [ ] Must have OpDiv
- [ ] Must have _either_ title or number
- [ ] Must have at least 1 section
- [ ] Every section must have at least 1 subsection

admoorgit:
@pcraig3 allegedly GrantSolutions has a 250 character limit on NOFO Titles. It's way too long, but it's a start.

pcraig3:
⚠ This comment is pending approval from @admoorgit ⚠ 

After some Big Data ™ analysis, we have landed on 400 chars for titles. Since the html_id is a derived title, I think it makes most sense to take the limit off html_id and just make sure `subsection.name` both have a 400 char limit.

GS has a 250 char limit on nofo titles, and we've seen that section titles are also way shorter than that, so I let's say 250 chars for both.

Which means this is what we end up with:

- [ ] Must have OpDiv
- [ ] Must have _either_ title or number
- [ ] Nofo must have at least 1 section
- [ ] Every section must have at least 1 subsection
- [ ] `nofo.title` max-length: 250 chars (251 is too long)
- [ ] `section.name` max-length: 250 chars (251 is too long)
- [ ] `subsection.name` max-length: 400 chars (401 is too long)

pcraig3:
<img width="419" alt="Image" src="https://github.com/user-attachments/assets/ed016fc7-9bc1-4848-835b-fe8d5a7b61a8" />

pcraig3:
Since #138 went in, this is officially closed!

PR #117, opened 2024-12-13
Admin api ninja spike

jtmst:
Issue: https://github.com/HHS/simpler-grants-pdf-builder/issues/99

# Ninja API Spike Findings

## Overview
- Created import/export endpoints for NOFOs
- Created schemas using Ninja's ModelSchema for automatic model validation
- Implemented mock bearer token authentication
- removes JSON view endpoints and tests for import/export
- adjusts admin fields to reflect
- adds ENV var for api_token

## Key Findings
1. *Model Validation
   - Django model constraints are preserved (e.g., max_length=511 for html_id)
   - Validation errors are properly caught and returned
   - Complex model relationships (NOFO -> Sections -> Subsections) are handled correctly

2. Error Handling
   - Meaningful error messages with specific details
   - Proper error structure using ErrorSchema
   - Example error response:
   ```json
   {
     "message": "Validation error",
     "details": {
       "html_id": ["Ensure this value has at most 511 characters (it has 4081)."]
     }
   }
   ```

3. **Authentication**
   - Bearer token authentication implemented
   - Routes protected behind authentication
   - Currently using test implementation (accepts "secret" prefix)

## Benefits
1. Schema Management
   - ModelSchema reduces code duplication
   - Automatically stays in sync with model changes
   - Clear documentation via OpenAPI/Swagger UI at `/api/docs`

2. Maintainability
   - Clear separation of concerns (schemas.py, api.py)
   - Less overhead and simple setup
   - Type hints and validation built-in

## Recommendation
1. Overall Ninja seems pretty rad. I recommend continue and implement

Screenshots:
API Docs: 
image

Schema: 
image


bloom_nofos/nofos/api/api.py
```diff
@@ -0,0 +1,53 @@
...
+from django.core.exceptions import ValidationError
+from .schemas import NofoSchema, ErrorSchema, SuccessSchema
+from nofos.models import Nofo
+from nofos.views import _build_nofo
+
+
+class BearerAuth(HttpBearer):
+    def authenticate(self, request, token):
+        # For testing purposes accept any token that starts with "secret"
+        if token and token.startswith("secret"):
```
pcraig3, line 12: Maybe we harden up our security here by also accepting "s3cr3t"

bloom_nofos/nofos/api/schemas.py
```diff
@@ -0,0 +1,74 @@
...
+
+
+class NofoSchema(ModelSchema):
+    sections: List[SectionSchema]
+
+    class Config:
+        model = Nofo
+        model_fields = [
+            "title",
+            "filename",
```
pcraig3, line 36: Let's also return "id".

The section and subsection ids don't matter (we would never return a subsection on its own) but I feel like NOFO ID is a useful handle so that you know how to find it again.

pcraig3 (changes requested):
Over this looks great!

I've updated #75 with what I think is our go-to initial schema, but let's chat about the next steps with regards to that.

bloom_nofos/nofos/api/api.py
```diff
@@ -0,0 +1,58 @@
...
+
+
+api = NinjaAPI(
+    auth=BearerAuth(),
+    urls_namespace="api",
+    docs_url="/docs",
+)
+
+
+@api.post("/nofo/import", response={201: SuccessSchema, 400: ErrorSchema})
```
pcraig3, line 24: I should have mentioned this earlier, but let's follow REST conventions a little closer.

This means:

- Let's use a plural object noun for this `/api/nofos/{id}`
- POST route should use the same path, doesn't need 'import': `/api/nofos`
- Return a 201 response 👍  (we are doing this)
- Return the newly created object rather than a success message
- Include a `Location` header with the URL of the newly created resource (`Location: /api/nofo/123`

bloom_nofos/nofos/api/api.py
```diff
@@ -0,0 +1,53 @@
...
+def import_nofo(request, payload: NofoSchema):
+    try:
+        data = payload.dict()
+        sections = data.pop("sections", [])
+
+        # Create NOFO
+        nofo = Nofo(**data)
```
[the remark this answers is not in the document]
pcraig3: This is fixed!

bloom_nofos/nofos/api/api.py
```diff
@@ -0,0 +1,58 @@
...
+        sections = data.pop("sections", [])
+
+        # Remove fields we dont want set on import
+        excluded_fields = ["id", "archived", "status", "group"]
+        for field in excluded_fields:
+            data.pop(field, None)
+
+        # Create NOFO
+        nofo = Nofo(**data)
+        nofo.group = "bloom"
```
pcraig3, line 37: This is fine for now.

bloom_nofos/nofos/api/schemas.py
```diff
@@ -0,0 +1,77 @@
...
+            "keywords",
+            "cover_image",
+            "cover_image_alt_text",
+            "inline_css",
+        ]
+
+
+class NofoSchema(NofoBaseSchema):
+    sections: List[SectionSchema]
+
```
pcraig3, line 69: Looks like this solves the ordering problem! How does this work (and how did you figure it out)?

bloom_nofos/nofos/api/schemas.py
```diff
@@ -0,0 +1,77 @@
...
+        ]
+        model_fields_optional = ["html_class"]  # Fields that should have defaults
+
+
+class SectionSchema(ModelSchema):
+    subsections: List[SubsectionSchema]
+
+    class Config:
+        model = Section
+        model_fields = ["name", "html_id", "order", "has_section_page"]
```
pcraig3, line 26: Can we do the same thing with creating a base schema for sections as well? Would be nice to see section metadata  before the list of subsections.

bloom_nofos/nofos/api/tests.py
```diff
@@ -0,0 +1,140 @@
...
+        response = self.client.get(f"/api/nofo/{self.nofo.id}")
+        self.assertEqual(response.status_code, 401)
+
+    def test_import_nofo(self):
+        """Test importing a valid NOFO using fixture data"""
+        # Prepare fixture data by removing fields we don't want
+        import_data = self.fixture_data.copy()
+        excluded_fields = ["id", "archived", "status", "group"]
+        for field in excluded_fields:
+            import_data.pop(field, None)
```
pcraig3, line 80: I think we don't even need to do this, right? Even if these fields are included, they aren't used.

How about if we do this:

- keep them
- assert that the values for these on the new NOFO are different than these values (`assertNotEqual`)

bloom_nofos/nofos/api/tests.py
```diff
@@ -0,0 +1,140 @@
...
+            section=self.section,
+            name="API Test NOFO: Subsection 1",
+            order=1,
+            tag="h3",
+        )
+
+        # Load fixture data for import tests
+        fixture_path = os.path.join(
+            settings.BASE_DIR, "nofos", "fixtures", "json", "cms-u2u-25-001.json"
+        )
```
pcraig3, line 37: Great! 

There are a few other fixtures in there I used for making sure bad data was overwritten (`hrsa-` and `published-`), but I think we can include those checks in the `import_nofo` test and then just remove these other fixtures.

pcraig3 (changes requested):
Hello!

This is all looking super good! Thanks so much for converting those tests over, this is such a neat idea that we changed all the code underneath but the tests still (mostly) work. 

I've made some other comments in below, but I also noticed that the validation of the NOFO object doesn't match what's in our `models.py`.

After removing every field I could, it looks like this is our minimum viable NOFO we can import:

```json
{
  "title": "",
  "filename": "",
  "short_name": "",
  "number": "",
  "opdiv": "",
  "agency": "",
  "tagline": "",
  "application_deadline": "",
  "subagency": "",
  "subagency2": "",
  "author": "",
  "subject": "",
  "keywords": "",
  "inline_css": "",
  "sections": [
  ]
}
```

Some of these are reasonable, like title, number, etc, but `inline_css` and `subagency2` are examples where most NOFOs won't have those, so they don't need to be there (and they aren't required by `models.py`). 

Also, one test we had earlier with the ImportJSON route would look for `len(nofo.sections)`, so it wasn't enough to have a `sections` key, you would also need to have at least 1 section (and then ditto with subsections). 

For this PR, it's already very large, and, to be honest, it does 95% of what we want. So let's say we can work out the sections/subsections stuff when we look at #75, and what fields are required. 

For this one, let's just remove some of the rigidity. These are the fields that I would say we should not require, but use them if they exist in the data: 

- `subagency2`
- `inline_css`
- `flename`

If that's too complicated, let me know.

Overall, this is looking awesome and I am excited to see this go in!

pcraig3:
Once this PR goes in, it resolves a bunch of issues:

- #99 (spike worked!)
- #100 (we don't need this follow-up since the spike worked)
- #101 (shipped!)

bloom_nofos/nofos/api/api.py
jtmst: updated.  changed to get and create method names to be more conventional as well

bloom_nofos/nofos/api/schemas.py
jtmst: Completely by accident. Glad it worked though

pcraig3 (approved):
This is super, I am well pleased with this. 

Amazing work!

PR #119, opened 2024-12-17
Convert single-item lists in table cells into paragraphs

pcraig3:
## Summary

This PR converts lists (`ol` or `ul` elements) with 1 item into paragraphs if they are inside of table cells.

The reason for doing this is that we received an accessibility report that pointed this out to us. For someone using a screen reader, it would be very annoying to have to navigate through all the list semantics in order to just read 1 line of text. It really should just be a paragraph. 

It does make some of our tables look less even though: you can see in the "before" image below that one side has little list bullets all the way down the column, while the "after" side has some bullets and some paragraphs.

| before | after  |
|--------|--------|
|  all lists are lists      |  lists with 1 bullet have been converted to paragraph      |
| Screenshot 2024-12-17 at 10 13 07 AM | Screenshot 2024-12-17 at 10 12 57 AM |


jtmst (approved):
looks good on my local - 🚀

pcraig3:
I'm going to close this because I don't think we are going ahead with it.

PR #123, opened 2024-12-18
New theme + new staging area + CSS fixes

pcraig3:
## Summary

This PR does 4 things:

1. Adds a new theme: we are calling it "CDC Portrait (IOD)" (* adds a migration)
2. Adds a new "staging" group for users to test the NOFO Builder (* adds a migration)
3. Makes H7 headings bold in the ACF theme
4. Makes all links blue

⚠️ Note that you will have to run a migration to test these changes ⚠️ 

This PR also bumps up the version because of the migrations included.

### 1. New theme: "CDC (IOD)"

Adding a new theme! 

To add a new theme, we need a new `nofo.theme` value, and then we add a new CSS file with new colours. This needs a migration, but otherwise is a pretty easy process.

The Figma mocks are here: [https://www.figma.com/design/Vz66aRHVoI1dJ8oCFfhA1r/Untitled?node-id=0-1&p=f&t=jkLBQuHc9NFKs4XL-0](https://www.figma.com/design/Vz66aRHVoI1dJ8oCFfhA1r/Untitled?node-id=0-1&p=f&t=jkLBQuHc9NFKs4XL-0)

Screenshots are below. Note that the HTML proportions don't look dead on, but they look fine in the PDF.

| Section page: figma | Section page: HTML  |
|--------|--------|
|    Screenshot 2024-12-18 at 3 59 08 PM    |     Screenshot 2024-12-18 at 3 58 39 PM   |

| Basic info page: figma | Basic info page: HTML  |
|--------|--------|
|    Screenshot 2024-12-18 at 3 59 25 PM    |     Screenshot 2024-12-18 at 3 57 53 PM   |

### 2. New user group: "Staging environment"

Sometimes we want to add users who get a little sandbox to play around in, and we need to be able to do that without associating them with a specific OpDiv. Once you are assigned to an OpDiv, you can see all of their past and present NOFOs.

Adding a 'staging' group lets us do this, although it could be a problem if we want them all to be hermetically sealed from one another.

That is to say, if we want two different staging users to not see each others' NOFOs, then we would have to change this up.

But for now, this is good.

| NOFO index with a "staging" nofo label |
|--------|
|  Here's a screenshot of a new "Staging" NOFO.      |      
| Screenshot 2024-12-18 at 3 54 08 PM | 

### 3. Make ACF H7 headings bold 

They were hard to distinguish from body text before, but now they are good. Also made the font a tiny bit smaller to counteract how emphasized they are now.

| before | after  |
|--------|--------|
|    H7 is size 13pt, normal weight    |   H7 is size 12.5pt, bold weight        |
| Screenshot 2024-12-16 at 4 10 11 PM | Screenshot 2024-12-16 at 4 10 52 PM |


admoorgit:
@pcraig3 there are a few more tweaks to ACF heading styles — check with Jackie, but we adjusted H1 and H7. We also need to install semi-bold font.

pcraig3:
Yes, gotcha. We will make a separate PR for the ACF font overhaul.

PR #134, opened 2024-12-24
Speed up NOFO importing

pcraig3:
As we have continued to add steps to the NOFO import step, we've ended up increasing the total time needed to import a NOFO.

After running some profiling of the entire import step, we discovered the largest contributor to the import slowdown is the `add_headings_to_nofo` function, which was acocunting for over 60% of the import time for a typical NOFO.

Using a CMS NOFO as an example, we were seeing this online:

- Total import time: 21.829363 seconds 
- `add_headings_to_nofo` runtime: 14.137533 seconds

So if we can make this function more efficient, we can dramatically improve our NOFO import times.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,6 @@ The format is based on Keep a Changelog, and this project adheres to Semantic Ve
 - Add cover image for CMS-2V2-25-001
   - Also add inline images
 - Add utility classes for different list counters
-- TEMP: add benchmarking numbers so that we can test speed of imports
 
 ### Changed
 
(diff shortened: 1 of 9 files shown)


pcraig3:
Updated times for are here:

- Total import time: 8.694077 seconds (~60% reduction)
- `add_headings_to_nofo` runtime: 0.953009 seconds (~94% reduction)

pcraig3:
'Twas the night before Christmas, and all through the app,
Our imports were speeding, no time for a nap.
With batch updates flowing, and regex refined,
We're saving the hours, and leaving lag behind!

PR #135, opened 2024-12-26
Batch create Sections and Subsections in _build_nofo function

pcraig3:
## Summary

This PR speeds up NOFO importing (again) by using batch processing inside of the `_build_nofo` function, which we use when creating a NOFO.

### Details

On Friday December 20th, we received the largest NOFO we had ever seen, and trying to import it actually would result in an app timeout on the server. 

So we started profiling the various functions we run in `import_nofo`, and discovered that we had a couple of noticeable bottlenecks:

1. `add_headings_to_nofo`: took `102 seconds` out of `129  sections` 
2. `create_nofo`: took `20 seconds` out of `129 seconds`

In PR #134, we refactored `add_headings_to_nofo` in 2 ways:

1. Move regex compilation out of the for loop: this accounted for the bulk of our time savings
2. Use batch updates to update all Sections and Subsections at once instead of calling `save()` on individual Sections/Subsections

That was a huge improvement, check out the data for a generic CMS NOFO.

|                             | before | after |
|-----------------------------|--------|-------|
| total import time           |   21.829363 seconds     |   8.694077 seconds (~60% reduction)    |
| `add_headings_to_nofo ` runtime |  14.137533 seconds      |   0.953009 seconds (~94% reduction)    |

This PR uses the second strategy, batch creation, to refactor the `_build_nofo` function (used by create_nofo). We haven't seen such a dramatic speedup in absolute terms, but it's definitely a change worth making.

Here're the new numbers:

|                             | before | after |
|-----------------------------|--------|-------|
| total import time           |   8.694077 seconds     |   3.691065 seconds (~58% reduction)    |
| `create_nofo` runtime |  5.670648 seconds      |   0.78112 seconds (~86% reduction)    |

After the improvements to `add_headings_to_nofo`, `create_nofo` was the next slowest function, and one that is entirely within our codebase. This change speeds up importing significantly. 

### A note on the `save()` function

One thing to note here is that because our Subsection's `save()` function would previously run some logic to add `html_id` to subsections automatically, we now do this manually inside of `_build_nofo` because the `.save()` function for individual subsections was not getting called.

Doesn't seem to make a difference in my testing, importing a bunch of NOFOs, I still see the same number of broken links in both (so we are not severing links in the original word doc by doing this change) and all the tests pass as before, so I think we are good to ship this.

### Note 2: on CRUD events

Since we were updating the `updated` timestamp previously when calling `save`, and now we don't call `save`, our audit event trail is much less noisy. 

We don't care about this: this is a good thing.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,8 @@ The format is based on Keep a Changelog, and this project adheres to Semantic Ve
 ### Changed
 
 - All tables in appendices are full-width
+- Speed up "create_nofo" function
+  - Batch create sections and subsections in "\_build_nofo"
 
 ### Fixed
 
--- a/bloom_nofos/nofos/models.py
+++ b/bloom_nofos/nofos/models.py
@@ -7,7 +7,7 @@ from django.urls import reverse
 from django.utils import timezone
 from martor.models import MartorField
 
-from .utils import create_subsection_html_id
+from .utils import add_html_id_to_subsection
 
 COACH_CHOICES = [
     ("aarti", "Aarti"),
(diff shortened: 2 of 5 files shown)


jtmst (approved):
tests fine on local, curious to see how must faster it is in prod

pcraig3:
Rebased since the other PR went in (#137). Merging!

Issue #136, opened 2024-12-26
Copy metadata for matching subsections when re-importing


pcraig3:
When re-importing a NOFO, we are currently creating entirely new sections and subsections and then assigning them over the old ones. 

This means we lose all the updates to the old subsections, if any.

Some of the metadata fields which would be easy for us to preserve on re-imports are:
- subsection classnames
- whether or not something is a callout box

The classnames are a string, and the callout box information is a boolean.

Matching the subsections to each other is not an exact science here though, so we will have to use a best-guess for this.

There are 2 ways to approach this, both of which should be the same thing:

1. exact match for `subsection.section.name` and `subsection.name`, or
2. match `html_id` after removing the `counter` number

The `subsection.html_id` is created ([see here](https://github.com/HHS/simpler-grants-pdf-builder/blob/main/bloom_nofos/nofos/utils.py#L60)) by combining a counter (for uniqueness), the section name, and the subsection name. The counter can't be relied on to be the same, but if the other two line up, we are 99% sure they are the same subsection.

If there is a match, transfer the `html_class` (if exists) and `callout_box` (either value) to the new subsection before replacing them.

This is an easy first step we can take to make reimporting a little bit easier.

pcraig3:
Update to this ticket as of Jan 7th 2025. 📆 

Demoed to our HRSA users this morning and they generally loved the idea but make a request that we also add a confirmation checkbox, and then in chatting to Adam it sounds like we should reverse course on callout boxes.

## Add a checkbox on the reimport page

We heard from HRSA that keeping page breaks is usually a good idea, but that sometimes they want them to go away. Their suggestion was to add a checkbox that defaults to checked and says something like “Keep page breaks from current NOFO”

- If the checkbox is checked, preserve page breaks.
- If the checkbox is not checked, overwrite everything (current behaviour)

You can find checkbox info here: https://designsystem.digital.gov/components/checkbox/

Also, some of our Django forms are Django form classes, but the reimport forms are raw HTML, so we don't need to faff around with that.

## Do not carry over callout boxes

Adam and I talked afterwards about the callout boxes. Unlike page breaks, callout boxes can’t be easily edited in the NOFO UI, so there is a low chance that people managed to change callout box information _after_ importing (although callout boxes can be created when you add a new section). 

So we discussed not overwriting the callout box field, just page breaks. 

I’ve asked in the NOFO Builder channel for a recco, but I think we are probably going to not carry these over.

[...]

PR #143, opened 2025-01-03
Copy metadata for subsections on re-import of nofo

jtmst:
This code allows us to save metadata for subsections on reimport, and apply them to the new import.  It tracks those changes and tells the user what changes were preserved as part of the success message on reimport.

QA Steps:
- Import a nofo
- make a subsection a callout field
- change a subsections html class
- reimport the same nofo
- verify that the success message reflects the changes you made
- verify that the changes are actually still what you set them to be

Note: There is currently behavior for page breaks that is showing them as changed, whether they were in the original document or not.  I'm not sure if this is actually an issue or not, to verify this, upload a nofo, and then reupload the same nofo.  The success message will tell you the page breaks are preserved, even if they were originally there.

Screenshot: 
image

image

### DIFF
--- a/bloom_nofos/nofos/nofo.py
+++ b/bloom_nofos/nofos/nofo.py
@@ -327,6 +327,36 @@ def create_nofo(title, sections, opdiv):
         raise e
 
 
+def _get_subsection_section_key(section_name, subsection):
+    """
+    Helper function to generate a unique key for a subsection based on section name and subsection content.
+    Returns a string in format "section_name|subsection_name" or empty string if no valid key can be generated.
+    """
+    if hasattr(subsection, "name"):
+        subsection_name = subsection.name
+        body = subsection.body if hasattr(subsection, "body") else None
+    else:
+        subsection_name = subsection.get("name", "")
+        body = subsection.get("body", "")
+
+    if not subsection_name and body:
+        # If no name, use first line of body as fallback
+        if isinstance(body, list):
+            body = "".join(str(item) for item in body)
+
+        body = str(body).strip()
+        if body:
+            # Convert markdown to HTML for consistent comparison
+            html = markdown.markdown(body)
+            soup = BeautifulSoup(str(html), "html.parser")
+            subsection_name = soup.get_text().split("\n")[0].strip()
+
+    if not subsection_name:
+        return ""
+
+    return f"{section_name}|{subsection_name}"
+
+
 def preserve_subsection_metadata(nofo, new_sections):
     """
     Preserves page break classes from existing subsections.
@@ -346,48 +376,19 @@ def preserve_subsection_metadata(nofo, new_sections):
     for section in new_sections:
         section_name = section.get("name", "")
         for subsection in section.get("subsections", []):
-            subsection_name = subsection.get("name", "")
-            if not subsection_name and subsection.get("body"):
-                try:
-                    # If no name, use first line of body as fallback
-                    body_text = subsection.get("body", "")
-                    if body_text:
-                        if isinstance(body_text, list):
-                            # Join list items into a string
-                            body_text = "".join(str(item) for item in body_text)
-                        # Extract text from HTML
-                        from bs4 import BeautifulSoup
-
-                        soup = BeautifulSoup(str(body_text), "html.parser")
-                        subsection_name = soup.get_text().split("\n")[0].strip()
-                except Exception:
-                    continue
-
-            if section_name and subsection_name:
-                key = f"{section_name}|{subsection_name}"
-                new_section_lookup[key] = True
+            try:
+                key = _get_subsection_section_key(section_name, subsection)
+                if key:
+                    new_section_lookup[key] = True
+            except Exception:
+                continue
 
     # Store page break classes from existing sections if they exist in new sections
     for section in nofo.sections.all():
         for subsection in section.subsections.all():
             try:
-                subsection_name = subsection.name
-                if not subsection_name and subsection.body:
-                    # If no name, use first line of body as fallback
-                    body = subsection.body.strip() if subsection.body else ""
-                    if body:
-                        # Convert markdown to HTML for consistent comparison
-                        import markdown
-
-                        html = markdown.markdown(body)
-                        soup = BeautifulSoup(html, "html.parser")
-                        subsection_name = soup.get_text().split("\n")[0].strip()
-
-                if not subsection_name:
-                    continue
-
-                name_key = f"{section.name}|{subsection_name}"
-                if name_key in new_section_lookup:
+                name_key = _get_subsection_section_key(section.name, subsection)
+                if name_key and name_key in new_section_lookup:
                     # Extract only page break classes
                     html_class = getattr(subsection, "html_class", "") or ""
                     classes = html_class.split()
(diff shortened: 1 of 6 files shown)


bloom_nofos/nofos/nofo.py
```diff
@@ -234,6 +234,84 @@ def create_nofo(title, sections):
...
+
+    # Create lookup dict for new sections
+    new_section_lookup = {}
+    for section in new_sections:
+        section_name = section.get("name", "")
+        for subsection in section.get("subsections", []):
+            if not subsection.get("name"):
+                continue
+
+            key = f"{section_name}|{subsection.get('name')}"
```
pcraig3, line 252: Subsections don't need to have names (for example subsections that come immediately after a callout box often don't have a `.name`), so we need to think of something to match on if there is no section name. 

Maybe we look at the first line/sentence of body content?

bloom_nofos/nofos/nofo.py
```diff
@@ -234,6 +234,84 @@ def create_nofo(title, sections):
...
+
+def restore_subsection_metadata(nofo, preserved_metadata):
+    for section in nofo.sections.all():
+        for sub in section.subsections.all():
+            name_key = f"{section.name}|{sub.name}"
+            id_key = sub.html_id.split("--", 1)[-1] if sub.html_id else None
+
+            metadata = preserved_metadata.get(name_key) or preserved_metadata.get(
+                id_key
+            )
```
pcraig3, line 305: Not sure I follow — when would the name not match but the id would?

I guess if you changed the name of a section manually, the old id would still be around? 
What happens if the subsection has no name?

bloom_nofos/nofos/nofo.py
```diff
@@ -234,6 +234,84 @@ def create_nofo(title, sections):
...
+
+            if metadata:
+                sub.html_class = metadata["html_class"]
+                sub.callout_box = metadata["callout_box"]
+                sub.save()
```
pcraig3, line 310: This once-per-sub `.save()` is the kind of thing that gets us in trouble with larger nofos. I would say it's probably better to batch update all subsections once their attributes have been updated.

bloom_nofos/nofos/views.py
```diff
@@ -356,22 +358,44 @@ def nofo_import(request, pk=None):
...
+                    )
+
+                if html_class_count > 0:
+                    success_msg += (
+                        f"<br><br>Preserved {html_class_count} section settings for HTML classes:<br>• "
+                        + "<br>• ".join(html_class_items)
+                    )
+
+                messages.success(request, success_msg)
+
```
pcraig3, line 395: Nice idea, but I think this is too verbose for anything except debugging. 

When you have a lot of changed sections, there's a lot of text to sort through. 

Screenshot 2025-01-06 at 5 47 10 PM

Maybe we simplify this to just say how many of each are preserved, and we just say "page breaks" even though that is not strictly the only class we preserve.

Screenshot 2025-01-06 at 5 50 24 PM

Very annoying that everything is just 1 long text string — maybe we do something about that at some point.

pcraig3: Alternatively, maybe we don't change the reimport message at all. We can just silently update the behaviour and act like this was always the correct way to do things. We don't have too many users right now so we can just tell our existing users we have made this change.

Do you have a preference?

pcraig3 (commented):
I like the way it's heading, maybe we don't need to return as many things from that initial method, but I will defer to you on that. 

One change that is getting caught up in this is the appropriately-named [`add_page_breaks_to_headings`](https://github.com/HHS/simpler-grants-pdf-builder/blob/main/bloom_nofos/nofos/nofo.py#L152) function, which adds 3 page breaks on its own to every NOFO we import with those matching headings (approximately 100% of them).  Since its misleading to "preserve" those, which were never added by a user, we should offset the count by 3. 

Would be nice to see a few tests as well, but I assume those would come later as this is a draft PR.

pcraig3:
Also! New update to the issue in this comment: https://github.com/HHS/simpler-grants-pdf-builder/issues/136#issuecomment-2575971233

bloom_nofos/nofos/nofo.py
jtmst: That was the thinking, that a manual edit could create a mismatch.  Added a similar fallback as above to default to body text when no subsection name exists.

bloom_nofos/nofos/views.py
jtmst: Agreed with the second point as we discussed.  updated the help text to let users know page breaks will be preserved

bloom_nofos/nofos/views.py
```diff
@@ -357,13 +359,42 @@ def nofo_import(request, pk=None):
...
+                    "Re-imported NOFO from file: {}".format(nofo.filename),
+                )
+
+                # Create audit event for reimporting a nofo
+                create_nofo_audit_event(
+                    event_type="nofo_reimport", nofo=nofo, user=request.user
+                )
+
+                return redirect("nofos:nofo_edit", pk=nofo.id)
+
```
pcraig3, line 397: Does moving this up make a difference? I can't really tell if it does.

bloom_nofos/nofos/views.py
```diff
@@ -357,13 +359,42 @@ def nofo_import(request, pk=None):
...
                 nofo = overwrite_nofo(nofo, sections)
+
+                # If preserve_page_breaks is not checked, remove page break classes from preserved metadata
+                if request.POST.get("preserve_page_breaks") != "on":
+                    for key in preserved_metadata:
+                        if "html_class" in preserved_metadata[key]:
+                            classes = preserved_metadata[key]["html_class"].split()
+                            classes = [
+                                c for c in classes if not c.startswith("page-break")
+                            ]
```
pcraig3, line 374: Maybe I am being dim, but why are we excluding "page-break*" classes?

bloom_nofos/nofos/templates/nofos/nofo_import_overwrite.html
```diff
@@ -12,23 +12,37 @@
...
+          id="preserve_page_breaks"
+          type="checkbox"
+          name="preserve_page_breaks"
+          checked
+        >
+        <label class="usa-checkbox__label" for="preserve_page_breaks">
+          Keep page breaks from current NOFO
+        </label>
+      </div>
+    </div>
```
pcraig3, line 45: Can we move this under the "import" input but before the button?

bloom_nofos/nofos/templates/nofos/nofo_import_overwrite.html
```diff
@@ -12,23 +12,37 @@
...
       </p>
     </div>
   </div>
 
   {% with nofo|nofo_name as nofo_name_str %}
-    {% with "Edit  “"|add:nofo_name_str|add:"”" as back_text %}
+    {% with "Edit “"|add:nofo_name_str|add:"“" as back_text %}
       {% url 'nofos:nofo_edit' nofo.id as back_href %}
-      {% include "includes/page_heading.html" with title="Re-import “"|add:nofo_name_str|add:"”" back_text=back_text back_href=back_href only %}
+      {% include "includes/page_heading.html" with title="Re-import “"|add:nofo_name_str|add:"“" back_text=back_text back_href=back_href only %}
```
pcraig3, line 26: The second quotation mark is now pointing the wrong way:

Screenshot 2025-01-17 at 10 06 09 AM

We want to use this one:
- `”`

Not this one:
- `“`

bloom_nofos/nofos/nofo.py
```diff
@@ -256,6 +256,84 @@ def create_nofo(title, sections, opdiv):
...
+    for section in new_sections:
+        section_name = section.get("name", "")
+        for subsection in section.get("subsections", []):
+            subsection_name = subsection.get("name", "")
+            if not subsection_name and subsection.get("body"):
+                # If no name, use first line of body as fallback
+                body_text = subsection.get("body", "")
+                if isinstance(body_text, list):
+                    body_text = "".join(str(item) for item in body_text)
+                subsection_name = body_text.split("\n")[0].strip()
```
pcraig3, line 284: I am getting an error when importing a NOFO caused by this line (I can slack it to you):

Screenshot 2025-01-17 at 9 56 29 AM

It seems like it's because we are calling `.split()` on a BeautifulSoup `Tag` when the `body_text` is not a `list` (the error message is not helpful in diagnosing this).

We can use  or `.get_text()` to get the inner text content as a string before calling this, but then I don't think there will be any newlines in there. Maybe we limit the key by some arbitrary length (like first 100 chars or something). 

We could also get a hash from the content if we wanted something unique and not full or weird characters. (does that matter though?)

This is getting a bit confusing. What do you think?

pcraig3 (changes requested):
Hello! I ran into a bug while testing this so I am kicking it back. I've left feedback in the comments.

bloom_nofos/nofos/views.py
jtmst: The idea was to copy all classes over except for the page breaks when the box is unchecked.  I was thinking other classes could have been added potentially, but in retrospect that might not actually be necessary

bloom_nofos/nofos/nofo.py
jtmst: Made some modifications to simplify this, tested it with a few different NOFOs

pcraig3 (changes requested):
It's getting there! Works fine for me, just had some minor feedback.

pcraig3 (approved):
🎈 We did it!! 🎈

Issue #125, opened 2024-12-19
Improvements to NOFO reimporting


pcraig3:
This is a proto-story, not yet a story. It's a place to capture the discussion around reimporting. 

So, one of our users mentioned that reimporting causes lost work, and if there is anything we can do about this.

When a NOFO is re-imported, the sections and subsections are completely recreated and assigned to the current NOFO object. This means that anything that's in the existing NOFO is completely wiped out by the re-import action. 

I pulled some data to validate this need and saw that there is a pretty healthy amount of re-importing going on, although it is heavily concentrated in our non-Bloom users. 

Some numbers for you:

| total reimports (since Nov 8)        | 209                 |
|--------------------------------------|---------------------|
| Bloom users reimports                | 57                  |
| Other users reimports                | 148                 |
| Highest re-imports for a single user | 41 (8 individual NOFOs) |
| Total NOFOs reimported               | 46                  |
| Average re-imports per NOFO listed   | 4.5                 |
| Highest re-imports for a single NOFO | 17 (HRSA-25-061)    |

So it seems like there is reason here to try and address this in some way. 

Preference goes to ideas that are easy to do.

Here are some easy ideas:

- Get list of all subsections that have been modified since import (could display in a list like broken links) 
- Offer to create a new NOFO when there are modified subsections instead of overwriting current NOFO
- Copy over metadata onto matching subsections. 
  - For example: if we find a subsection with the same name in the same section, and it has a "page break" or "callout box" applied, we should carry that over to the new subsection

Here is a medium-effort idea:

- If there are changed subsections, offer 3 options:

  - 1. Overwrite everything (current behaviour)
  - 2. Create a copy of this NOFO, do not overwrite it. You can manage replacing the content.
  - 3. Try to replace body content from subsections in new NOFO with body content of matching edited subsection in current object. (We would have to match subsections on matching section + subsection name, so this could still lose data).

pcraig3:
Update: I've merged #211 which adds a "compare nofos" feature to the NOFO Builder.

PR #144, opened 2025-01-03
[WIP] NOFO re-import text diff UI concept

pcraig3:
## Summary

⚠️  DO NOT MERGE THIS PR. IT IS A CONCEPT ONLY ⚠️ 

This is related to #125.

### What is it?

The idea here is to demo a UI concept for re-importing a NOFO with changes from the current NOFO.

The objective here is to provide a simple way for people to y/n changes that have happened since their last import. The idea would be that you can see all the changed subsections and either select "yes, import the new content" (check the box) or "no, keep my content" (uncheck the box). 

Only the titles of edited subsections would be shown and then people would be able to click into each one to see the specific change. 

Changes can happen in 2 ways: 

1. You have manually edited a subsection using the NOFO Builder
2. The content in the Word Doc has changed

Regardless, the text diff will show additions and deletions, it doesn't differentiate where the change comes from.

Added some buttons at the bottom for people who will either say "just import and take everything", or "do not overwrite anything" (metadata and classnames* etc will still get imported).

### How to test this

1. Make a "save as" version of an existing NOFO word doc you have.
2. Add some changes to the new doc
3. Import the first Word doc to create a new NOFO
4. Go to the "edit" page for your new NOFO
5. Manually navigate to `/nofos/(id)/compare`
6. Import your "save as" version of the NOFO
7. Be amazed

### gif

Screen Recording 2025-01-03 at 5 03 32 PM

### TODOs

There are some major problems with this code as it stands:

1. It doesn't import anything
2. It creates a new NOFO every time you run it
3. It is extremely moist (not even trying to be DRY)
4. It doesn't handle new subsections or removed subsections
5. Lots of problems

But anyway, it's a start. Let's discuss further next week. Adam is on board.

### DIFF
--- a/bloom_nofos/bloom_nofos/static/styles.css
+++ b/bloom_nofos/bloom_nofos/static/styles.css
@@ -875,3 +875,27 @@ details.usa-accordion .usa-accordion__content ul li:not(:last-of-type) {
 .usa-error-message {
   font-size: 1rem;
 }
+
+/* comparison form */
+
+.comparison-form {
+  max-width: 100%;
+}
+
+.comparison-form ul {
+  list-style: none;
+}
+
+.comparison-form summary {
+  margin-bottom: 10px;
+}
+
+.comparison-form .comparison-form--subsection-list li label {
+  display: flex;
+  align-items: baseline;
+  gap: 1em;
+}
+
+.comparison-form .diff {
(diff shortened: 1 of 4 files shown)


pcraig3:
Closing since #211 is now the replacement for this. 👍

PR #145, opened 2025-01-06
Show H7 headings in "edit" and "preview" modes

pcraig3:
## Summary

This PR makes 3 changes:

1. Add h7 styles to the "edit" mode and "preview" pane in the markdown editor
2. Update body-text header styling in the NOFO content for the edit page
3. Remove the "Fix in Word and reimport" in heading error messages

### 1. Add h7 styles to the "edit" mode and "preview" pane 

Since our H7 styles are shimmed, they don't show up as regular HTML elements would. These means we need to add custom styling if we want to see them show up in the edit mode. 

Here is how it looks now in the "Edit" screen:

| before | after |
|--------|-------|
|  h6 is too small. <br> h7 has no styling.       |    h6 is proportional. <br> h7 is styled like a heading.  |
|      Screenshot 2025-01-06 at 3 30 43 PM |    Screenshot 2025-01-06 at 3 30 21 PM    |

And here is how it looks on the "Preview" pane of the markdown editor:

| before | after |
|--------|-------|
|  h7 looks like body text      |   h7 has styling that differentiates if from body text     |
|  Screenshot 2025-01-06 at 3 27 51 PM      |    Screenshot 2025-01-06 at 3 27 13 PM   |

Note that the heading levels in the preview pane are not the same as on the "edit" page because there is some opinionated styling (using lots of `!important`s) being applied to that preview content, and I figured I would just leave it intact rather than trying to override everything.

### 3. Remove the "Fix in Word and reimport" 

When we wrote that message, it wasn't yet possible to change a heading level using the NOFO Builder. Since then, you can do it and people are doing it. So no need for the "Reimport" message at this point, since you don't need to do that.

| before | after |
|--------|-------|
|   Screenshot 2025-01-06 at 4 32 26 PM     |    Screenshot 2025-01-06 at 4 39 27 PM   |

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ The format is based on Keep a Changelog, and this project adheres to Semantic Ve
 ### Changed
 
 - Change heading sizes in edit mode
+- Remove "Fix in Word and reimport." note in warning message
 - Change ACF heading styles
   - h1s: 36 pt, 700
   - h4s: 600
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,11 +8,13 @@ The format is based on Keep a Changelog, and this project adheres to Semantic Ve
 
 ### Added
 
+- Add heading styles for H7s in edit and preview modes
 - Add cover image for HRSA-25-007
 - Add list style for quint nested lis
 
 ### Changed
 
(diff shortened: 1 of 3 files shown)


jtmst (approved):
🚀

Issue #148, opened 2025-01-07
Links in callout boxes are touching the colon


pcraig3:
There is a visual bug where links in our callout boxes are touching the colon. 

When we look at the content, there is a space there, so something funny is going on here.

| link is touching colon in the UI | in the edit view, it looks normal |
|--------|-------|
|    ![Image](https://github.com/user-attachments/assets/65bad851-93b9-4023-b8d5-9516c678935a)    |   ![Image](https://github.com/user-attachments/assets/dc31585f-295a-4b3a-91c9-33ed68deca1b)    |

Someone needs to go down and find out!

pcraig3:
Solved by #149

Issue #147, opened 2025-01-07
Links should all be blue


pcraig3:
With Simpler NOFOs, OpDivs can have any link colour they want as long as it's `--color--cdc-blue`. (White is also allowed). 

Sometimes links are in the theme colour, other times they are blue. After some discussion in Slack, we have settled on `--color--cdc-blue` as the colour for all links on a non-contrasty background. 

There are 3 main areas we need to check:

1. Links in body text
2. Links in tables
3. Links in callout boxes

For body text and tables, it should be pretty obvious, but callout boxes come in 2 flavours: there is always a 'primary' colour and an alternate colour. 

Here is a screencap of an ACL NOFO where the callout boxes have non-blue links. 

![Image](https://github.com/user-attachments/assets/43f740c1-299b-4c71-9eef-04d1a45d39c1)

For this ticket:

- Create a NOFO with links in both kinds of callout boxes
- Switch to every theme and see what colour they show up as
- White links are okay if they are on a dark background
- All other links should be CDC blue _unless_ they create an a11y violation
  - If so, maybe we leave it? Or kick it back to Adam.

pcraig3:
Solved by #149

PR #149, opened 2025-01-08
Fix links in callout boxes + All links are blue

pcraig3:
## Summary

There are 3 (small) changes in this PR:

- Fixes links in callout boxes that touch the colon (solves issue #148)
- Standardizes our links as blue (Solves issue #147)
- Adds Jana as a designer (may remove this from the PR depending on whether #138 will go in afterwards)

### 1. Fixes links in callout boxes that touch the colon

This was to do with how HTML handles whitespace. When we had text only, the HTML looked like this:

```html
<p><strong>Stat authority:</strong><span> Here it is</span></p>
```

In that case, the space in the `span` element is preserved when rendered.

However, when the text after the colon started with a link, we would get this:

```html
<p><strong>Stat authority:</strong><span> <a href="https://google.com">Here it is</a></span></p>
```

This space disappears, and if we were to add a space to the link itself, we would see the underline continue up to the colon. Not what we want. 

The solution was to insert a space into the `strong` tag after the colon, so now we have this:

```html
<p><strong>Stat authority: </strong><span> <a href="https://google.com">Here it is</a></span></p>
```

And that seems to work just fine.

| before | after |
|--------|-------|
|    Screenshot 2025-01-08 at 10 11 47 AM    |   Screenshot 2025-01-08 at 10 11 18 AM    |

### 2. Standardizes our links as blue 

Links should always be blue, we have decided. 

| before | after |
|--------|-------|
|   Screenshot 2025-01-08 at 10 14 18 AM     |   Screenshot 2025-01-08 at 10 14 45 AM    |

Also, added a white circle to this HRSA callout box because it looks better.

| before | after |
|--------|-------|
|    Screenshot 2025-01-08 at 10 15 19 AM    |    Screenshot 2025-01-08 at 10 15 48 AM   |

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,9 @@ The format is based on Keep a Changelog, and this project adheres to Semantic Ve
 
 ### Fixed
 
+- Fix for links after colons in callout boxes
+  - Visually, they would look like "Key:Link", now they look like "Key: Link"
+
 ## [1.42.0] - 2023-12-24
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,8 @@ The format is based on Keep a Changelog, and this project adheres to Semantic Ve
 - All tables in appendices are full-width
 - Speed up "create_nofo" function
   - Batch create sections and subsections in "\_build_nofo"
+- All links are blue now, not black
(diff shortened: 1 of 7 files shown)


jtmst (approved):
code looks good and verified visual differences on local. nice!

Issue #142, opened 2025-01-02
[Bug] Lists not explicitly styled as lists get turned into paragraphs


pcraig3:
Due to how Word works and how mammoth works, sometimes we define a particular class of element as a `paragraph` when it might also be a list.

Here's a good example:

| Word document styling | mammoth document HTML conversion |
|-----------------------|----------------------------------|
|    lots of lists                   |    almost everything is a paragraph now                              |
|         <img width="1285" alt="Image" src="https://github.com/user-attachments/assets/85fc1a51-441b-4678-9f75-be9e85b4b553" />              |                        <img width="1285" alt="Image" src="https://github.com/user-attachments/assets/31d83539-3359-4c89-a75a-1af160bda75c" />          |

What is happening mechanically is that [I am defining style maps](https://github.com/mwilliamson/python-mammoth?tab=readme-ov-file#writing-style-maps) that map particular classes to HTML elements. However, some of these classes are very general ([`paragraph`](https://github.com/HHS/simpler-grants-pdf-builder/blob/main/bloom_nofos/nofos/utils.py#L276), [`normaltextrun`](https://github.com/HHS/simpler-grants-pdf-builder/blob/main/bloom_nofos/nofos/utils.py#L159), `Default`) and can apply equally to a paragraph or a list, depending on how the list is created. 

I am not sure if there is a good way to 

- preserve the intent of the original Word doc, while
- being explicit about how to treat content that comes in

Something I've done in the past has been to ignore styles that are too general ([like `Default`](https://github.com/HHS/simpler-grants-pdf-builder/blob/main/bloom_nofos/nofos/utils.py#L256)), but it feels risky to me because then we don't really know what it will come in as. However, if we can't do anything else about this, then that's probably the best thing to do on balance. 

I've opened up an issue in the mammoth repo to see if there is a better answer to this: https://github.com/mwilliamson/python-mammoth/issues/151

Will update this issue once that one gets an answer.

pcraig3:
Closing since #158 was merged.

PR #158, opened 2025-01-13
Better support for lists that are not literally styled as lists in MS Word

pcraig3:
## Summary

This PR adds support for unstyled lists from Word docs that we import. 

It means there will be less work for the NOFO design team, but I am not sure what proportion of list import problems it solves (eg, does this solve 20% of issues, or 80% of issues?).

Here are some screenshots of the initial Word Doc, with before and after shots of the rendered HTML NOFO.

| Word | NOFO.rodeo (before) | NOFO.rodeo (after |
|--------|-------|------|
| Lists everywhere, but they are not all lists.       |   Only the very last list items are converted. The rest are paragraphs.    |  All list items in the original doc are list items in NOFO builder.   |
|   Screenshot 2025-01-13 at 4 11 13 PM    |  Screenshot 2025-01-13 at 4 10 32 PM  |   Screenshot 2025-01-13 at 4 14 27 PM  |

To test this:

- Copy the "lists.docx" file to your desktop
- Import a new document from `main` branch
- Import a new NOFO from the `undocumented-list-interpreter` branch
- See that new list items are there

## More details

So this problem has plagued us for some time. The mammoth library has a default rules that it applies when it converts Word content into HTML, but you can add your own via [style maps, which are documented in the github repo](https://github.com/mwilliamson/python-mammoth?tab=readme-ov-file#writing-style-maps). The problem I have been running into is that when I add explicit style for basic styles (like "normaltextrun" or "Default") these can sometimes be list items and other times they are just paragraphs, so it is not granular enough to apply to all styles.

I raised [this issue on the mammoth repo](https://github.com/mwilliamson/python-mammoth/issues/151) and was directed to some [undocumented style map rules that are in the source code which try to identify list items explicitly](https://github.com/mwilliamson/python-mammoth/blob/master/mammoth/options.py#L75). This is pretty much what I had been looking for: a way to identify list items independently of their explicit style map rules, because the style map rules are often the same for paragraphs and lists. 

More here: https://github.com/HHS/simpler-grants-pdf-builder/issues/142

Anyway, this PR pulls in the mammoth style rules specifically for list items (which are interpreted sequentially, first has priority (opposite of CSS)), and also adds some tests and a `.docx` fixture document. 

I've not been testing the mammoth conversion this way in the past, but since we have some known problematic text blocks, now we have a document we can add to if we want to test specific interventions we are making, instead of just winging it on a per-NOFO basis as I have been doing.

This is not an exact science by any means. Word is a ubiquitous, phenomenally successful software with decades of tech debt, I'm sure, and trying to capture all possible style maps is impossible. What this PR gives us is slightly improved performance over an understood pain point, and adds tests that convert some known text that is representative of what we see all the time.


jtmst (approved):
Tested on local and looks good.  Had to tweak the lists.docx file to account for model validation, otherwise everything checks out!

pcraig3:
Merging!

Issue #161, opened 2025-01-14
Add support for Single Source NOFOs


pcraig3:
## Summary

Single Source Justification (SSJ) NOFOs only have 1 applicant. 

In goverment-speak, we would say:

> [Procurement by noncompetitive proposals, referred to as sole source procurement, is procurement through solicitation of a proposal from only one source](https://www.ojp.gov/sites/g/files/xyckuh241/files/media/document/Sole_Source_FactSheet_C.pdf).

For us, the main difference is that we will show a shorter version of the Before you begin page, because we don't need to tell people to sign up on SAM.gov or anywhere else. 

Here's what that looks like in practice:

| before | after |
|--------|-------|
|    lots of stuff on how to apply    |   Just say when to apply     |
|    <img width="717" alt="Image" src="https://github.com/user-attachments/assets/ebf9317b-455c-46e4-b8df-9fb4fe87b701" />    |   <img width="717" alt="Image" src="https://github.com/user-attachments/assets/a26dd010-529d-4229-add2-3e3683bd99f7" />    |

To accomplish this, I think it makes sense to add a new NOFO field and then we can add/remove this in the admin section.

pcraig3:
This is closed since #162 is in!

PR #162, opened 2025-01-14
Add sole_source_justification field to NOFOs

pcraig3:
## Summary

This PR adds a new field to our NOFO objects: `sole_source_justification`

Currently, the only difference between a SSJ NOFO and a regular one is that we make the Before You Begin page much shorter. Since applicants are not really applying to these SSJ NOFOs, they don't follow the same process.

We had some logic in the app already for a specific SSJ NOFO to show the shorter BYB page, but it was hard-coded logic in our template file, which is not scalable: https://github.com/HHS/simpler-grants-pdf-builder/commit/00f3ae728feccbff234bec358840272907663c07

[Upcoming SSJ NOFOs are listed Figma](https://www.figma.com/board/ZQtWayFyCkBQcfg4j06wpG/FY25-NOFO-Production-Jam?node-id=2222-4570&t=zxOhxy03WOrj86rT-4) (non-public link).

Because this PR contains a migration, we also bump the version number (minor update because this is a new feature).

Note: most of the following text is coped in from issue #161

### Details

Single Source Justification (SSJ) NOFOs only have 1 applicant. 

In goverment-speak, we would say:

> [Procurement by noncompetitive proposals, referred to as sole source procurement, is procurement through solicitation of a proposal from only one source](https://www.ojp.gov/sites/g/files/xyckuh241/files/media/document/Sole_Source_FactSheet_C.pdf).

For us, the main difference is that we will show a shorter version of the Before you begin page, because we don't need to tell people to sign up on SAM.gov or anywhere else. 

Here's what that looks like in practice:

| before | after |
|--------|-------|
|    lots of stuff on how to apply    |   Just say when to apply     |
|    Image    |   Image    |

And there is a new field in our admin view:

Screenshot 2025-01-14 at 1 51 41 PM


jtmst (approved):
LGTM.  Only thing I'd consider changing is finding another way to manage the list of ssj_ids, but I'm not sure off hand what would be better

Issue #159, opened 2025-01-13
[Bug] "Basic information" heading missing after importing HRSA-25-066


pcraig3:
## Summary

Okay, so I cloned one of our NOFOs via API and then `POST`ed a new version to my local DB and also to prod. 

In the new versions we have, the "Basic Information" heading is missing.

| original | new version |
|--------|-------|
|     <img width="583" alt="Image" src="https://github.com/user-attachments/assets/c18f44b8-eaf9-4938-9429-ee4e51d32f2c" />   |   <img width="583" alt="Image" src="https://github.com/user-attachments/assets/21efdaf6-7ef8-47ba-8c36-f09180432010" />    |

This is weird, because the data should all the be same. What gives?

### What is happening?

There's a few things going on here:

The actual code that shows this heading is pretty specific about what it is looking for: https://github.com/HHS/simpler-grants-pdf-builder/blob/main/bloom_nofos/nofos/models.py#L313

It looks for the "first()" section, and _then_ the first subsection ordered by the "order" number.

When we `GET` NOFO data from the API, the sections and subsections themselves are _not_ returned in order of their "order" number. To be fair, they are mostly returned in order, except that the last "Endnotes" section is returned first. Basically, this is happening:

```json
{
  "opportunity_name": "My awesome NOFO",
  "sections": [ 8, 1, 2, 3, 4, 5, 6, 7 ]
}
```

And then, when we look at subsections, it's the same again. For the first section ("Step 1"), this is the order of subsections that were returned: `4, 2, 11, 17, 22, 14, 1, ..." 

So this is pretty random looking. 

This is a category of bug we have seen before, because when NOFO documents are imported, the sections and subsections are created sequentially, leading to a 1:1 relationship between `order` and the sequence of `id`/`created` values. That is to say, when you import a NOFO, the order in which they are created are the same as the `order` numbers. 

However, the `order` value exists so that we can, say, add a new subsection into the middle of an existing NOFO, and not have to reimport everything again.  So the `order` should be treated as canonical whenever the NOFO is being viewed. 

## Things to fix

Note that only the first item needs to be fixed. The others are aesthetic. However, I would really like to get 2 in, and then for 3, we can just see if it's feasible or if we would have to rewrite a ton of code.

- [x] 1. Fix this code to return sections by "order" as well: https://github.com/HHS/simpler-grants-pdf-builder/blob/main/bloom_nofos/nofos/models.py#L313
[...]

PR #177, opened 2025-01-16
logic and tests for returning nofos with ordered sections

jtmst:
Issue: https://github.com/HHS/simpler-grants-pdf-builder/issues/159

This specifically addresses the second task in the above ticket.  This change makes it so that the api response has its sections and subsections are correctly ordered by their "order" property

Includes unit test

QA Steps:
- create a nofo
- add sections
- manually swap some sections around so that a they are in a different order than their creation order
- use the http://localhost:8000/api/nofos/{id} endpoint to retrieve that nofo
- verify that they are correctly ordered by order, not creation date


pcraig3 (approved):
This is great! Thanks for solving this 🍰 

giphy

Issue #175, opened 2025-01-16
Shrink the Table of Contents based on the number of items


pcraig3:
## Summary

The Table of Contents generally fits on one page, but sometimes there is an extra header or two (or Endnotes, or an Appendix), and 1 item falls onto the next page, completely on its own. See image: 

| image of 1 ToC item breaking onto a new page |
|--------|
|  ![Image](https://github.com/user-attachments/assets/51dfd3b2-85f6-4e7a-826e-a44dc62fa424)      |

When this happens, I generally use the admin backend to add custom CSS which removes some top padding from the list items. 

The current rule we have is to add 8px to the top of our list items:

```html
.toc ol li {
    margin-top: 8px;
}
```

We should add a templatetag or filter or something that can automatically take this down to `6px` if the table of contents is too long.

pcraig3:
This is closed since #190 went in!

PR #190, opened 2025-01-22
Add classes to Table of Contents based on number of items

pcraig3:
## Summary

This PR adds classes to the ToC section based on how many items are in the table of contents.

The reason for this change is that when the table of contents has too many items, sometimes one item drops onto the next page.

This PR tries to get ahead of that by shrinking the spacing between `<li>` elements if there are too many items in the table of contents.

I did a bunch of testing to arrive where I did, but still don't have a strong sense if this is the right cutoff / CSS to use. We will see in regular use if this works out or needs to be tweaked.

| ToC with 20 items | ToC with 27 items  |
|--------|-------|
|  Has classes: <br>-`toc--normal`<br>-`toc--items--20`       |     Has classes: <br>-`toc--small`<br>-`toc--items--27`   |
|   Screenshot 2025-01-22 at 4 33 24 PM     |    Screenshot 2025-01-22 at 4 33 11 PM   |

The cutoff between `toc--normal` and `toc--small` is 23 items. As in, 23 means `normal` and 24 means `small`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@ Versioning since version 1.0.0.
 
 ### Added
 
+- Add classes to Table of Contents based on number of items
+  - The idea is to shrink the table of contents a bit if lots of items show up
 - Add a new icon for "award" (only for 2 CDC NOFOs for now)
 - Add 2 new images for CDC-RFA-CE-25-0114-b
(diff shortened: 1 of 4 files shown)


pcraig3:
Merging this closes #175

PR #196, opened 2025-01-27
Fix _even more_ heading links

pcraig3:
## Summary

I bet you didn't believe this was possible, but this PR updates our `preserve_heading_links` function to catch _even more_ (!!) heading links that would otherwise show up as "broken" on import. 

| the problem | the HTML |
|--------|-------|
|  3 broken links! (all the same one)      |   `<a>` tag is before the heading, inside of a `<p>` tag    |
|    Screenshot 2025-01-27 at 11 13 22 AM   |   Screenshot 2025-01-27 at 11 14 31 AM    |

## Details

The specific thing that this commit fixes is to preserve heading links when the link target is in a `<p>` tag that _precedes_ a heading, and:

- the `<p>` is empty of text
- the `<p>` only contains `<a>` elements
- those `<a>` elements only have an `id` attribute

Previously, we would preserve links that preceded headings but not if they were wrapped.

```html
<!-- this worked before -->
<a id="_Contacts_&_Support"></a>
<a id="_Contacts_and_Support_1">
<h2>Contacts and Support</h2>

<!-- this works now (in this PR) -->
<p>
  <a id="_Contacts_&_Support"></a>
</p>
<p>
  <a id="_Contacts_and_Support_1"></a>
</p>
<h2>Contacts and Support</h2>
```

This is something I noticed on a couple of recent NOFOs. They had 3 broken links each that were all related to this problem, so since it's something we can solve algorithmically, it seems like something we should do.

These no longer show up as broken because we have fixed them.

### DIFF
--- a/bloom_nofos/nofos/nofo.py
+++ b/bloom_nofos/nofos/nofo.py
@@ -1971,16 +1971,49 @@ def preserve_heading_links(soup):
     This function mutates the soup!
 
     Preserves heading links by transferring IDs from empty <a> tags to their parent heading elements.
-    ALSO preserves heading links by transferring IDs from empty <a> tags directly preceding headings to those heading elements.
+    Also preserves heading links by transferring IDs from empty <a> tags that immediately precede headings
+    (either as standalone elements or as part of a <p> tag containing only <a> elements) to those heading elements.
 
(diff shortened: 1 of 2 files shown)


jtmst (approved):
lg!

Issue #170, opened 2025-01-16
Re-importing a NOFO should create 'archived' past version of NOFO


pcraig3:
## Summary

When someone re-imports a NOFO document, all of the previous sections and subsections are deleted and replaced with completely new sections and subsections.

This has the potential to cause problems — and one time it did, leading us to restore from a database backup to get the old version of the NOFO again.

We don't allow (non-admin) users to completely delete a NOFO, which is why we introduced the concept of "archiving" a NOFO. By the same token, we shouldn't allow people to wipe out all of a NOFO's content, which is akin to what deleting a NOFO actually means.

### Functional requirements

We want this all to be invisible from a user's perspective, which means that we should _not_ archive the current NOFO they are using and give them a new URL. People link to NOFOs by id number, so we should keep their existing URLs working. Instead, let's clone their NOFO, archive it, and then overwrite sections & subsections as we do now.

As context: When re-import is successful, [we create an internal dictionary of `sections` and then call `overwrite_nofo`](https://github.com/HHS/simpler-grants-pdf-builder/blob/main/bloom_nofos/nofos/views.py#L363). This wipes out all existing sections (and subsections).

Instead, let's do something like this:

- Before we `overwrite_nofo`, duplicate our original NOFO
- The duplicated NOFO should immediately be set to "archived"
- _Then_ we call `overwrite_nofo`

Also, we should introduce a new NOFO model field called `ancestor` or something (`int` field, default `null`), which would contain the ID of the archived NOFO. The idea here is that a NOFO should know how to find its previous version. If there are multiple previous versions, we can just pull them sequentially.

This puts us in a way better spot so that we can prevent users from messing things up unintentionally or intentionally.

Issue #156, opened 2025-01-13
Show a warning message if a re-imported NOFO has a different Opportunity Number


pcraig3:
## Summary

We recently had an incident where a NOFO was "reimported" over the wrong NOFO, which means we wiped out the content of the original NOFO. 

That's not good, an obviously something that happened was that the person re-importing didn't get enough of a prompt to say "hey, this might be the wrong NOFO". 

So this story is to add a warning message/extra screen if you are importing a NOFO with a different opportunity number, to just be like "Just so you know, this opportunity number is different". 

Then give them the change to say: 
- yes, this is what I want to do, or
- "[Nevermind (Official Video)](https://www.youtube.com/watch?v=naW6-WxmMiU)"

---

### Tasks

- [x] Refactor the import/re-import route
- [x] Add a new view for confirming re-import if NOFO opportunity number changes
- [ ] @admoorgit says there will be some UI feedback

pcraig3:
#197 went in, so there is code that does this now (see PR for screenshots), 

However, @admoorgit said that he would have some UI feedback on it, so I am keeping it open for now.

pcraig3:
I am closing this issue as completed because the feature works! 

And we have bigger fish to fry. 🐟

PR #197, opened 2025-01-28
Refactor nofo_import: Create "NofosImportNewView" and "NofosImportOverwriteView"

pcraig3:
## Summary

This PR breaks up our `nofo_import` function into 2 class-based views: 

1. NofosImportNewView: for importing a new NOFO
2. NofosImportOverwriteView: for overwriting an existing NOFO

Logic and functionality are unchanged from before, this is strictly a refactor so that sets us up for some of the 'reimport' stories we have in the pipeline (#125, #156, #170).

The diff is not very helpful in this PR. I would suggest just reading through the final `views.py` to understand what's happening.

It is the final PR in the "Refactor nofo_import" series, which included these other ones:
- #191 
- #192 

## Details

For a long time, we were using a function-based view called `nofo_import` which handled both importing a new NOFO and overwriting an existing NOFO. This made sense because there is a bunch of shared logic here: 

- uploading the file
- converting the doc string to HTML
- cleaning up the HTML
- creating sections and subsections

Once we had these sections and subsections, we would create the `Nofo` object. Initially, this was almost identical between the 'new' and 'reimport' routes, but over time they have begun to diverge.

The 2 previous PRs were about making `nofo_import` simpler by encapsulating shared functionality in functions that now live in `nofo.py`. However, [when you split `nofo_import` into 2 functions](https://github.com/HHS/simpler-grants-pdf-builder/pull/197/commits/66b46a0bda846171666077a74ff5d702098d4e65), this still led to a lot of copy+paste code: essentially, the first 25 lines of `nofo_import_new` and `nofo_import_overwrite` was 100% copy+paste, only the second half (the logic around creating a NOFO) was different. 

The solution I landed on was _inheritance_: something I learned in first-year CS. Instead of copying a bunch of code between 2 functions, we can inherit the same import/processing steps and then have each class separately implement the "create NOFO" logic. 

Much cleaner than before, and it also makes our `views.py` file more consistent as a large majority of our other views are class-based views.

### DIFF
--- a/bloom_nofos/nofos/nofo.py
+++ b/bloom_nofos/nofos/nofo.py
@@ -127,6 +127,18 @@ def process_nofo_html(soup, top_heading_level):
     return soup
 
 
+def build_nofo_sections_and_subsections(soup, top_heading_level):
+    """
+    Builds sections and subsections as python dicts, ready to be used to create a new NOFO.
+    Raises ValidationError if no sections found.
+    """
+    sections = get_sections_from_soup(soup, top_heading_level)
+    if not len(sections):
+        raise ValidationError("That file does not contain a NOFO.")
+    sections = get_subsections_from_sections(sections, top_heading_level)
+    return sections
+
+
 ###########################################################
 #################### UTILITY FUNCS ####################
 ###########################################################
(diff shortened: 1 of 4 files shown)


jtmst (approved):
tested locally, looks good

Issue #77, opened 2024-12-02
Use Login.gov for user accounts


pcraig3:
We have been using Django's user models for signups and accounts, but as part of moving to HHS infra, we will need to have better user security (2fa and stuff).

If we use login.gov, it already has that stuff, so it will make our job a lot easier. Plus it just makes sense -- we don't want people to have different accounts they have to log into all the time. 

So this story is about migrating to login.gov for user account management.

pcraig3:
## Getting started with testing Login.gov

To get started with Login.gov, you will need to:

- Create an account here: https://developers.login.gov/testing/#using-the-sandbox
- Once you log in, you should see a "team" that includes a sandbox app called: `Dev Simpler ...`
- Create app
  - To create an app, you will need to generate a [ppublic-private cert](https://developers.login.gov/testing/#creating-a-public-certificate)
  - (See image I shared)
- Read through these docs: 
  - https://github.com/HHS/simpler-grants-gov/blob/main/documentation/api/authentication.md
  - https://github.com/HHS/simpler-grants-gov/blob/main/OPERATIONS.md#yearly-rotations

pcraig3:
This is half done, but can't be completed until we finish the migration to AWS.

PR #200, opened 2025-01-29
Feat/login dot gov integration

jtmst:
Issue: https://github.com/HHS/simpler-grants-pdf-builder/issues/77

Adds integration with login.gov for auth.

Requires .env vars added, and login-gov-private.pem cert file added to `bloom_nofos/bloom_nofos/certs`

Launch checklist:
https://developers.login.gov/production/#confirm-interagency-agreement-iaa

NOTE:
This PR has been edited to allow for the traditional django email/password login to be used in addition to the login.gov flow.

Screenshot: 
image


bloom_nofos/users/templates/users/login.html
```diff
@@ -5,36 +5,65 @@
...
+    <a href="{% url 'users:login_gov' %}" class="usa-button usa-button--outline width-full">
+        Sign in with Login.gov
+    </a>
+
+    <div class="margin-top-2 text-center">
+        <small class="text-base-dark">
+            Don't have a Login.gov account?
+            <a href="https://secure.login.gov/sign_up/enter_email" target="_blank" rel="noopener noreferrer">
+                Create one here
+            </a>
```
pcraig3, line 65: Should we use this URL? `https://idp.int.identitysandbox.gov/sign_up/enter_email`

That's the one that will actually work with our login.gov integration

bloom_nofos/bloom_nofos/settings.py
```diff
@@ -240,7 +240,7 @@
 # Users
 
 AUTH_USER_MODEL = "users.BloomUser"
-LOGIN_URL = "/login"
+LOGIN_URL = "users:login"
```
pcraig3, line 243: Seems like we can delete this since it's duplicated below (line 526)

bloom_nofos/users/auth/login_gov.py
```diff
@@ -0,0 +1,159 @@
...
+            public_key = self._get_login_gov_public_key(kid)
+            if not public_key:
+                raise ValueError("Could not get Login.gov public key")
+
+            decoded = jwt.decode(
+                id_token,
+                public_key,
+                algorithms=["RS256"],
+                audience=self.client_id,
+            )
```
pcraig3, line 151: Should we also add `issuer="https://google.com",` so that we make sure these requests come from the right login.gov url?

This should be the same value as `LOGIN_GOV_OIDC_URL`.

bloom_nofos/users/templates/users/login.html
```diff
@@ -5,36 +5,65 @@
...
+<div class="messages">
+    {% for message in messages %}
+    <div class="usa-alert usa-alert--{{ message.tags }}">
+        <div class="usa-alert__body">
+            <p class="usa-alert__text">{{ message }}</p>
+        </div>
+    </div>
+    {% endfor %}
+</div>
+{% endif %}
```
pcraig3, line 30: Let's move this above the H1

bloom_nofos/users/views.py
```diff
@@ -66,3 +71,103 @@ def post(self, request):
...
+    return redirect(settings.LOGOUT_REDIRECT_URL)
+
+
+def traditional_login_view(request):
+    """Handle traditional login with username/password."""
+    form = LoginForm(request.POST or None)
+
+    if request.method == "POST":
+        if form.is_valid():
+            email = form.cleaned_data["email"]
```
pcraig3, line 162: Let's make sure we lowercase the emails people use: `email = form.cleaned_data["email"].strip().lower()`

Some HRSA emails are sent to me like `PCraig@hrsa.gov`, but I always enter them as lowercase.

pcraig3 (commented):
Hello! 

Overall this looks good, I left a few comments but they are minor.

I think we should also surface the "login_gov_user_id" in the admin (both on the listing page and on the user profile page).

I think we should also talk about if we need a 'staging' instance and if the `pem` file is good where it is.

bloom_nofos/users/auth/login_gov.py
jtmst: Yes, good call. added this

bloom_nofos/users/auth/login_gov.py
```diff
@@ -0,0 +1,157 @@
...
+            raise Exception(
+                "Private key not configured. Cannot create client assertion."
+            )
+
+        now = int(time.time())
+
+        try:
+            private_key = load_pem_private_key(
+                self.private_key, password=None, backend=default_backend()
+            )
```
pcraig3, line 105: This is also called in `__init__` -- we could store the result of `load_pem_private_key` in there and save a couple lines of code in there, unless there's a reason not to.

Small thing though, not sure it matters very much.

pyproject.toml
```diff
@@ -29,8 +29,10 @@ django-mirror = "^0.1.13"
 mammoth = "^1.7.1"
 tomli = "^2.0.2"
 django-ninja = "^1.3.0"
+cryptography = "^42.0.0"
 
-# sortable = "@3", check nofo_index.html
```
pcraig3, line 33: Can I leave this here? 

It's a dependency we have in JS (we are loading it from a CDN) and I didn't want to lose track of it so I added this commented line here. 

Here's the line we care about: https://github.com/HHS/simpler-grants-pdf-builder/blob/main/bloom_nofos/nofos/templates/nofos/nofo_index.html#L97

pcraig3 (changes requested):
Mostly looks good, but I am running into a large delay getting the public key from GCP.

Here's what I am seeing for time:

```
def get_login_gov_keys(environment: str = "dev"):
    """Gets Login.gov keys from Secret Manager."""
    last_time = timer()
    private_key = get_secret(f"login-gov-private-key-{environment}")
    last_time = timer(last_time) // 1.393965 seconds
    public_key = get_secret(f"login-gov-public-key-{environment}")
    last_time = timer(last_time) // 64.145196 seconds
    return private_key, public_key
```

It think the solution for this might be easy enough though: can't we just not use the public key in our app? My understanding is that the public key is uploaded to Login.gov (under "Public certificates") and then we retrieve it from there whenever we are actually doing the sign-in. Therefore, the only key we need in our app is the private key, yes?

If that's wrong, then let's discuss! I could be wrong for sure, this is not my area of expertise.

Second comment, which might not be relevant anymore, but I think we should be calling the public key a `crt` file, as per this answer: https://stackoverflow.com/questions/63195304/difference-between-pem-crt-key-files

When I generated my keys previously, it gave me 2 files: `public.crt` and `private.pem`, so I think that's what we're doing here as well.

bloom_nofos/bloom_nofos/context_processors.py
```diff
@@ -13,3 +13,8 @@ def add_docraptor_live_mode(request):
 
 def add_github_sha(request):
     return {"GITHUB_SHA": settings.GITHUB_SHA}
+
+
+def settings_context(request):
+    """Makes selected Django settings available to all templates."""
+    return {"settings": settings}
```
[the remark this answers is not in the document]
jtmst: Yep thats right, allows us to set enabled based on whether the keys are successfully retrieved and conditionally render our login.gov component based on that

pcraig3 (approved):
Awesome work, thanks Josh!

PR #202, opened 2025-01-30
Add "Are you sure?" confirmation page when re-importing a NOFO with a different ID

pcraig3:
## Summary

This PR does 1 major thing and 1 minor thing:

1. Add a confirmation page when re-importing a NOFO with a different ID
2. Suppress the "DELETE" warning in our test suite

### 1. Add a confirmation page when re-importing a NOFO

This is based on a post-mortem we had a couple of weeks ago. Basically, someone re-imported a NOFO document _for a different NOFO_ and then wiped out a bunch of content. 

In the event, we were able to restore from a backup, but it signalled to us that we should take some action around this.

This PR introduces a new view (`NofosConfirmReimportView`) which shows up when you upload a document to re-import but the document is found to have a different NOFO opportunity number, which suggests you may have the wrong document. At this point, you still _can_ reimport if you want, or you can back out of it and return to the original reimport screen. 

A couple things to note:
- Because we have to pass the NOFO data to a new view between requests, we are using `request.session` for the first time. This means the session data needs to be removed when we get the chance, so we do that on get requests to the "nofo_edit" page. 
- I am trying to keep things DRY here so I created some new static methods in my View classes because `NofosConfirmReimportView` has to essentially do all the same steps as the Overwrite view to overwrite the NOFO object.

| current import page | reimport page |
|--------|-------|
|  This is the current page we have in the app for reimporting      | This is the *new* page that is added in this PR      |
|     Screenshot 2025-01-29 at 12 29 28 PM   |  Screenshot 2025-01-29 at 12 29 37 PM     |

### 2. Suppress the "DELETE" warning in tests

One of our tests tries to "archive" a published NOFO, and this is not allowed so it throws a warning. Annoyingly, it also meant we would see a warning when running our test suite. 

Now we don't.

| before | after |
|--------|-------|
|    Screenshot 2025-01-30 at 11 50 08 AM    |    Screenshot 2025-01-30 at 11 50 42 AM   |

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ Versioning since version 1.0.0.
 - Add cover image for CDC-RFA-CK-25-0125 and CDC-RFA-JG-25-0178
 - Add inline images for CDC-RFA-CE-25-0114
 - Can use 'page-break' to add a page-break in markdown body
+- Add confirmation page for re-importing a NOFO if the IDs don't match up
 
 ### Changed
 
--- a/bloom_nofos/nofos/templates/nofos/nofo_confirm_delete.html
+++ b/bloom_nofos/nofos/templates/nofos/nofo_confirm_delete.html
@@ -29,15 +29,14 @@
 
   <form id="nofo-delete--form" method="post">
     {% csrf_token %}
-    <p>Are you absolutely sure you want to delete “{% if nofo.short_name %}{{ nofo.short_name }}{% else %}{{ nofo.title }}{% endif %}”?</p>
-    {{ form }}
+    <p>Are you absolutely sure you want to delete “{{ nofo|nofo_name }}”?</p>
 
     <ul class="usa-button-group margin-top-3">
       <li class="usa-button-group__item">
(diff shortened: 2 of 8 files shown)


admoorgit:
@pcraig3 I will work with Jackie on a clearer/simpler way to present the differences. In practice, this does what it should be doing. I think we can do a better job on how it's displayed (naturally).  If it's easier to commit and fix it later, have at. Otherwise we'll try to get you an updated version first thing next week

jtmst (approved):
typo spotted, otherwise looks good. I dig the session data based approach

pcraig3:
> @pcraig3 I will work with Jackie on a clearer/simpler way to present the differences. In practice, this does what it should be doing. I think we can do a better job on how it's displayed (naturally). If it's easier to commit and fix it later, have at. Otherwise we'll try to get you an updated version first thing next week

I am going to merge, for 2 major reasons:

1. Keeping code around in a PR sucks
2. Once it's up, you and Jackie will be able to see it yourselves, not just in this screenshot

PR #205, opened 2025-02-04
Ignore login.gov directories and files

pcraig3:
## Summary

This is a very small PR that is all about DX.

Basically, since the Login.gov PR (#200) is up, I have had this `bloom_nofos/bloom_nofos/certs` directory in my app and it is constantly getting flagged by git as a new file on all other branches. Very annoying!

We can't exactly merge in the login.gov stuff yet, but we _can_ merge in this change to the `.gitignore` file and it will make switching branches seamless.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,3 +19,8 @@ debug_output.html
 
 # export file
 export_links.csv
+
+# Login.gov certificates
+login-gov-cert*.pem
+certs/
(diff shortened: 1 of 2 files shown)


pcraig3:
Should be an easy one @jtmst

jtmst (approved):
good call

PR #207, opened 2025-02-05
Add DOCRAPTOR_IPS to Django Constance

pcraig3:
## Summary

This PR converts the `DOCRAPTOR_IPS` env var to a constance variable that is editable in the admin backend by superusers. 

We need DocRaptor to be able to access our NOFO pages so that it can request them, load them, and then give us back a PFD document. We use the IPs to identify these incoming requests, and DocRaptor provides a list of IPs that their requests come from.

Sometimes, DocRaptor's IPs change, and then when the NOFO Builder asks for a printout, we end up getting a PDF of our login because the request was not whitelisted. 100% of the time this happens, it's discovered by someone printing a NOFO. _Usually_, it is also someone who has superuser access. 

This PR gives NOFO Builder production staff the ability to solve this problem for themselves by grabbing the latest IP addresses and updating them in the system when they notice a "login" pages getting printed. No more waiting for a code deploy!

### Screenshot

Here is what the interface looks like for this:

Screenshot 2025-02-05 at 3 50 49 PM

Only superusers have access to this value, not any user with an account in the NOFO Builder.

### Format string

A list of IPs can be passed in in a variety of formats:

- comma-separated (default): `18.233.48.178,18.235.199.18,23.20.110.13`
- space-separated: `18.233.48.178 18.235.199.18 23.20.110.13`
- newline separated: 
```
18.233.48.178
18.235.199.18
23.20.110.13
```

You can also combine any of these, eg: `18.233.48.178, \n18.235.199.18, \n23.20.110.13`


jtmst (approved):
Looks good to me.  We could consider some kind of validation for IPs to prevent mis-inputs, but optional and not merge blocking

pcraig3:
Approved! Merging!

PR #208, opened 2025-02-07
Add a "modifications" status to published NOFOs

pcraig3:
## Summary 

Adding a 'modifications' property to published NOFOs.

Published NOFOs sometimes need to be updated in-flight (otherwise they have to be taken down and re-published, and then the application process completely restarts), but the NOFO Builder doesn't really have a concept for this. 

This PR introduces the notion of a `modifications`. When you 'Add modifications' to a NOFO, this is what happens:

- There is a 'modified' message on the cover page
- There is a new setting in the NOFO edit page allowing you to set a modifications date
- A "Modifications" section is added to the end of your NOFO*
  - This section shows up in the table of contents, but it does not have a section page
  - It comes with 1 subsection: a table for you to list changes and the date of those changes

*If there is already a "Modifications" section in your NOFO, this won't get added.

## Some notes:

- Modifying a NOFO is 1-way: a modified NOFO can't be un-modified except by a superuser
- Technically, a NOFO of _any_ status can be modified. The interface only shows you the option to modify a NOFO once it is published, but there's no rule for this in the code.
- All published NOFOs can be modified. Since this is a feature only for HRSA, we could restrict it to HRSA NOFOs.
- Mechanically, the `modifications` field is exactly like the `archived` field. It holds a `Date`, which we use in 2 ways. (1) If it is set at all, we know this NOFO has been modified, and (2) we use the date on the 'modified' on the cover page.

Also, as a final observation, I wrestled with whether `modifications` shouldn't just be a string. If it was a string, we could just let people write their own modifications message. In the end, I decided against it because we _do_ want to add a data _and_ a link, and if we let people have a completely open string, they could potentially forget/ignore one of those. So making it a date allows us to hardcode how the message on the cover page looks.

###  published nofo 'edit page' overview

| before | after |
|--------|-------|
|    "published" nofo : has button to add modifications  |  "published" nofo with modifications: explains to update "Modifications" section     |
| Screenshot 2025-02-07 at 9 38 45 AM       |   Screenshot 2025-02-07 at 9 39 32 AM |

###  cover page

| before | after |
|--------|-------|
|    "published" nofo : normal cover page  |  "published" nofo with modifications: cover page includes "modifications" message     |
|   Screenshot 2025-02-07 at 9 38 54 AM    |   Screenshot 2025-02-07 at 9 41 13 AM  |

### customize date on the cover page "modifications" message

| before | after |
|--------|-------|
|  Added new edit row for customizing the modifications date      |   Page for customizing modifications date includes [USWDS datepicker](https://designsystem.digital.gov/components/date-picker/)    |
|   Screenshot 2025-02-07 at 9 40 00 AM     |  Screenshot 2025-02-07 at 9 40 22 AM     |

### "Modifications" section

| before | after |
|--------|-------|
|  edit page "Modifications" section      |   HTML view "Modifications" section    |
|    Screenshot 2025-02-07 at 11 26 08 AM    |   Screenshot 2025-02-07 at 9 41 34 AM    |

## gif

Screen Recording 2025-02-07 at 3 50 04 PM


jtmst (approved):
works great locally.  lgtm!

pcraig3:
We are merging this now that people are starting to need it. 

There's been no strong opinions about it so far, so I think we will just have to wait and see what people do with it.

Issue #194, opened 2025-01-24
Improve the external link checker by retrying failed requests


pcraig3:
We are using HEAD but sometimes a regular GET would work.

The idea here is we will get less false negatives if we retry all the 500 errors again as regular GET requests.

PR #209, opened 2025-02-07
Feat/retry in link checker

jtmst:
Issue: https://github.com/HHS/simpler-grants-pdf-builder/issues/194

Changes:
- Adds get request to be attempted for link checker if head request fails
- Adds more detailed request headers for get requests
- Unit tests to verify this logic

QA Steps:
- Upload a nofo that has a bunch of external links
- Run the link checker
- It should mostly return accurate results for good links/broken links*

* NOTE: Some link sources like eCFR are flaky, especially when running the checker multiple times in a row, I suspect they have some rate limiting/throttling/bot detection in place thats going to results inconsistent

Screenshots:
BEFORE CHANGES:  
image

AFTER CHANGES:
image

### DIFF
--- a/bloom_nofos/nofos/nofo.py
+++ b/bloom_nofos/nofos/nofo.py
@@ -28,13 +28,9 @@ from .utils import (
 )
 
 DEFAULT_NOFO_OPPORTUNITY_NUMBER = "NOFO #999"
-# Use Firefox user agent
-REQUEST_HEADERS = {
-    "User-Agent": "Mozilla/5.0 (Macintosh; Intel Mac OS X 10.15; rv:131.0) Gecko/20100101 Firefox/131.0"
-}
 
-# More realistic browser like request headers for GET requests
-REQUEST_HEADERS_GET = {
+# More realistic browser-like request headers (Firefox) for HTTP requests
+REQUEST_HEADERS = {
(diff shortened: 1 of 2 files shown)


pcraig3:
I was curious so I ran some tests locally for 3 scenarios:

- just the HEAD request (current code)
- just a GET request (if we used GET instead of HEAD)
- this PR (HEAD, then GET)

I ran it on a NOFO I have locally with 78 external links. Here's what I found:

## Just `HEAD` request 

| bad links | duration |
|-----------|----------|
|   21        |     16.2767 s     |
|      31     |    22.5957 s      |
|      31     |   21.9012 s       |

## Just `GET` request 

| bad links | duration |
|-----------|----------|
|   17        |     17.5743 s     |
|      19     |    16.9996 s      |
|      24     |   18.8653 s       |

## `HEAD` request, then `GET` request 

| bad links | duration |
|-----------|----------|
|   19        |     17.5089  s     |
|      19     |    17.4745 s      |
|      19     |    17.1278 s       |

pcraig3:
There is definitely something weird going on here. I think we hit some rate limit probably and then we start getting rejected by requests that might have worked the first time. 

In any case, it looks like this change helps and doesn't add too much latency, let's ship it!!! 🚢

pcraig3 (approved):
👍

PR #211, opened 2025-02-13
Compare NOFOs including text diff

pcraig3:
## Summary

- This is related to https://github.com/HHS/simpler-grants-pdf-builder/issues/125,
- It's also the continuation of #144, which has been closed.

### What is it?

The idea here is a UI concept for comparing a NOFO to another NOFO document. (The more expansive concept is to add a step like this on re-import, but I think that "compare" is a big enough feature on its own and something that we can ship and get feedback on before we make this part of the re-import flow.) 

The objective is to provide a simple way for people to identify changes that have happened since their last import, so they can see what the damage would be if they re-imported and overwrote what they currently have.

Only the NOOF attributes and subsections that have changed are shown on the comparison page. This means attributes/subsections that are:
- modified (text is changed)
- added (did not exist previously)
- deleted (used to exist, now doesn't) 

You can get to the "compare" screen from the reimport screen. I added a link using the little [Tag](https://designsystem.digital.gov/components/tag/) banner and ["slim" alert](https://designsystem.digital.gov/components/alert/) so that hopefully people will notice it. The little "BETA" message follows you throughout the compare flow since this is new and different.

In terms of a feedback form, I added a link to a Google Form in the "Beta" banner, and also at the very bottom of the "compare" page. It's kind of crappy sending people off-site, but I suspect the form will probably be used by 1 or 2 people max, so it's easy enough. Added a new template partial so that the link to the form is in 1 place and not 3 places.

### How to test this
- Make a "save as" version of an existing NOFO word doc you have.
- Add some changes to the new doc (slash add or remove some subsections)
- Import the first Word doc to create a new NOFO
- Go to the "re-import" page for your new NOFO
- Click the link in the "Beta" banner
- Import your "save as" version of the NOFO
- Be amazed

### screens 

| 're-import' screen | 'import' screen | 'diff' screen |
|--------|-------|-------|
|  Screenshot 2025-02-17 at 2 17 28 PM   |   Screenshot 2025-02-19 at 2 54 02 PM  |   localhost_8000_nofos_485_compare    |

### gif

Here's a gif of me actually doing it.

Screen Recording 2025-02-17 at 2 23 12 PM

## TODO

~~@jtmst, I am _pretty_ sure we can use this "is_matching_subsection" method in your preserve_page_breaks stuff, but you would be more familiar with that. The idea is that you should be able to grab a new subsection from a different nofo and compare to an existing one, using a function call like this: `old_subsection.is_matching_subsection(new_subsection)`. It is not 100% since it doesn't do a comparison of text, but it should be close enough, and if we can delete some LoC, all the better.~~

Nvm, I looked at the code for this and it is approaching this problem in a different way.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@ Versioning since version 1.0.0.
 
 ### Added
 
+- Add new "{id}/compare" route for comparing a new NOFO document to an existing one
+
 ### Changed
 
 - Breadcrumb links are no longer clickable (instead, they are just visual indicators)
--- a/bloom_nofos/bloom_nofos/static/styles.css
+++ b/bloom_nofos/bloom_nofos/static/styles.css
@@ -966,7 +966,8 @@ details.usa-accordion .usa-accordion__content ul li:not(:last-of-type) {
 
 /* compare page */
 
-.usa-alert--info.usa-alert--alpha .usa-alert__body {
+.usa-alert--warning.usa-alert--alpha .usa-alert__body {
+  border-left: 2px solid black;
   background-color: #f0f0f0;
 }
 
--- a/bloom_nofos/bloom_nofos/static/styles.css
+++ b/bloom_nofos/bloom_nofos/static/styles.css
@@ -966,6 +966,10 @@ details.usa-accordion .usa-accordion__content ul li:not(:last-of-type) {
 
 /* compare page */
 
+.usa-alert--info.usa-alert--alpha .usa-alert__body {
+  background-color: #f0f0f0;
+}
+
 del {
   color: red;
   background-color: #ffebe9;
--- a/bloom_nofos/bloom_nofos/static/styles.css
+++ b/bloom_nofos/bloom_nofos/static/styles.css
@@ -297,6 +297,10 @@ details > summary > span:hover {
   width: 132px;
 }
 
+.nofo_import.nofo_import--compare .submit-button {
+  width: 165px;
+}
+
 .nofo_import .submit-button.submit-button--loading {
   text-align: left;
 }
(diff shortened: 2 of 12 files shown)


admoorgit:
@pcraig3 agree with adding a link to the re-import page. Maybe a text link under the import button that says "Test the compare feature [WIP/BETA/TEST/WHATEVER]." (or something similar) that pops em over to it.

Links out after viewing compare:
- Button that points to re-import page so they can then re-import the word doc. Now that they've seen the differences, they'll know what to look for and can reimport as usual. Safe to assume people are accessing the re-import page to actually re-import, and MAY click on the compare out of curiosity, but they'll ultimately want to do the task they set out to do.
- Could we add a quick lil form/entry field to submit feedback? Maybe something we can implement a little more fully moving forward to gather input + feedback from NOFO Builder users.

jtmst (approved):
This looks great, tested locally with a handful of nofos and didn't encounter any major issues.  

Only thing I noticed is that the contrast on the del css rules doesnt meet typical contrast guidelines, chrome recommended something like:

`del {
    color: #a40000;
    background-color: #ffebe9;
}`

But anything higher contrast would do

Thats likely non-launch-blocking, approved!

admoorgit:
@pcraig3 wonder if we can make that red darker so it meets contrast standards? We should def address this before we roll out out to, say, HRSA.

pcraig3:
Will work on this today + a feedback form!

pcraig3:
> I noticed is that the contrast on the del css rules doesnt meet typical contrast guidelines, chrome recommended something like:
> 
> `del { color: #a40000; background-color: #ffebe9; }`
> 
> But anything higher contrast would do


This has been fixed! I am using [`text-secondary-dark`](https://designsystem.digital.gov/utilities/color/) now, which is the same as we use for error messages. No more contrast issue.

pcraig3:
This has gone through review, so I am merging it so that it is ready for today's HRSA meeting.

PR #212, opened 2025-02-13
Migration to add public info subsection to existing NOFOs

jtmst:
Mimics the logic in add_final_subsection_to_step_3 but for a migration to add the PUBLIC_INFORMATION_SUBSECTION to existing NOFOs.


pcraig3:
Made some changes to it:

- do not add to published nofos or archived nofos
- add an `html_id` because it won't get added otherwise

pcraig3 (approved):

PR #214, opened 2025-02-18
Nullify clickable "top-nav" breadcrumb links on every page

pcraig3:
## Summary

This PR makes two changes to our final NOFO Builder PDF output:

- No longer have clickable top-nav breadcrumb links
- When opening in Acrobat, the Bookmarks tab will be open by default

The top-nav links being clickable was super annoying for screen readers, as they would be constantly reading all those links on every single page of the NOFOs, and there was no way to prevent this.

In the end, we decided that the best way forward is to leave the links in there so that they signpost where the user is in the NOFO (eg, it will indicate they are in section 3, for example), but to nullify them so they are no longer clickable.

To offset this, we also always want the Bookmarks tab to be open in Adobe, so that the structure of the NOFO is visible and navigable using in-app tools.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,9 @@ Versioning since version 1.0.0.
 
 ### Changed
 
+- Breadcrumb links are no longer clickable (instead, they are just visual indicators)
+- When you open a NOFO Builder PDF in Acrobat, the Bookmarks tab will be open by default
+
 ### Fixed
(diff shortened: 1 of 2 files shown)


jtmst (approved):

admoorgit:
Just noting I tested the PDF you sent @pcraig3 and everything worked as expected.

PR #215, opened 2025-02-19
Adds fallback to section and subsection queries

jtmst:
Adds a fallback for cases where there is no section or subsection data for whatever reason.  This should at least avoid our 500 error. Also convert the obeject_id to a string before making the query (https://github.com/soynatan/django-easy-audit/blob/d2941373dd2107bc7994884cb044fcd2af2bb086/easyaudit/models.py#L35)


pcraig3 (approved):
Good catch!

PR #218, opened 2025-02-21
Improvements to the diff interface

pcraig3:
## Summary 

This PR makes a bunch of changes to the diff interface. If you thought it was good before, prepare to be WOW'd!!

 1. Use full-word diffs
 2. Show linebreaks that exist in body content (easier for lists, for example)
 3. Reclassify whitespace-only changes as "MATCH" not "UPDATE" (in markdown, they generally don't matter)
 4. Hide the "Basic information" subsection because it duplicates information

Here is a big before/after screenshot (new screenshot is much shorter), but each of these is explained more before.

| before | after |
|--------|-------|
|     localhost_8000_nofos_405_compare (1)   |   localhost_8000_nofos_405_compare    |

###  1. Use full-word diffs

Since "March" and "January" both have "ar" in the middle, we would get
really weird looking diffs when we deleted "January" and added March":

- `<del>Janu</del><ins>M</ins>ar<del>y</del><ins>ch</ins>`

Very weird and confusing.

Now it is like this:

- `<del>January</del><ins>March</ins>`

Clean!!

| before | after |
|--------|-------|
|   very hard to read     |   still slightly weird showing that we removed "1" and added "14", but generally much more readable     |
|    Screenshot 2025-02-21 at 4 02 20 PM    |    Screenshot 2025-02-21 at 4 02 48 PM   |

### 2. Show linebreaks that exist in body content (easier for lists, for example)

When we have list elements in the body text, they were represented in one big glob before, which doesn't really give people a true representation of that content. Same idea with content that is multiple paragraphs broken up.

So now we show the linebreaks to give more of a true rendition of how that content is laid out.

| before | after |
|--------|-------|
|    Screenshot 2025-02-21 at 4 26 09 PM    |     Screenshot 2025-02-21 at 4 25 30 PM  |

### 3. Reclassify whitespace-only changes as "MATCH" not "UPDATE"

We were seeing this kind of diff with the linebreaks after we saved it:

```
'This program is authorized under:<del>\r\n\r</del><ins>\n</ins>\n* The United States Leadership Against HIV/AIDS, Tuberculosis and Malaria Act of 2003, Public Law 108-25 (22 U.S.C. 7601, et seq.).<del>\r</del>\n* The Tom Lantos and Henry J. Hyde United States Global Leadership Against HIV/AIDS, Tuberculosis, and Malaria Reauthorization Act of 2008, Public Law 110-293.<del>\r</del>\n* The PEPFAR Stewardship and Oversight Act of 2013, Public Law 113-56.<del>\r</del>\n* Public Health Service Act (42 U.S.C. 242I).<ins>\n</ins>'
```

The part we care about is this: `<del>\r\n\r</del><ins>\n</ins>`.

Basically, this is the same change, just whitespace. So this would show up as an "update" when nothing material had changed. Now, we ignore stuff like this.

| before | after |
|--------|-------|
| says "UPDATE" but the only visible changes are whitespace-related |  N/A   |
|    Screenshot 2025-02-21 at 4 04 20 PM    |     N/A   |

### 4. Hide the "Basic information" subsection because it duplicates information

The very first subsection is just metadata for the NOFO, so we don't consider it body content like everything else. All the changes in this section should be captured in the "Basic information" table above. 

This is what we do in the [`nofo_view`](https://github.com/HHS/simpler-grants-pdf-builder/blob/main/bloom_nofos/nofos/templates/nofos/nofo_view.html#L383) template as well. 

| before | after |
|--------|-------|
|    localhost_8000_nofos_405_compare (1)   |    localhost_8000_nofos_405_compare   |

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,11 @@ Versioning since version 1.0.0.
 
 ### Fixed
 
+- Fixed a bunch of things in the diff interface:
+  - Use full-word diffs, so `<del>January</del><ins>March</ins>` instead of `<del>Janu</del><ins>M</ins>ar<del>y</del><ins>ch</ins>`
+  - Show linebreaks that exist in body content (easier for lists, for example)
+  - Reclassify whitespace-only changes as "MATCH" not "UPDATE" (in markdown, they generally don't matter)
+  - Hide the "Basic information" subsection because it duplicates information
 - Fix for more list styles that were not importing correctly ("List Bullet1, Bullet 2", etc)
 
 ## [2.4.0] - 2025-02-13
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ Versioning since version 1.0.0.
 
 ### Changed
 
+- Use a MS 365 form for feedback instead of a Google Form
 - Breadcrumb links are no longer clickable (instead, they are just visual indicators)
 - When you open a NOFO Builder PDF in Acrobat, the Bookmarks tab will be open by default
 
--- a/bloom_nofos/bloom_nofos/static/styles.css
+++ b/bloom_nofos/bloom_nofos/static/styles.css
@@ -1005,7 +1005,6 @@ ins {
   background-color: #aceebb;
 }
 
-del > br,
-ins > br {
+del + ins > br {
   display: none;
 }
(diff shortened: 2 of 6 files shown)


jtmst (approved):
Consider me wow'd

PR #96, opened 2024-12-10
Add new subsection "Important: public information" at the end of step 3

pcraig3:
## Summary

This PR adds a new subsection to the end of step 3 with the following text:

```
Important: public information

When filling out your SF-424 form, pay attention to Box 15: Descriptive Title of Applicant's Project.

We share what you put there with [USAspending](https://www.usaspending.gov/). This is where the public goes to learn how the federal government spends their money.

Instead of just a title, insert a short description of your project and what it will do.
```

This will be added to the end of any section named "Step 3: Prepare Your Application" or Step 3: Write Your Application".

Once this goes in, it will be added to every uploaded NOFO from now on. This is what we want, right, @admoorgit?

### screenshot

| editing view | html view  |
|--------|--------|
|  this new section in the editing view      |  new section in the HTML view      |
| Screenshot 2024-12-10 at 10 33 20 AM | Screenshot 2024-12-10 at 10 34 03 AM |

Merging this PR closes #88

PR #219, opened 2025-02-26
Add "Important" subsection after specific headings in step 3

pcraig3:
# Summary

This PR adds logic to the function that inserts a new subsection inside of "Step 3": https://github.com/HHS/simpler-grants-pdf-builder/pull/96

In particular, we add logic so that the new subsection can be inserted after specific headings, not just at the very end of Section 3. 

The headings we are looking for are:

- Other required forms
- Standard forms
- Application components

If multiple of these exist, the new subsection will be inserted after the first one we find. If none of them exist, the new subsection would be inserted at the end of section 3 (which is the current behaviour).

## Screenshots

| "Other required forms": before | "Other required forms": after |
|--------|-------|
|   Screenshot 2025-02-26 at 4 13 14 PM     |    Screenshot 2025-02-26 at 4 12 29 PM   |

| "Standard forms": before | "Standard forms": after |
|--------|-------|
| Screenshot 2025-02-26 at 4 11 36 PM |   Screenshot 2025-02-26 at 4 11 00 PM |

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,8 @@ Versioning since version 1.0.0.
 - Use a MS 365 form for feedback instead of a Google Form
 - Breadcrumb links are no longer clickable (instead, they are just visual indicators)
 - When you open a NOFO Builder PDF in Acrobat, the Bookmarks tab will be open by default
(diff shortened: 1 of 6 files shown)


pcraig3:
Adam has said in Slack that we can merge, so I am going to merge.

PR #232, opened 2025-03-06
add public certs to git

jtmst:
adds public certs to git. private/pem files are still excluded


pcraig3 (approved):
👍 

I don't know why we didn't do this to begin with.

PR #236, opened 2025-03-07
Add green buttons for St. Patricks' Day 🍀 

pcraig3:
## Summary

This is huge, we have added a new button colour.

Previously, I have stuck with the USWDS button colours but they can be a bit limiting, so now I have added this new one.

Used colours that are already part of USWDS.
(Look for ".usa-icon-list--success-dark .usa-icon-list__icon")

Added for these states:

- normal
- hover
- active

"disabled" already has CSS, so all good there.

### screens

| before | after |
|--------|-------|
| default button colour | lucky! 🍀 button colour |
|    Screenshot 2025-03-07 at 3 10 21 PM    |    Screenshot 2025-03-07 at 3 10 01 PM   |

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,13 +9,18 @@ Versioning since version 1.0.0.
 
 ### Added
 
+- Added "Login.gov" user attributes to admin page
+
 ### Changed
 
+- Green button colour 🟢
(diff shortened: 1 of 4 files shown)


jtmst (approved):
🟢

PR #242, opened 2025-03-10
Use Office 365 for our feedback form, not the Airtable form 

### LINKED ISSUE #241: NOFO Builder is using the wrong feedback form
### Task name

Let's change to use a Github Form or a MS form

### Type of task

General suggestion

### Description

We're trying to remove airtable, and so Github forms might be the way forward.

### Upload an image

Image

### Your email

paul@pcraig3.ca

pcraig3:
## Summary

This PR is very small: it gets rid of the link to our Airtable feedback form and replaces it with a Microsoft form.

Once this is in, we are 100% off airtable for good.

| before | after |
|--------|-------|
|  Airtable form we used previously      |  Office 365 form we are using now      |
|   Screenshot 2025-03-10 at 2 07 57 PM     |   Screenshot 2025-03-10 at 2 08 16 PM    |

Merging this closes #241


admoorgit:
@pcraig3 I can't see from the screen shots, but does the new ms form have the same fields (outside of upload) as the air table did?

pcraig3:
> @pcraig3 I can't see from the screen shots, but does the new ms form have the same fields (outside of upload) as the air table did?

There are some differences:

- no image upload field (we have discussed this already)
- combined "name" and "description" because they are probably the same thing for someone submitting
- added "name" and "email" fields for contact information

pcraig3:
Merging since this was approved over Slack.

Issue #268, opened 2025-03-25
Add column width classes to HTML tables like we do with markdown tables


pcraig3:
Writers are often confused how to add/adjust table widths using HTML, so one thing we can do is add classes by default like we do with markdown tables.

This will achieve 2 things: 

- standardize our approach to tables (so all 3 column tables start with the same widths (kind of))
- give nofo editors an example of how to add classnames

👍

pcraig3:
This is closed since earlier today.

PR #269, opened 2025-03-26
Add default column classes to HTML tables 

pcraig3:
## Summary

This PR adds default column width classes to HTML tables. This means we will have parity between markdown tables and HTML tables. 

The idea is, for tables with x columns, we add classes to make them equal width:

- 3 columns: `w-33`
- 4 columns: `w-25`
- 5 columns: `w-20`

Once your table is more than 6 columns, I am assuming it's too weird and I am not going to add presumptive classnames to your table.

Previous commit is here: https://github.com/HHS/simpler-grants-pdf-builder/commit/335d73829b7b5480608d0db72a7f8bb6663fc501

### Screenshots

| before | after |
|--------|-------|
|   second table has uneven column widths     |  both tables look the same 👯      |
|    Screenshot 2025-03-26 at 10 45 56 AM    |   Screenshot 2025-03-26 at 10 46 05 AM    |

### DIFF
--- a/bloom_nofos/nofos/nofo_markdown.py
+++ b/bloom_nofos/nofos/nofo_markdown.py
@@ -2,6 +2,46 @@ from bs4 import BeautifulSoup
 from markdownify import MarkdownConverter
 
 
+def get_width_class(th):
+    def _get_num_columns_th(th):
+        return len(th.parent.find_all("th"))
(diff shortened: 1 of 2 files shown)


pcraig3:
Merging this resolves #268

Issue #224, opened 2025-03-04
Make sure width CSS for tables is reflected everywhere


pcraig3:
## Summary

Sometimes table columns need to be readjusted to try to equalise the length of a table. For example, some tables have lots of content in the first column, and less in the remaining columns (see example)

| before (no column width adjustments) | after (with column with adjustments) |
|--------|-------|
|   <img width="561" alt="Image" src="https://github.com/user-attachments/assets/85d4a4a2-5954-4e34-99e3-0066e9e3c553" />     |     <img width="561" alt="Image" src="https://github.com/user-attachments/assets/592ac13a-c9b5-4805-b5a5-2a48dada2498" />  |

However, we want to make sure that these manually-added widths are visible in every view:

- In the HTML view ✅ 
- In the "Preview" view ❓ 
- In the "edit_nofo" view ❓ 

We don't want writers to apply these widths and then be guessing if they are correctly applied.

pcraig3:
Following up on this, PR #298 finally resolves this issue:

- In the HTML view ✅
- In the "Preview" view ✅
- In the "edit_nofo" view ✅

All good! I'm closing this issue.

PR #298, opened 2025-04-21
Show table widths explicitly on "nofo_edit" page

pcraig3:
## Summary

This is a relatively small PR but hopefully one that has a big impact. 

This PR is aimed at helping NOFO designers when they apply table widths. The problem to be solved here is that writers/designers adding table widths didn't know if they were being applied correctly.

Given that, there are 3 changes in this PR regarding table widths:

1. Actually implement the widths in the nofo_edit page
    - Previously, applied widths would have no effect on the nofo_edit layout
2. Show the classnames on the nofo_edit page, in the th elements
3. Show the classnames in the martor preview, in the th elements

#### nofo_edit page

| before | after |
|--------|-------|
|  equal-width columns. no indication that widths had been explicitly applied.      |  columns are 25%, 25%, 50%. width classnames have been added to the `th` element.     |
|    Screenshot 2025-04-21 at 11 12 35 AM    |    Screenshot 2025-04-21 at 11 11 28 AM   |

#### martor preview

| before | after |
|--------|-------|
|   widths were applied correctly, but no explicit indicator.     |   added width classnames to `th` elements.    |
|  Screenshot 2025-04-21 at 11 11 03 AM      |     Screenshot 2025-04-21 at 11 10 23 AM  |

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,14 @@ Versioning since version 1.0.0.
 
 ### Added
 
+- Compare content guides
+- Delete a content guide
+- Use width classes for HTML tables on nofo_edit page
+
 ### Changed
+
+- Editing a content guide name from the nofo_edit page works better
+- Show HTML table classnames on nofo_edit page and martor preview
 
 ### Fixed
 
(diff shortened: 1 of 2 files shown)


pcraig3:
Merging this resolves #224.

Issue #272, opened 2025-03-27
Change application deadline everywhere all at once


pcraig3:
## Summary

NOFOs have a high-level metadata attribute [`application_deadline`](https://github.com/HHS/simpler-grants-pdf-builder/blob/main/bloom_nofos/nofos/models.py#L156) which is a string (not a date) containing the application deadline. 

The rest of the content of the NOFO is spread out across the `subsection.body` attributes. The application deadline is in a handful of other places and we also know that it's one of the most frequently updated data elements. 

Here are some of the other places that it is typically mentioned:

- key dates box
- application deadline section
- sometimes another rogue section - like cdc has it in their responsiveness criteria

We should add a function specifically to the application deadline edit view that uses regex to match application deadlines across the NOFO and then updates it to be the same as the latest change. 

This should _not_ work in reverse. If someone updates the body of the NOFO to change app deadline, on their head be it. 

We may also want to clean up & standardize app deadline during import if that would be feasible, that way this later function could just rely on simple string matching.

pcraig3:
I am closing this issue as completed. 

I was holding off until we talked to the NOFO editors, but they said they liked it and are actively using it now, so I think we can call this done.

PR #303, opened 2025-04-30
Change all matched instances of application_deadline value

pcraig3:
## Summary

This PR modifies the "Change Application Deadline" view by showing all other subsections where the current application deadline exists and offering to change them all at the same time.

## Details 

This story came out of our NOFO Builder retro with the coaches and HRSA.

We use the `application_deadline` field for the date that we have on the cover of the NOFO as well as the "Before you begin" page. However, it is also commonly in the body of the NOFO itself, usually inside of the "Key Dates" callout box.

If there is a change to the application date, writers have to hunt down all instances of this value and make sure they are all in sync.

The update here is to list them all on the edit page for the top-level `application_deadline` and offer to change them all at once.

**Note:** We don't have an understanding of which subsections the application deadline exists in, we literally just string match. If the date exists in another subsection we show it. If it doesn't exist, or if it does but it is in a different format ("June 10th, 2025" vs "June 10 2025") then we won't match it. This is a 'best-effort' type of feature. 

### Screenshots

| when you first load the page | after selecting subsections | success message | 
|--------|-------|------|
|   show the subsections with matching dates underneath the save button    |   selected subsections are highlighted and the button text is updated when you select a subsection    | success message tells you what changed |
|    Screenshot 2025-04-29 at 3 36 02 PM    |    Screenshot 2025-04-29 at 3 36 12 PM   | Screenshot 2025-04-29 at 3 36 47 PM |

Here's a gif of it working:

Screen Recording 2025-04-29 at 3 40 42 PM

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ Versioning since version 1.0.0.
 - Delete a content guide
 - Use width classes for HTML tables on nofo_edit page
 - Add cover image for HHS-2025-IHS-ALZ-0002
+- Early concept of "change this everywhere" feature for application deadline
 
 ### Changed
 
--- a/bloom_nofos/bloom_nofos/static/styles.css
+++ b/bloom_nofos/bloom_nofos/static/styles.css
@@ -1000,6 +1000,26 @@ main .back-link a::before,
   width: 16px;
 }
 
+/* Nofo edit properties pages */
+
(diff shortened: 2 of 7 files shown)


pcraig3:
Shipping this is the first step in solving https://github.com/HHS/simpler-grants-pdf-builder/issues/272 but it is not the last step.

PR #313, opened 2025-05-08
Migrate existing integer ids to UUIDs, bump to 3.0.0

pcraig3:
## Summary

This is a very large PR that migrates us away from incrementing integers and towards UUIDs.

There are no functional changes included in this PR, other than the URLs breaking.

### Description

Using UUIDs is generally a good idea because they are unguessable and also unique between systems. It would have been nice to start off with UUIDs but since we are moving to AWS very shortly, this is really our last best chance to get them in while we still have database access. 

We are also doing this to be more aligned with the CommonGrants schema, which prefers using a UUID as the primary key: https://commongrants.org/protocol/models/opp-base/

This is a breaking pull request because it breaks all the urls for our Nofos (and also all the urls for our content guides, but we don't care about these as much because nobody is using them).

The list of objects that we replaced the ids for are:

- `Nofo`
- `Section`
- `Subsection`
- `ContentGuide`
- `ContentGuideSection`
- `ContentGuideSubsection`

Since we were changing the `Nofo` and `ContentGuide` objects, I figured we might as well go the whole 9 yards and do the same for all the nested objects as well, as it would be weird for them to keep using integer ids.

The other thing we needed to do was keep the audit events updated. Here is what the CRUD events table looks like:

```
                                   Table "public.easyaudit_crudevent"
      Column       |           Type           | Collation | Nullable |             Default              
-------------------+--------------------------+-----------+----------+----------------------------------
 id                | integer                  |           | not null | generated by default as identity
 event_type        | smallint                 |           | not null | 
 object_id         | character varying(255)   |           | not null | 
 object_repr       | text                     |           | not null | 
 object_json_repr  | text                     |           | not null | 
 datetime          | timestamp with time zone |           | not null | 
 content_type_id   | integer                  |           | not null | 
 user_id           | bigint                   |           |          | 
 user_pk_as_string | character varying(255)   |           | not null | 
 changed_fields    | text                     |           | not null | 
Indexes:
    "easyaudit_crudevent_pkey" PRIMARY KEY, btree (id)
    "easyaudit_c_object__82020b_idx" btree (object_id, content_type_id)
    "easyaudit_crudevent_content_type_id_618ed0c6" btree (content_type_id)
    "easyaudit_crudevent_user_id_09177b54" btree (user_id)

```

And a random entry in the table:

```
 3843 |          2 | 1         | (Guide 1) CDC FY26 Content Guide | [{"model": "guides.contentguide", "pk": 1, "fields": {"uuid": "fc12b919-631e-47fc-a78c-a71442d3ce84", "filename": "CDC_Content_Guide.docx", "group": "bloom", "opdiv": "Centers for Disease Control and Prevention (CDC)", "status": "draft", "archived": null, "successor": null, "created": "2025-05-08T01:13:13.864Z", "updated": "2025-05-08T01:13:18.821Z", "title": "CDC FY26 Content Guide"}}] | 2025-05-07 21:13:18.824436-04 |              14 |       1 | 1                 | {"updated": ["2025-05-08 01:13:14.147589", "2025-05-08 01:13:18.821259"], "title": ["Enter the opportunity name.", "CDC FY26 Content Guide"]}
 ```
 
 So basically, we had to go in and update the `object_id` and the `object_json_repr > pk` fields for any nofos, sections, etc.  As you can imagine, this took forever, but it's done now!

### Did not migrate

Users are still using sequential ids. I don't know if we care about that so much because our user accounts are fundamentally disposable, so they will be easier to manage if we have to change them later, although we will need to decide how we handle this audit events stuff.


pcraig3:
🙏  📿  🧔‍♂️ 

let's blow this popsicle stand

PR #323, opened 2025-05-11
Change NOFO Number & Title Everywhere All At Once

cory-ignite:
## Summary
This PR modifies both the "Change NOFO Number" and "Change NOFO Title" views by showing all other subsections where these values exist and offering to change them all at the same time.

## Details
This story came out of our NOFO Builder retro with the coaches and HRSA.

We use the NOFO number and title fields for information that appears on the cover of the NOFO. However, these values are also commonly referenced throughout the body of the NOFO itself, such as in instructions or reference sections.

If there is a change to either the NOFO number or title, writers have to hunt down all instances of these values and make sure they are all in sync.

The update here is to list all instances on the respective edit pages for both the NOFO number and title, offering to change them all at once.

Note: We don't have an understanding of which subsections these values exist in, we literally just string match. If a value exists in another subsection we show it. If it doesn't exist, or if it is formatted differently ("HRSA-24-019" vs "HRSA 24 019") then we won't match it. This is a 'best-effort' type of feature.

## Screenshots for Number

| **when you first load the page** | **after selecting subsections** | **success message** |
|--------------|------------------|--------------|
| show the subsections with matching numbers underneath the save button | selected subsections are highlighted and the button text is updated when you select a subsection | success message tells you what changed |
| Screenshot 2025-05-11 at 8 39 52 AM | Screenshot 2025-05-11 at 8 39 59 AM | Screenshot 2025-05-11 at 8 40 11 AM |

## Screenshots for Title

| **when you first load the page** | **after selecting subsections** | **success message** |
|--------------|------------------|--------------|
| show the subsections with matching titles underneath the save button | selected subsections are highlighted and the button text is updated when you select a subsection | success message tells you what changed |
| Screenshot 2025-05-11 at 8 39 04 AM | Screenshot 2025-05-11 at 8 39 14 AM | Screenshot 2025-05-11 at 8 39 38 AM |

### DIFF
--- a/bloom_nofos/nofos/test_nofo.py
+++ b/bloom_nofos/nofos/test_nofo.py
@@ -6800,7 +6800,7 @@ class FindSubsectionsWithFieldValueTests(TestCase):
 
     def test_returns_title_match_with_basic_information_if_order_is_not_1(self):
         self.matching_subsection.name = "Basic information"
-        self.matching_subsection.order = 2
+        self.matching_subsection.order = 3
         self.matching_subsection.body = "This NOFO is titled Test NOFO"
         self.matching_subsection.save()
         results = find_subsections_with_nofo_field_value(self.nofo, "title")
@@ -6844,6 +6844,7 @@ class FindSubsectionsWithFieldValueTests(TestCase):
         self.assertEqual(self.matching_subsection.id, match["subsection"].id)
 
     def test_returns_multiple_matches_for_same_field(self):
+        # NOTE - Create a second subsection that also contains the title
         second_subsection = Subsection.objects.create(
             section=self.section,
             name="Another Section",
@@ -6862,8 +6863,8 @@ class FindSubsectionsWithFieldValueTests(TestCase):
         self.assertIn(self.matching_subsection.id, subsection_ids)
         self.assertIn(second_subsection.id, subsection_ids)
 
-    def test_returns_empty_when_field_value_is_none(self):
-        self.nofo.number = None
+    def test_returns_empty_when_field_value_is_empty(self):
+        self.nofo.number = ""
         self.nofo.save()
         results = find_subsections_with_nofo_field_value(self.nofo, "number")
         self.assertEqual(results, [])
@@ -7024,7 +7025,6 @@ class ReplaceValueInSubsectionsTests(TestCase):
             "HRSA-24-019",
             "HRSA-24-020",
         )
-        # NOTE - This should not match due to different formatting
         self.assertEqual(len(updated), 0)
         self.subsection_1.refresh_from_db()
         self.assertIn("HRSA - 24 - 019", self.subsection_1.body)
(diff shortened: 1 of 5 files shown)


cory-ignite:
For cross reference purposes - This is similar to #272 and #303.

bloom_nofos/nofos/test_nofo.py
```diff
@@ -6683,7 +6683,7 @@ def test_returns_match_with_highlight(self):
         self.assertIn("Deadline Details", match["subsection"].name)
         self.assertEqual(self.section, match["section"])
         self.assertEqual(self.matching_subsection.id, match["subsection"].id)
-        self.assertEqual(self.matching_subsection.body, match["subsection"].body)
+        self.assertEqual(match["subsection"].body, self.matching_subsection.body)
```
pcraig3, line 6687: Looks like this line and the next one are causing a merge conflict (if you do `git rebase main`), but easy enough to fix.

bloom_nofos/nofos/test_nofo.py
```diff
@@ -6815,3 +6964,132 @@ def test_mixed_match_and_nonmatch(self):
...
+    #     # we should test the actual behavior rather than what we initially expected
+    #     original_body = self.subsection_1.body
+    #     updated = replace_value_in_subsections(
+    #         [self.subsection_1.id],
+    #         "",
+    #         "New Value",
+    #     )
+    #     self.assertEqual(len(updated), 1)
+    #     self.subsection_1.refresh_from_db()
+    #     self.assertNotEqual(self.subsection_1.body, original_body)
```
pcraig3, line 7095: Interesting, are you saying this will run the `replace` with an empty string (effectively, erasing the original value)?

pcraig3 (changes requested):
Overall it looks good and is working! Hooray. 

A few bits of feedback from me:

### 1. We should default to checked for all of these pages. 

We showed it to our writers and they said more or less "it's good that we can uncheck if we need to, but I don't imagine I would ever uncheck this". 

99% of the time, they will just accept all the changes, so we should just load the page with all of them checked.

### 2. There is a lot of copy+paste going on here

I think there are a few obvious places to reduce some if it:

- maybe make the classname similar so that we can reduce some of the CSS selectors
- the javascript in the `<script>` tags seems like it's all basically the same, so we could easy make this a script and then include it in `{% footer_scripts %}`
- I am wondering what we could harmonize from the views, since those `form_valid` functions are looking pretty bulky

The only one I would not generalize yet would be the template themselves, as we have a future story coming up which is also about building tables like this and I think we might potentially want to update this HTML.

I am also okay to merge this PR as is and then do a second refactor PR. 

3. Replace with an empty value if an empty string is passed in

I think it's fair to assume we would never do this. Currently, these pages aren't submittable with empty values, so it can't be called this way, but we should close this path down and return early if the value to update to is an empty string.

pcraig3:
Oh, and a final thing to mention is that I have been generally putting new updates in the Changelog file. 

I don't update the version on every change, since versions are not really important to us. I have generally been versioning the app whenever the database is being changed in some way. Other changes end up in [the "Unversioned" section at the top of the Changelog](https://github.com/HHS/simpler-grants-pdf-builder/blob/main/CHANGELOG.md#unversioned). 

This is not a strictly correct implementation of a Changelog, but it does have eventual consistency (each numbered version contains the changes that are listed) and in practice it doesn't really matter because nobody is relying on our version numbers.

Issue #322, opened 2025-05-10
Add a fixed flash message on the nofo_edit page for NOFO edits


pcraig3:
We used to have a simple layout (from an HTML perspective) that we eventually learned people hated. 

What we would do for Nofo edits was:  

- someone makes an edit to a Subsection
- Redirect back to the top of the `nofo_edit` page
- Show a flash message with a link to the edited subsection
- If you click the link, you zip down to it

What we heard from writers was that they often are going subsection by subsection when they are making edits, so they want to go from one subsection to the next one quickly. However, in this old interaction flow, they are constantly being yanked back to the top of the document. No good!

Based on that feedback, [I changed the redirect to slap an anchor link to the end of the `nofo_edit` url](https://github.com/HHS/simpler-grants-pdf-builder/pull/275), so now you are scrolled back down to the same subsection you edited. 

But now we have a slight usability problem which is that our sad little flash message is still at the top of the page but there is no way to see it. So you essentially get no feedback as an editor now.

Adam has suggested we change the existing flash message to be a “fixed-to-the-top-of-the-browser” one, which makes sense in principle. It's a little complicated though because we already have a floating header for this page, so we would want to maybe add an offset to accommodate it or something. 

Also, if we add a fixed alert message, we have to make sure (a) you can dismiss it, and (b) if you don't dismiss it, it eventually goes away. Otherwise it will just live on your screen until you refresh. 

The new flow would look like:

- someone makes an edit to a Subsection
- Redirect back to the part of the `nofo_edit` page that they just edited
- Show a success flash message that is fixed to the viewport. 
    - Doesn’t need to have a link since we are already linking you there.
    - Message goes away after 3-5 (?) seconds, or can be dismissed by clicking “x”. 

I am assuming that this completely replaces our existing flash message for this page, so all messages on this page would then use this format (deleting/adding a subsection, for example) now use this same format. We don't need to keep the old flash message around.

Notes:

- USDS has “slim alerts”, we could probably use that: https://designsystem.digital.gov/components/alert/
- Accessibility-wise, it probably makes sense to keep the markup where the current flash messages are (before the h1), so that you would see it first and can dismiss it right away if you are doing keyboard navigation.

pcraig3:
This ticket is closed since #324 went in.

pcraig3:
Adam messaged me about this yesterday:

[...]

PR #324, opened 2025-05-12
Fixed flash message after editing

cory-ignite:
## Summary
This PR adds sticky alert functionality to the NOFO Builder, with auto-dismiss behavior and proper accessibility features. The alerts will stay at the top of the page while scrolling, auto-dismiss after 4 seconds, and can be manually dismissed. The implementation ensures proper keyboard navigation and screen reader support.

## Details
This update came from the need to keep alerts visible while users scroll through long NOFO documents, particularly when making edits or reviewing changes. The alerts now:

- Stay fixed at the top of the page while scrolling / auto-scrolling
- Auto-dismiss after 4 seconds with a fade animation
- Can be manually dismissed via a close button
- Support keyboard navigation and screen reader announcements
- Properly layer with other page elements through z-index management

### Accessibility Features
- Success messages use `aria-live="polite"` to announce changes when the user is idle
- Error messages use `aria-live="assertive"` to announce changes immediately
- Close button is keyboard accessible with proper focus states
- Proper ARIA labels for screen readers
- Focus management for keyboard users

## Screenshots
Screenshot 2025-05-11 at 11 09 19 PM

This addresses issue #322.

### DIFF
--- a/bloom_nofos/bloom_nofos/static/styles.css
+++ b/bloom_nofos/bloom_nofos/static/styles.css
@@ -520,6 +520,36 @@ label.usa-label {
   z-index: 1;
 }
 
+.usa-alert__close {
+  position: absolute;
+  right: 1rem;
+  top: 50%;
+  transform: translateY(-50%);
+  background: none;
+  border: none;
+  font-size: 1.5rem;
+  cursor: pointer;
+  padding: 0;
+  width: 32px;
+  height: 32px;
+  line-height: 32px;
+  text-align: center;
+  opacity: 0.7;
+  color: inherit;
+  border-radius: 4px;
+}
+
+.usa-alert__close:hover,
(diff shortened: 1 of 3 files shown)


pcraig3 (changes requested):
Amazing, I think this looks good!

It is a shame that the alert overlaps the other sticky header, but let's just get it out and if Adam complains maybe we add an offset or something. But other than that, it looks good. 👍 

The only think I noticed is that there is a console error in the browser after the alert disappears. 

Screenshot 2025-05-12 at 3 11 28 PM

It seems to happen on lots of alerts (editing a subsection, removing a subsection, adding a subsection), so it is probably just looking for some element that is gone.

Once that's done, let's update the CHANGELOG

pcraig3 (approved):
Looks great!

PR #326, opened 2025-05-12
Change top-level directory from `bloom_nofos` to `nofos`

pcraig3:
## Summary

This is both a small change but also a major one.

We are changing the top-level directory for the app to be `./nofos`, rather than `./bloom_nofos`. 

This is not a huge change functionally, but it does mean we will have really painful diffs for any branches that are opened before this change goes in. Best to merge it ASAP, once we know that it is going to work.

The motivation here is so that we are consistently naming folders in our repo as well as in the simpler-grants-gov repo, where our code is ultimately going to be deployed from.


pcraig3:
I'm pulling the trigger on this.

PR #328, opened 2025-05-13
Add the "Important: public information" section after LAST matching subsection

pcraig3:
## Summary

This is a small PR that changes how we insert the `Important: public information` subsection. Instead of inserting it after the _first_ matching heading in Section 3, now we insert it into the _last_ matching heading. Essentially, we are just looping through subsections in reverse to find a match.

### Details

Since February, we have been injecting a new subsection (`Important: public information`) into step 3 of all Nofos: https://github.com/HHS/simpler-grants-pdf-builder/pull/96

It's generally been working as expected, until last week when we had a few ACF NOFOs come through where the "Public info" subsection was inserted right at the beginning of the section (it was 3rd in the list of subsections) and this confused everyone. This subsection is intended to come towards the end of Section 3, not at the beginning.

The reason for this is that `add_final_subsection_to_step_3` looks to match on several different of subsection headings and then adds the "Public info" subsection after the _first_ match. Based on the amount of confusion it created, this is not actually what we expect this function to do. We expect the "Public info" subsection to be closer to the end of section 3.

The cleanest way to solve this seemed like we can just looking for matching subsection headings _in reverse order_ and then add the "Public info" subsection after the matching subsection that is _closest to the end_ of the section.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,7 @@ Versioning since version 1.0.0.
 - IHS not-selected top nav link borders on section pages are white
 - Tag docker images built with Makefile with latest git sha (or "latest")
 - Show sticky flash message for edits on nofo_edit page
+- Add the "Important: public information" subsection after _last_ matching subsection
 
 ### Fixed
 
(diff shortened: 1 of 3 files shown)


pcraig3:
Merging this – very small change.

PR #332, opened 2025-05-14
Refactor for Edit View, Update Empty String Logic, & other Small Bugfixes

cory-ignite:
## Summary
The fixes in the PR here primary address concerns from [this comment](https://github.com/HHS/simpler-grants-pdf-builder/pull/323).

## Details
This is a bit of a refactor in order to de-duplicate logic / templating being shared across multiple edit screens.

In no particular order:
- Pulled out shared Javascript into the `static` folder as `nofo_edit_form.js`
- Pulled out shared templating logic into `nofo_edit_base.html`
- Updated logic in `forms.py` to check for empty strings
- Updated logic in `models.py` to check for empty strings
- Added/Fixed some tests

## Screenshots

| Empty String for Title | Empty String for App Deadline | Empty String for NOFO Number |
| -------- | -------- | -------- |
| Screenshot 2025-05-14 at 10 48 36 AM | Screenshot 2025-05-14 at 10 48 55 AM | Screenshot 2025-05-14 at 10 48 22 AM |

### DIFF
--- a/nofos/nofos/forms.py
+++ b/nofos/nofos/forms.py
@@ -33,15 +33,47 @@ def create_object_model_form(model_class):
 create_nofo_form_class = create_object_model_form(Nofo)
 
 
-# Creating form classes
-NofoNameForm = create_nofo_form_class(
-    ["title", "short_name"], not_required_labels=["Short name"]
-)
+# Creating form classes with validation
+class NofoNameForm(forms.ModelForm):
+    class Meta:
+        model = Nofo
+        fields = ["title", "short_name"]
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.fields["short_name"].required = False
+
+    def clean_title(self):
+        title = self.cleaned_data.get("title")
+        if title is None or not title.strip():
+            raise forms.ValidationError("Title cannot be empty.")
+        return title.strip()
+
+class NofoNumberForm(forms.ModelForm):
+    class Meta:
+        model = Nofo
+        fields = ["number"]
+
+    def clean_number(self):
+        number = self.cleaned_data.get("number")
+        if number is None or not number.strip():
+            raise forms.ValidationError("Number cannot be empty.")
+        return number.strip()
+
+class NofoApplicationDeadlineForm(forms.ModelForm):
+    class Meta:
+        model = Nofo
+        fields = ["application_deadline"]
+
+    def clean_application_deadline(self):
+        deadline = self.cleaned_data.get("application_deadline")
+        if deadline is None or not deadline.strip():
+            raise forms.ValidationError("Application deadline cannot be empty.")
+        return deadline.strip()
+
 NofoAgencyForm = create_nofo_form_class(["agency"])
-NofoApplicationDeadlineForm = create_nofo_form_class(["application_deadline"])
 NofoCoverForm = create_nofo_form_class(["cover"])
 NofoGroupForm = create_nofo_form_class(["group"])
-NofoNumberForm = create_nofo_form_class(["number"])
 NofoOpDivForm = create_nofo_form_class(["opdiv"])
 NofoSubagencyForm = create_nofo_form_class(
     ["subagency"], not_required_labels=["Subagency"]
--- a/nofos/nofos/models.py
+++ b/nofos/nofos/models.py
@@ -378,10 +378,14 @@ class Nofo(BaseNofo):
         super().clean()
         errors = {}
 
-        # Validate title or number is present
-        if not self.title and not self.number:
-            errors["title"] = "Either title or number must be provided"
-            errors["number"] = "Either title or number must be provided"
+        # Validate title or number is present and not just whitespace
+        title_empty = not self.title or not self.title.strip()
+        number_empty = not self.number or not self.number.strip()
+        
+        if title_empty and number_empty:
+            errors["title"] = "Either title or number must be provided and cannot be empty"
+            errors["number"] = "Either title or number must be provided and cannot be empty"
+
 
         if errors:
             raise ValidationError(errors)
--- a/nofos/nofos/nofo.py
+++ b/nofos/nofos/nofo.py
@@ -2427,7 +2427,7 @@ def replace_value_in_subsections(subsection_ids, old_value, new_value):
     
     If new_value is empty, no replacements will be made and an empty list will be returned.
     """
-    if not subsection_ids or not new_value:
+    if not subsection_ids or not new_value or not old_value:
         return []
 
     updated_subsections = []
--- a/nofos/nofos/nofo.py
+++ b/nofos/nofos/nofo.py
@@ -2424,8 +2424,10 @@ def replace_value_in_subsections(subsection_ids, old_value, new_value):
     Replaces `old_value` with `new_value` in bodies of given subsection_ids.
     Note that old_value uses a case-insensitive compare, so "JUNE 1" will match "June 1".
     Returns the list of updated Subsection objects.
+    
+    If new_value is empty, no replacements will be made and an empty list will be returned.
     """
-    if not subsection_ids:
+    if not subsection_ids or not new_value:
         return []
 
     updated_subsections = []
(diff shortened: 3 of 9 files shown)


nofos/nofos/models.py
```diff
@@ -378,10 +378,14 @@ def clean(self):
...
-            errors["title"] = "Either title or number must be provided"
-            errors["number"] = "Either title or number must be provided"
+        # Validate title or number is present and not just whitespace
+        title_empty = not self.title or not self.title.strip()
+        number_empty = not self.number or not self.number.strip()
+        
+        if title_empty and number_empty:
+            errors["title"] = "Either title or number must be provided and cannot be empty"
+            errors["number"] = "Either title or number must be provided and cannot be empty"
+
```
pcraig3, line 388: I am not sure it is possible to end up with an empty string here? (Maybe though a handwritten post request?)

Anyway, seems like it is safer to do this check so let's keep it.

nofos/nofos/forms.py
```diff
@@ -33,15 +33,47 @@ def __init__(self, *args, **kwargs):
...
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.fields["short_name"].required = False
+
+    def clean_title(self):
+        title = self.cleaned_data.get("title")
+        if title is None or not title.strip():
+            raise forms.ValidationError("Title cannot be empty.")
+        return title.strip()
```
pcraig3, line 50: Do we need these extra forms? I think this logic is already applied in our `create_nofo_form_class`. 

There is a better validation message from these forms though.

Were you able to find a scenario in the previous forms that were letting bad values through?

nofos/nofos/nofo.py
```diff
@@ -2424,8 +2424,10 @@ def replace_value_in_subsections(subsection_ids, old_value, new_value):
     Replaces `old_value` with `new_value` in bodies of given subsection_ids.
     Note that old_value uses a case-insensitive compare, so "JUNE 1" will match "June 1".
     Returns the list of updated Subsection objects.
+    
+    If new_value is empty, no replacements will be made and an empty list will be returned.
     """
-    if not subsection_ids:
+    if not subsection_ids or not new_value or not old_value:
```
pcraig3, line 2430: 👍

nofos/nofos/templates/nofos/nofo_edit_application_deadline.html
```diff
@@ -1,122 +1,31 @@
...
-    {% include "includes/form_macro.html" with hint2="eg: “Monday, January 1, 2024”" %}
+{% block title %}NOFO Application Deadline{% endblock %}
 
-    <button class="usa-button margin-top-3" type="submit">Save application deadline</button>
+{% with page_title="NOFO Application Deadline" %}
+{% with form_id="nofo_edit_application_deadline_form" %}
+{% with button_text="Save application deadline" %}
+{% with match_value=form.application_deadline.value %}
+{% with replace_text="application deadline" %}
+{% with checkbox_checked=True %}
```
pcraig3, line 12: I think in these Django templates, it's possible to do:

```
{% with 
    page_title="NOFO Application Deadline" 
    form_id="nofo_edit_application_deadline_form" 
    button_text="Save application deadline" 
    match_value=form.application_deadline.value 
    replace_text="application deadline" 
    checkbox_checked=True 
%}

some code

{% endwith %}
```

nofos/nofos/forms.py
pcraig3: My concern is that we have a factory function which is creating almost all of our attribute-level forms, and I don't really want to fork that into other base classes if we don't need to (particularly where seems like we aren't relying on extra functionality).

nofos/nofos/templates/nofos/nofo_edit_base.html
```diff
@@ -0,0 +1,79 @@
+{% extends 'base.html' %}
```
pcraig3, line 1: I think I had a slightly different vision for what this could be.

I was thinking we could keep the edit templates more or less as they were, but then create a macro here for the table-building part of the HTML, so everything between the `{% if subsection_matches %}` block. We could also stick the JS code in a script that is added at the end of the table so that it is only included where that table needs to be built. 

Similar issue with the Form classes above: the thing that is different about these pages (compared to our other edit_{attribute_name} templates) is that we are building a table of subsections, so I would much prefer if just the table was sheared off into its own macro.

What do you think, am I missing something?

nofos/nofos/test_nofo.py
```diff
@@ -1004,6 +1004,31 @@ def test_create_nofo_success_no_title(self):
         self.assertEqual(nofo.title, "")
         self.assertEqual(len(nofo.sections.all()), 0)
 
+    def test_replace_value_in_subsections_empty_value(self):
```
pcraig3, line 1007: 👍

nofos/nofos/test_nofo.py
```diff
@@ -7044,29 +7069,25 @@ def test_replace_multiple_occurrences_in_single_subsection(self):
...
-    #     original_body = self.subsection_1.body
-    #     updated = replace_value_in_subsections(
-    #         [self.subsection_1.id],
-    #         "",
-    #         "New Value",
-    #     )
-    #     self.assertEqual(len(updated), 1)
-    #     self.subsection_1.refresh_from_db()
-    #     self.assertNotEqual(self.subsection_1.body, original_body)
+    def test_replace_empty_old_value_behavior(self):
```
pcraig3, line 7084: Is this testing something different than the test above? It seems like the same test or very similar.

pcraig3 (changes requested):
Hello! I think the scripts + model improvements look generally good, but I am not convinced by the Form updates or the template updates. 

Let me know what you think of my comments — we can also meet to chat about this if you like.

nofos/nofos/models.py
cory-ignite: I was thinking thru a handwritten post request - which would be **super** edige case based on the user demographic. But, still good to keep around. I went down a bit of a rabbit hole after adding in [that one test](https://github.com/HHS/simpler-grants-pdf-builder/pull/332/files#diff-b6aa406f9c09bca25487a326428c79bbe0f1635a043cd1b210adcb2f49148b91R7084) during the previous PR #323.

nofos/nofos/templates/nofos/nofo_edit_application_deadline.html
cory-ignite: Good to know. First time with Django. 👍

Will fix in next Commit.

nofos/nofos/test_nofo.py
cory-ignite: Slightly different

This test (`test_replace_empty_old_value_behavior`) tests what happens when you try to replace an **empty** string with a new value.
The test above (`test_replace_value_in_subsections_empty_value`) tests what happens when you try to replace an existing value with an **empty** string.

nofos/nofos/test_nofo.py
cory-ignite: Will move down near other empty string test at the bottom of file.

nofos/nofos/templates/nofos/nofo_edit_application_deadline.html
cory-ignite: I tried this for a little, but wound up not getting the HTML to render. Don't want to keep spinning my wheels on it...

```
Error during template rendering
In template /Users/corytrimm/Documents/GitHub/simpler-grants-pdf-builder/nofos/nofos/templates/nofos/nofo_edit_title.html, error at line 34

Invalid block tag on line 34: 'endwith', expected 'endblock'. Did you forget to register or load this tag?
```

pcraig3 (approved):
This looks great! Let's ship it 👍

nofos/nofos/templates/nofos/nofo_edit_application_deadline.html
pcraig3: Hmmm, you know what? It could be that it doesn't work, or it could also be that the newlines are illegal characters and everything needs to be on one line. 

In any case, all good for this PR. Apologies!

PR #333, opened 2025-05-14
Added simple /health API endpoint

cory-ignite:
## Summary
Add a health check endpoint at `/api/health` that doesn't require auth.

## More Info
We've decided to add this such that when the app is running, we can quickly check the status. This is in anticipation of running this within AWS.

Please note that this is _very_ simple and could be expanded to include checking the Database Connection too.

## Screenshots
Screenshot 2025-05-14 at 1 19 31 PM


pcraig3:
Gonna merge this since we are happy with it.

PR #334, opened 2025-05-15
Add wget for docker & support `HEAD` method for /health

pcraig3:
## Summary

This PR adds `wget` to our Dockerfile and also allows the `HEAD` method for the `/health` endpoint.

### Description

The "healthcheck_command" in simpler-grants-gov is running `wget` inside of the container to check that the app is up and running.

This means we need to install wget as a Docker dependency.

Here's the command ([link](https://github.com/HHS/simpler-grants-gov/blob/ea2460fd5b83763f633b6f21ef51a09e873c515b/infra/modules/service/variables.tf#L178)):
- `wget --no-verbose --tries=1 --spider http://localhost:8000/health || exit 1`

Also, it's not immediately obvious, but the "--spider" option will send a `HEAD` request instead of a `GET` request:

So this PR also enables the "HEAD" method and also adds a test to confirm.

### How to test this PR

The way to test this PR is to run the `CMD-SHELL` inside of a docker container, which is [how the `healthcheck_command` works in simpler-grants-gov](https://github.com/HHS/simpler-grants-gov/blob/ea2460fd5b83763f633b6f21ef51a09e873c515b/infra/modules/service/variables.tf#L173).

- From the repo root, build a docker container: `make build` 
- Run container in detached mode: `container_id=$(docker run -d -p 8000:8000 nofos:latest)`
- Run health check: `docker exec $container_id sh -c "wget --no-verbose --tries=1 --spider http://localhost:8000/health || exit 1"`
- Stop container: `docker stop $container_id`
- Delete container: `docker rm $container_id`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@ Versioning since version 1.0.0.
 
 ### Added
 
+- Add healthcheck endpoint at /health
+
 ### Changed
 
 ### Fixed
--- a/Dockerfile
+++ b/Dockerfile
@@ -23,6 +23,7 @@ RUN apt-get update && \
   apt-get install -y --no-install-recommends \
   build-essential \
   curl \
+  wget \
   libffi-dev \
   libpq-dev \
   && rm -rf /var/lib/apt/lists/*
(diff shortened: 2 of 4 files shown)


cory-ignite:
Looks good to me.

[Here's a bit more info I've found](https://superuser.com/a/797372) on the `HEAD` request with the `--spider` option on `wget`.

coilysiren:
🚀

Issue #331, opened 2025-05-13
[Potential Bug] Hyperlink Still Exists in Notification Header When Adding a New Section


cory-ignite:
Would like your thoughts on this one, @pcraig3. Ran into this while testing #330.

I was trying to make heads / tails of what you said [over here](https://github.com/HHS/simpler-grants-pdf-builder/pull/323#pullrequestreview-2834141949), but am still unclear.

See screenshot below. You notice that the subsection title / name is still hyperlinked. This is different than the behavior when you simply `edit` a subsection.

<img width="969" alt="Image" src="https://github.com/user-attachments/assets/3aa45379-b45c-4f12-a7a1-75797aa61b96" />

cory-ignite:
Ok - dug into this a bit more. Below are the various tested user flows...

**User Flows Tested:**
| Flow | Has Hyperlink | Auto-Scrolls |
|------|--------------|--------------|
| Reimport NOFO | No | No |
| Import NOFO | No | No |
| Delete NOFO | No | No |
| Add Subsection | Yes | Yes (to new subsection) |
| Add Page Break to existing subsection | No | Yes (to new page break) |
| Edit Subsection | No | Yes (to updated subsection) |
| Delete Subsection | No | No |
| NOFO Edit Title | No | No |
| NOFO Edit Application Number **w/out** Subsection Change | No | No |
| NOFO Edit Application Number **w/** Subsection Change | Yes - to changed Subsections | No |
| NOFO Edit Application Deadline **w/out** Subsection Change | No | No |
| NOFO Edit Application Deadline **w/** Subsection Change | Yes - to changed Subsections | No |

pcraig3:
Interesting to see it all laid out like this. 

I think the hyperlinks on the "Edit Nofo Number" and "Edit Application Deadline" (and, I presume, "Edit Nofo Title") are all good, because they are linking to subsections that are not visible on the screen.

The “new subsection” link we can probably do away with now because we are just be linking to where you already are.

Everything else is okay to stay — do you agree?

cory-ignite:
Closing this issue out with the merge of #342.

PR #335, opened 2025-05-15
Updated id selector CSS + added animations for alert/success

cory-ignite:
## Summary
[This PR addresses the comment to move the alert/success to the bottom of the NOFO Builder](https://github.com/HHS/simpler-grants-pdf-builder/issues/322#issuecomment-2880610918). It still has with auto-dismiss behavior and proper accessibility features. The alerts will stay at the **bottom** of the page while scrolling, auto-dismiss after 4 seconds, and can be manually dismissed. The implementation ensures proper keyboard navigation and screen reader support.

## GIF Demo
NOFO-fade

## Screenshot
Screenshot 2025-05-15 at 2 00 48 PM

### DIFF
--- a/nofos/bloom_nofos/static/styles.css
+++ b/nofos/bloom_nofos/static/styles.css
@@ -484,14 +484,28 @@ label.usa-label {
 }
 
 /* Alert styles */
-.usa-alert {
-  position: sticky;
-  top: 0;
-  z-index: 9999;
-  transition: opacity 0.3s ease-out;
-  margin-bottom: 1rem !important;
+#nofo-editor-success-alert,
+#nofo-editor-error-alert {
+  position: fixed;
+  bottom: 0;
+  left: 50%;
+  transform: translateX(-50%);
+  width: 100%;
+  z-index: 99999;
+  margin: 0 !important;
   background: white;
-  isolation: isolate;
+  opacity: 0;
+  transition: opacity 0.3s ease-out;
+}
+
+#nofo-editor-success-alert.fade-out,
+#nofo-editor-error-alert.fade-out {
+  opacity: 0;
+}
+
+#nofo-editor-success-alert .usa-alert__body,
+#nofo-editor-error-alert .usa-alert__body {
+  max-width: none;
 }
 
 .usa-alert__body {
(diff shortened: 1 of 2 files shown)


nofos/bloom_nofos/static/styles.css
```diff
@@ -484,14 +484,28 @@ label.usa-label {
...
 
 /* Alert styles */
-.usa-alert {
-  position: sticky;
-  top: 0;
-  z-index: 9999;
-  transition: opacity 0.3s ease-out;
-  margin-bottom: 1rem !important;
+#nofo-editor-success-alert,
+#nofo-editor-error-alert {
```
pcraig3, line 488: I was thinking we should only target the alerts on the `nofo_edit` template (so, make this css look like `.nofo_edit #nofo-editor-success-alert` etc) so that it didn't affect other alerts, but I actually don't think it looks too bad. 

For example, if you delete a nofo or add one on the `/nofos` page, you see the flash message on the bottom. Maybe this is good for consistency?

So I guess I am thinking we can just keep it this way until such time as it becomes an issue. What do you think?

pcraig3 (approved):
It looks good! I am going to pre-approve because it can go in as is but I also made 1 comment.

nofos/bloom_nofos/static/styles.css
cory-ignite: Maybe we run this by @admoorgit for some design / product thinking?

> For example, if you delete a nofo or add one on the /nofos page, you see the flash message on the bottom. Maybe this is good for consistency?

Here's a screenshot of what that looks like on `localhost`. To @pcraig3's point - it makes it more consistent. Currently on the `main` branch, the notification lives above the list of NOFOs and then fades away - causing the page to 'jump' - see the GIF down below.

**Importing a new NOFO** -
Screenshot 2025-05-16 at 2 01 22 PM

**Deleting a NOFO** -
Screenshot 2025-05-16 at 2 01 30 PM

**GIF of Page Jumping on `main` branch** - 
NOFO-notification-jump

ctrimm:
Based on conversations from late last week, this is approved and ready to be merged into `main`.

If we want to revisit how / when there are hyperlinks & automatic scrollTo occurring, [we can review the notes left on another bug here](https://github.com/HHS/simpler-grants-pdf-builder/issues/331#issuecomment-2890777288).

cory-ignite:
Based on conversations from late last week, this is approved and ready to be merged into `main`.

If we want to revisit how / when there are hyperlinks & automatic scrollTo occurring, [we can review the notes left on another bug here](https://github.com/HHS/simpler-grants-pdf-builder/issues/331#issuecomment-2890777288).

PR #339, opened 2025-05-20
Update ALLOWED_HOSTS for AWS

pcraig3:
## Summary

This PR makes 2 changes:

1. Adds error checking to our `gethostbyname(gethostname())` call. 
2. Adds the Load Balancer DNS name to `ALLOWED_HOSTS` if it exists

These 2 changes will allow the container to run in non-Debug mode in AWS without using an "allow-everything" ALLOWED_HOSTS value of "*". 

They will also allow the app to run on Cory's laptop.

## Details

1. Adds error checking to our `gethostbyname(gethostname())` call.

Some people see an error when calling `gethostbyname(gethostname())`:

- https://stackoverflow.com/questions/39970606/gaierror-errno-8-nodename-nor-servname-provided-or-not-known-with-macos-sie
- https://github.com/zauberzeug/nicegui/issues/1178

We don't actually need to solve this error, we just need it to not crash. The point of this code is so that we can get our internal IP when we are running in a docker container in AWS. Locally, this condition will be skipped.

2. Adds the Load Balancer DNS name to `ALLOWED_HOSTS` if it exists

The DNS name is the semi-randomized web url that the load balancer accepts traffic at. If you visit it, you will see the app, but then Django will be mad if it doesn't know about the domain. 

This env var will exist in our environment if we are running in AWS, since this PR went in:  https://github.com/HHS/simpler-grants-gov/pull/5101

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ Versioning since version 1.0.0.
 ### Fixed
 
 - Show audit events for deleted subsections on the "All updates for NOFO" page
+- Add AWS Load Balancer domain to ALLOWED_HOSTS
 - Add current hostname to ALLOWED_HOSTS if using a private IP range starting with "10."
 
 ## [3.1.0] - 2025-05-12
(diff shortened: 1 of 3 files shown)


cory-ignite:
Pulled branch locally & tested this a couple ways:
- [x] Ran `poetry run start` to fire up app server - navigated to a few pages, all good here.
- [x] Ran `poetry run test` to run all tests - all tests passed, no weird output.

Looks good to me. 👍

PR #345, opened 2025-05-21
Subclass Django's DatabaseWrapper backend for RDS

pcraig3:
## Summary

This PR updates our database wrapper to account for AWS' periodic token refreshing. The problem with the previous Database wrappers is that the password is set at the beginning and then never changed. AWS refreshes their password token every 15 minutes, so we need to be continually requesting new tokens when we reconnect.

## Description 

Generally, when you set your database password, the value you set for "password" is static for the lifetime of the app that is running. Django has some settings around this (eg, [`CONN_MAX_AGE`](https://docs.djangoproject.com/en/5.2/ref/settings/#conn-max-age)), but they assume they will get a new connection with the same static variables. 

What we need to do when we connect to RDS is refresh the password periodically so that we are getting _new_ tokens, rather then trying to refresh the connection with the old token.

So this custom backend does that. Essentially, we are overriding the `get_connection_params` method to call our custom function `generate_iam_auth_token()` to set the password.

Useful sources in case you want to learn more!

- https://github.com/labd/django-iam-dbauth/tree/main
- https://stackoverflow.com/questions/57865837/how-to-do-rds-iam-authentication-with-django

### How to test this

Unfortunately, we can't test this locally. 

What you can do is visit : http://nofos-dev-1402022759.us-east-1.elb.amazonaws.com/ 

If you can see the website more than 15 minutes after our last deploy, this is working!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ Versioning since version 1.0.0.
 - Show audit events for deleted subsections on the "All updates for NOFO" page
 - Add AWS Load Balancer domain to ALLOWED_HOSTS
 - Add current hostname to ALLOWED_HOSTS if using a private IP range starting with "10."
+- Refresh AWS DB connection password periodically
 
 ## [3.1.0] - 2025-05-12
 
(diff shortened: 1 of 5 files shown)


coilysiren (approved):
Fancy fancy

PR #352, opened 2025-05-22
Remove CONN_MAX_AGE and CONN_HEALTH_CHECKS in aws database config

pcraig3:
## Summary

Remove `CONN_MAX_AGE` and `CONN_HEALTH_CHECKS` settings in the database config.  I suspect this may be caching our password for too long.

### Details

I noticed that our deployed nofo app was failing after a while because of the database connection, so my supposition is that the CONN_MAX_AGE variable holds onto the wrong password for too long and at some point they go out of sync.

Not totally sure this is a solution but I notice that the django-iam-dbauth library isn't recommending CONN_MAX_AGE or doing anything around this in general:

- https://github.com/labd/django-iam-dbauth

Note that I added a print statement to the DatabaseWrapper to prove that that code actually runs when we make requests.

Screenshot 2025-05-22 at 1 11 56 PM

That worked, so we know that code is running, and now I have removed those print statements.

### DIFF
--- a/nofos/bloom_nofos/db/backends/postgresql_iam/base.py
+++ b/nofos/bloom_nofos/db/backends/postgresql_iam/base.py
@@ -11,9 +11,7 @@ class DatabaseWrapper(BaseDatabaseWrapper):
     def get_connection_params(self):
         params = super().get_connection_params()
 
-        print("--- GET PARAMS")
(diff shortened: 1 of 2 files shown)


pcraig3:
I'm closing this since it doesn't appear that we actually need it.

PR #353, opened 2025-05-22
Create Dockerfile non-root user ("appuser") for container runtime

pcraig3:
## Summary

This commit adds a dedicated 'appuser' to the Docker image and switches to it before the application is started. 

### Details

This resolves the CIS-DI-0001 warning flagged by Dockle, which requires containers to avoid running as root.

Example here: https://github.com/HHS/simpler-grants-gov/actions/runs/15195040903/job/42737014418

By running as a non-root user:

- The container is less likely to be used as a pivot point in an attack
- Accidental or malicious writes to system paths are prevented
- Compliance scanners (e.g., Dockle, Trivy, etc.) pass with fewer flags


pcraig3:
Merging for compliance reasons

Issue #278, opened 2025-03-28
Add sticky side nav menu for section titles


pcraig3:
## Summary

This comes to us from the [NOFO Builder retro](https://www.figma.com/board/ZQtWayFyCkBQcfg4j06wpG/NOFO-Builder-Jam?node-id=4880-5912&t=cLrlAvy2m5b3r7Yt-0):

> A navigation pane, like you see in a Word document so you can click and go straight there.

> a navigation pane in the builder, while looking for a section to edit it might be helpful to not have to jump to the top, similar to word’s navigation window.

> It takes a bit of scrolling to reach the actual NOFO content — could we tighten the top of the tool up?

There are lots of examples of this sort of thing.

In Word, there is a "navigation page" which shows all the headings in the doc. Same thing in Google Docs and in Substack.

| Word  |  Substack | Google Docs |
|--------|-------|-------|
|    <img width="710" alt="Image" src="https://github.com/user-attachments/assets/db4085be-9565-4bc7-a61d-3e87ef0743a6" />    |    <img width="287" alt="Image" src="https://github.com/user-attachments/assets/09c6f126-68f7-4f9f-99bf-64a4a54b6043" />   |  <img width="264" alt="Image" src="https://github.com/user-attachments/assets/c731223a-e2da-4767-a15a-cf826a5a4cb4" /> |

This is clearly a good and useful feature, so we should do something like this. Maybe we want to include subsection headings at some point (or, say h3s but not anything lower) but a good first step would be to include the Section headings and highlight them based on the page scroll position. 

We also have a mockup for this in the [Composer Figma mocks](https://www.figma.com/design/KKTTQiX1gWrQ03HAkB9kJg/A---NOFO-Composer-WIP-wireframes-3.2025?node-id=144-4317&t=SeKsuuvRWfNLhaec-0), which looks exactly like what we want. 

<img width="300" alt="Image" src="https://github.com/user-attachments/assets/3f830a53-6634-4be0-825c-be3b92e53e95" />

First step though, is to just add headings. If that works, then maybe we add subsections too. But at least initially we will just keep things simple.

pcraig3:
Here is a link to a Substack post which has the little side popout menu: https://geekway.substack.com/p/the-doge-next-door

And in the USWDS design pattern library, here are a couple of potentially useful designs:

- In-page navigation: https://designsystem.digital.gov/components/in-page-navigation/
[...]

PR #401, opened 2025-07-16
Add Side Navigation to NOFO Edit Page

dfitchett:
# Add Side Navigation to NOFO Edit Page

## Overview
This PR adds a collapsible side navigation component to the NOFO edit page, providing users with a table of contents for easier navigation through long NOFO documents.

## Related Issue(s)
- https://github.com/HHS/simpler-grants-pdf-builder/issues/278
## 🚀 Features Added

- **Collapsible sidebar** with smooth animations automatically populated from H2 headings
- **Toggle button** with visual pipe indicators showing current section
- **Smooth scrolling** to sections with active state tracking
- **Full accessibility support** with ARIA attributes, keyboard navigation, and focus management
- **Responsive design** with mobile-optimized full-screen overlay

### Files Modified/Added
- `nofos/bloom_nofos/templates/includes/side_navigation.html` - New navigation component
- `nofos/bloom_nofos/static/styles.css` - CSS styles for navigation
- `nofos/nofos/templates/nofos/nofo_edit.html` - Integration with edit page
- `nofos/nofos/tests_nofos/test_side_navigation.py` - Comprehensive test suite

### Key Features
- **Dynamic content population** from existing H2 headings with scroll position tracking
- **CSS transitions** for smooth animations with USWDS integration
- **Mobile-first responsive design** with 768px breakpoint
- **Accessibility compliance** with proper focus management and ARIA attributes

## 🧪 Testing
- Unit tests for HTML structure and template inclusion
- Accessibility tests for ARIA attributes and keyboard navigation
- Integration tests with existing NOFO sections

## 📋 Testing Instructions
1. Navigate to any NOFO edit page
2. Click the toggle button on the left side to open navigation
3. Test clicking navigation links and keyboard navigation (Tab, Escape)
4. Verify responsive behavior on mobile devices

## 🎯 Benefits
- **Improved navigation** for long NOFO documents
- **Better user experience** with non-intrusive design
- **Accessibility compliance** with Section 508 standards
- **Mobile-friendly** design that works across all devices

### DIFF
--- a/nofos/bloom_nofos/static/styles.css
+++ b/nofos/bloom_nofos/static/styles.css
@@ -1411,3 +1411,60 @@ del + ins > br {
   position: relative;
   top: -1rem;
 }
+
+/* Side Navigation Styles */
+.side-nav-container {
+    top: 50%;
+    pointer-events: none;
+    transition: transform 0.3s ease;
+    transform: translateY(-50%);
+}
+
+.side-nav-container.is-open {
+    pointer-events: auto;
+}
+
+.side-nav-toggle {
+    transform: translateY(-50%);
+    pointer-events: auto;
+    transition: left 0.3s ease;
+}
+
+.side-nav-pipe.current {
+    font-size: 1.25rem;
+    font-weight: bold;
+}
+
+.side-nav-content {
+    left: -330px;
+    width: 320px;
+    transition: left 0.3s ease, transform 0.3s ease;
+    transform: translateY(-50%);
+}
+
+.side-nav-container.is-open .side-nav-content {
+    left: 60px;
+    transform: translateY(-50%);
+}
+
+/* Navigation List Styles */
+#side-nav-list {
+    padding: 1rem;
+    margin: 0;
+    list-style: none;
+}
+
+/* Responsive Design */
+@media (max-width: 768px) {
+    .side-nav-content {
+        width: 100vw;
+        height: 100vh;
+        left: -100vw;
+        top: 0;
+    }
+
+    .side-nav-container.is-open .side-nav-content {
+        left: 0;
+        top: 0;
+    }
+}
--- a/nofos/bloom_nofos/templates/includes/side_navigation.html
+++ b/nofos/bloom_nofos/templates/includes/side_navigation.html
@@ -1,0 +1,255 @@
+<!-- Side Navigation Component -->
+<div id="side-nav-container" class="side-nav-container position-fixed left-1 z-top height-auto width-auto" aria-hidden="true">
+    <!-- Toggle Button -->
+    <button id="side-nav-toggle" class="side-nav-toggle position-fixed left-1 z-top padding-1 radius-md display-flex flex-align-center flex-justify-center text-primary border-primary bg-white hover:bg-gray-10 " aria-expanded="false"
+        aria-controls="side-nav-content" aria-label="Toggle table of contents navigation" type="button">
+        <div id="side-nav-pipes" class="side-nav-pipes display-flex flex-column flex-align-center gap-1">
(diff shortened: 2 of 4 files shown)


pcraig3:
Awesome, this looks great!

A few small things I am noticing:

1. sometimes you pick a heading and it scrolls down correctly but then highlights the _previous_ header
2. let's move the position of the menu closer to the top so that you tab to it sooner
3. let's remove tabbing from nav links when menu is closed
4. lots of headings behaviour (it works!)

### 1. Highlighting the previous heading.

Usually when you pick a heading in the nav menu, it scrolls to that place and keeps the item you just clicked in the nav menu highlighted. 👍 great, this is what we want. 

However, on some of them you see the _previous_ nav item become highlighted even though visually it seems like the scroll is working just fine. 

You can see it in this gif: whenever I select "Step 5", it scrolls there and then highlights "Step 4" instead.

Screen Recording 2025-07-17 at 10 44 46

Not a big deal, I think if you add a pixel or two to the scroll offset, it will probably fix this. 

### 2. Move the menu up in the HTML

To tab in and out of the menu, you have to be very close to the bottom of the page. We should move it up in the HTML so that you can tab into the menu near the top of the page, so that if you want to keyboard nav your way to Step 6 or something, you can do it.

Here's my suggestion:

Screenshot 2025-07-17 at 10 38 20


### 3. Do not tab through nav links when the menu is closed

I noticed this when I was doing the tabbing stuff ☝️. When you tab around even if the menu is closed, it goes through all of the links. Obviously this is not right.

Screen Recording 2025-07-17 at 09 30 19

When opening and closing the nav menu, maybe also add/remove the tabindexes on the nav links links?

### 4. What happens when there are lots of headings

Yesterday, I was speculating that this design could break when there are lots of headings. It turns out that it's really not too bad though, hooray! 

We pretty much always have the 7 steps, but writers can add appendices, and sometimes there are a fair number of them. I have an example NOFO with 11 appendices locally which is an outlier but good for testing this feature. 

| nav menu desktop view | nav menu mobile |
|--------|-------|
|  This screen is fine. Maybe we could make the menu wider but for now let's not.      |    This screen works   |
|  Screenshot 2025-07-17 at 10 50 18      |       Screenshot 2025-07-17 at 10 50 24 |

I don't think there's anything to do here, I just wanted to mention it. 

As a side note, I will send you the NOFO I am using in an email and you can try importing it via the API (I will send instructions for that as well).

pcraig3 (approved):
Hey man, I think it looks good. 

Looks like the "selects previous header" thing is fixed, and not tabbing through the offscreen links is fixed also.

I noticed a couple things this time that I didn't notice before. Neither are dealbreakers, I think we can just ship this and wait and see what happens. 

1. Layout shift can happen between selected and unselected state
2. Full keyboard nav is not really possible

### 1. Layout shift

Not really an issue, just an observation.

Sometimes when we bold the text, it takes 1-line nav items to 2 lines. This means that if you click twice in the same spot, you can end up clicking different menu items, like in this gif. (I can click 4 and 5 without moving the mouse because step 3 changes height when it is bold).

Screen Recording 2025-07-17 at 13 48 13

I would rank this 0/10 in terms of things we care about, but just wanted to note it.

### 2. Keyboard nav issue

Okay so now we can keyboard into the nav menu from the top, that's great. And then if we tab to "Step 6" and press enter, it scrolls down, great. But then if I want to then resume tabbing, I go right back up to the top because my focus is still on the nav menu, not on the thing I clicked on. 

I think a quick fix could be explicitly managing focus to the heading we just clicked. But then it's kind of a one-way thing, where you can't get back to the nav menu anymore. 

Idk, we can just leave it for now and wait for feedback or set focus explicitly — up to you.

Given all that, I am approving this PR.

PR #403, opened 2025-07-17
Improve side navigation functionality to NOFO editor

dfitchett:
# Improve Side Navigation Population in NOFO Editor

## Summary
Refactored the side navigation component to use server-side rendering instead of client-side JavaScript population, improving performance and reliability.

## Related Issue(s)
- https://github.com/HHS/simpler-grants-pdf-builder/issues/278

### Backend Changes
- **Added `get_side_nav_links()` function** in `nofos/nofo.py` to generate navigation links based on NOFO sections
- **Updated `NofosEditView`** to pass `side_nav_headings` to template context
- **Created `math_filters.py`** template tag with `multiply` filter for dynamic height calculations

### Frontend Changes
- **Enhanced side navigation template** to render navigation links and pipe characters server-side
- **Simplified JavaScript logic** by removing client-side navigation population
- **Added conditional rendering** to hide navigation when no sections exist

### Testing
- **Added comprehensive tests** for `get_side_nav_links()` function
- **Enhanced side navigation tests** to verify correct HTML element generation
- **Added test coverage** for NOFOs with no sections

## Technical Details
- Navigation links are now generated from NOFO sections in the view layer
- Template conditionally renders navigation only when sections exist
- JavaScript focuses on interaction logic (toggle, scroll tracking) rather than content generation
- Maintains all existing functionality while improving underlying architecture

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@ Versioning since version 1.0.0.
 
 ### Added
 
+- Add left-hand floating menu for in-page navigation of a NOFO's major (H2) headings
+
 ### Changed
 
 - Replace multiple icons in a table cell, should it come to that
--- a/nofos/bloom_nofos/static/styles.css
+++ b/nofos/bloom_nofos/static/styles.css
@@ -1430,8 +1430,13 @@ del + ins > br {
     transition: left 0.3s ease;
 }
 
+.side-nav-pipe {
+    height: 1.25rem;
+    max-height: 1.25rem;
+    overflow: hidden;
+}
+
 .side-nav-pipe.current {
-    font-size: 1.25rem;
     font-weight: bold;
 }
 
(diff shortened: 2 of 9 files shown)


pcraig3:
Made some CSS changes and one JS change. 

CSS changes:

- On mobile view, body of container was still visible, so I moved it away by adjusting left value by -1vw
- Added `cursor: pointer` to menu toggle
- Move the "transition" CSS rule for the `.side-nav-content` object to only when it is open because otherwise there was a weird flash of content when moving to mobile
- Made the little em dashes bigger when `.current` without moving their vertical position
- Switched to using `em` for media query because that's the convention

JS change: 

- Manually move focus back to `.nav-button-toggle` after clicking the Close button (this is to avoid getting a console error about focus remaining on an offscreen element (it would otherwise remain on the close button, which goes offscreen).

PR #411, opened 2025-07-23
Swap image names for diff-match-patch

pcraig3:
Small change, but I misnamed these images initially.


pcraig3:
Merging because it is just documentation

PR #415, opened 2025-07-31
Rewrite diffing engine + add new diff page

pcraig3:
## Summary

This PR gets us 90% of what we want for user testing the diff compare screen. Sorry for the massive PR but it couldn’t really be helped.

Some pretty major code changes here:

1. Complete rewrite of our diffing algorithm
2. Compares against a real-looking NOFO render
3. Allows for a side-by-side diff or a consolidated diff
4. Changes the default comparison to all text and all headings (this means we will have noisy diffs)

### Screenshots 

| side-by-side diff | consolidated diff ("single view") |
|--------|-------|
|    (ignore the weird floating header, this actually works fine). Two-column comparison using our HTML-render styles for the NOFO. Your original NOFO on the left, your new NOFO on the right.   |   Single-column comparison using our HTML-render styles for the NOFO.  All changes are in the one column: deletions and new additions.       |
|    localhost_8000_guides_4e94136d-1b81-4c97-bfe5-11f3af5b5ec6_compare_df996baf-ae6f-4a51-835f-8936c26afdff    |  localhost_8000_guides_4e94136d-1b81-4c97-bfe5-11f3af5b5ec6_compare_df996baf-ae6f-4a51-835f-8936c26afdff_display=single     |

#### 1. Complete rewrite of our diffing algorithm

The basic problem with the previous diffing function (`nofos/nofo_compare/html_diff.py`) was that it was very simple (we like simple good, but it means it had limitations).

Since we want to render a real-looking NOFO, this means we have to show them to the user in HTML instead of showing the markdown text as we do in other places. _But_ if we are diffing markdown tables, or lists, we can get really weird stuff that breaks the markdown render and then they aren’t properly converted to HTML. 

eg, we can get stuff like this:

```
<del>+</del><ins>*</ins> List item 1
<del>+</del><ins>*</ins> List item 2
```

☝️ this would break markdown rendering of this list, for example.

So we can’t diff them as markdown and _then_ convert to HTML. 

Instead, we have to just diff lines of HTML against each other. This is reasonably challenging, as adding or removing an element early on can create a crazy diff that is noisy and unintuitive. 

| before | after | naive diff | what the diff should look like |
|--------|-------|--------|-------|
|  list before      |   list after inserting item in position 2    |  what the diff looks like naively |  what the diff _should_ look like |
|  <pre>* One <br>* Two <br>* Three<br>* Four<br>* Five</pre>    |   <pre>* One <br>* One and a half <br>* Two <br>* Three<br>* Four<br>* Five</pre>    |  <pre>* One <br>(ADD)* One and a half (DELETE)* Two<br>(ADD)* Two (DELETE)*  Three<br>(ADD)* Three (DELETE)* Four<br>(ADD)* Four (DELETE)* Five <br>(ADD)* Five</pre> |  <pre>* One <br>(ADD)* One and a half <br>* Two <br>* Three<br>* Four<br>* Five</pre> |

☝️ Eg, if you have a list of 10 items and you insert a new item in the second position, a naive “loop through elements and compare” will end up telling you that you have 9 deletions and 10 additions. But in reality, there is only 1 addition: the new list item you added. So the diffing engine I wrote tries to make up for stuff like this. It’s not perfect, but I think for what we are likely to see it should be okay.

#### 2. Compares against a real-looking NOFO render

I mentioned this previously, but when we user tested the previous compare, it would compare the markdown output of each NOFO. In testing, we got a very strong reaction that people want to see something that looks like the thing they were just looking at: the Word document. For us, this means rendering it in “HTML” view and running the diff against the HTML. Already written about in the last section ☝️ . But this is the reason we changed how they are rendered.

### 3. Allows for a side-by-side diff or a consolidated diff

This is another thing that jumped out in user testing: lots people want to see a side-by-side version. In terms of a mental model, it’s really clean. There are 2 documents and you want to show changes between them. Showing people one document with both sets of changes is confusing to lots of people. So now we can show the “before” and “after” docs or the doc that includes all changes: you pick.

The default comparison is side-by-side.

### 4. Changes the default comparison to all text and all headings (this means we will have noisy diffs)

Changes the default comparison strategy to all text and all headings per subsection.

This is a minor point compared to the others, but it is just a note to say that by default we will compare 100% of your NOFO content to the next thing you upload. We have the ability to change what kind of comparison we do, but for this initial user testing, we aren’t really going into the “edit your Content Guide” screen so much, just turning on and off comparisons, not really configuring them. 

### How to test this

| document 1 | document 2 |
|--------|-------|
|  original document      |   modified document    |
|  [Kansas_short_1.docx](https://github.com/user-attachments/files/21532067/Kansas_short_1.docx)      |   [Kansas_short_2.docx](https://github.com/user-attachments/files/21532068/Kansas_short_2.docx)   |

1. Manually visit the "Guides" route at [http://localhost:8000/guides/](http://localhost:8000/guides/)
2. "Upload content guide" button 
3. Select "Kansas_short_1"
4. Pick a name (doesn't matter)
5. Now you should be on the initial "Check out the differences" page, with only 1 NOFO
6. Scroll and click "Upload a NOFO to compare"
7. Select "Kansas_short_2"
8. Now you get the side-by-side comparison. Look at it
9. Single-page comparison. Look at it also.
10. That's it for now!

All the changes are described in the text. The diff should clearly show what the changes are. You can also play with it and see how it behaves with new changes.   It shouldn’t be too hard to get it into a weird diff, but because we can control this during testing, I think it suffices. 

### Notes:

- Still have to overhaul the Content Guide edit page
- The Next and Previous buttons don’t do anything
- There is no “Fullscreen” 
- I still need to build a CSV export (I think people will hate it)
- The side menu should probably track scrolling.
- The nofos are not synced up at all, so if they are different lengths, they are just different sizes. I feel like this may have to change at a later point.

Probably there are other issues here, but this is a good start. 

Note that we don’t have tests for the `/compare` route specifically. What we are testing is all the smaller parts of the compare logic, which is split between `bloom_nofos/html_diff.py` and `nofos/compare_nofos.py`

Last note is that my ADR is no longer accurate: [2025-07-23-html-diffing-library.md](https://github.com/HHS/simpler-grants-pdf-builder/blob/main/documentation/adr/2025-07-23-html-diffing-library.md). I ended up using [`diff-match-patch`](https://pypi.org/project/diff-match-patch/) for the diffs, so I will eventually have to come back and rewrite that thing.


nofos/bloom_nofos/html_diff.py
```diff
@@ -0,0 +1,176 @@
...
+DIFFABLE_TAGS = {"p", "li", "td", "th", "h1", "h2", "h3", "h4", "h5", "h6", "div"}
+BLOCK_WRAPPER_UNSAFE_TAGS = {"li", "td", "th", "tr", "thead", "tbody"}
+
+
+def has_diff(diff_string):
+    return "<ins>" in diff_string or "<del>" in diff_string
+
+
+def html_diff(original_html, modified_html):
+    dmp = diff_match_patch()
```
dfitchett, line 16: You could have some input validation and return here early
```python
    # Input validation
    if original_html is None:
        original_html = ""
    if modified_html is None:
        modified_html = ""
    
    # Handle identical content early
    if original_html == modified_html:
        return original_html
```

nofos/bloom_nofos/html_diff.py
```diff
@@ -0,0 +1,176 @@
...
+
+    # Treat as plain text if no tags
+    if not original_html.strip().startswith(
+        "<"
+    ) and not modified_html.strip().startswith("<"):
+        return diff_plaintext_normalize_whitespace(original_html, modified_html, dmp)
+
+    soup1 = BeautifulSoup(original_html, "html.parser")
+    soup2 = BeautifulSoup(modified_html, "html.parser")
```
dfitchett, line 25: Could be worth doing a try catch here to catch invalid html, then you could fallback to plaintext diff: 
```python
try:
    soup1 = BeautifulSoup(original_html, "html.parser")
    soup2 = BeautifulSoup(modified_html, "html.parser")
except Exception as e:
    # Fallback to plain text comparison if HTML parsing fails
    logger.warning(f"HTML parsing failed, falling back to text diff: {e}")
    return diff_plaintext_normalize_whitespace(original_html, modified_html, dmp)
```

dfitchett (approved):
This looks awesome! Great work! 🔥 

A couple things not blocking because this really still a WIP.
* On the initial import content guide screen, if you click upload without putting in a file, it actually sends you to `/nofos/import` screen with the validation error, then when you upload, its uploads the content guide as a nofo. I think its due to that view extending from `nofos/views.py` `BaseNofoImportView`
* When uploading the second compare guide, if you dont add a file, the button will still take you back to the content guide compare page. When clicking `Upload a NOFO to compare` a second time, the validation error will show. 
image

nofos/bloom_nofos/html_diff.py
pcraig3: Added 👍

PR #427, opened 2025-08-07
Add error feedback for compare select

dfitchett:
- Enhanced the handling of subsection IDs by validating UUIDs and collecting failed IDs.
- Implemented bulk updates for comparison types to optimize performance.

### DIFF
--- a/nofos/guides/templates/guides/guide_edit.html
+++ b/nofos/guides/templates/guides/guide_edit.html
@@ -124,7 +124,7 @@
                 <td class="nofo-edit-table--subsection--body">
                   {{ subsection.body|safe_markdown|add_classes_to_tables|convert_paragraphs_to_hrs|get_value_or_none:"content" }}
                 </td>
-                <td class="width-4 nofo-edit-table--subsection--manage">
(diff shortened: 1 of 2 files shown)


nofos/guides/templates/guides/guide_edit.html
```diff
@@ -293,38 +289,62 @@ <h3 class="usa-accordion__heading usa-accordion__compare-heading">
...
-      // TODO: show a failure alert
+    function showMessage(data) {
+            // Create a simple message div
+            const messageDiv = document.createElement('div');
+            messageDiv.className = `usa-alert usa-alert--${data.status === 'success' ? 'success' : data.status === 'fail' ? 'error' : 'warning'} usa-alert--slim margin-top-2`;
+            messageDiv.innerHTML = `
+                <div class="usa-alert__body">
+                    <p class="usa-alert__text">${data.message}</p>
+                </div>
+            `;
```
pcraig3, line 332: Let's focus this div if it shows up. I think we may need a "tabindex" to do that.

pcraig3 (approved):
Same as before: I added one comment but am pre-approving!

This looks great and should be way faster to complete for those "all" checkboxes. 👍

PR #428, opened 2025-08-07
Add Section ("Step") name to CSV output

pcraig3:
## Summary

This PR changes our CSV files to include the section name as the second column. 

Since we are wanting users to see the word "Sections" for our subsections, I called the section.name "Step name" and the subsection.name is a "Section name" now.

So, the old column headings looked like this:

```
Status, Subsection name, Old value, New value
```

And new ones look like this: 

```
Status, Step name, Section name, Old value, New value
```

### DIFF
--- a/nofos/guides/test_views.py
+++ b/nofos/guides/test_views.py
@@ -566,7 +566,12 @@ class ContentGuideDiffCSVViewTests(TestCase):
         subsection.old_value = ""
         subsection.new_value = "New body"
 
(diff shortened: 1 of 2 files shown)


nofos/guides/test_views.py
dfitchett, line 1: Nice test additions 👍

dfitchett (approved):
EZ win 🚀

PR #434, opened 2025-08-08
Disabled buttons when no section/subsection selections are made in guide_edit page

dfitchett:
- Added role="button" to links in the guide edit template for improved accessibility.
- Implemented logic to disable buttons when no sections are selected, enhancing user feedback with a warning message.
- Updated CSS to prevent pointer events on disabled links, ensuring a consistent user experience.

| Before | After | 
| --- | --- | 
| Screenshot 2025-08-08 at 11 47 55 AM | Screenshot 2025-08-08 at 11 47 47 AM |

### DIFF
--- a/nofos/guides/templates/guides/guide_edit.html
+++ b/nofos/guides/templates/guides/guide_edit.html
@@ -44,13 +44,13 @@
   <ul class="usa-button-group margin-top-2 margin-bottom-4">
     <li class="usa-button-group__item">
       {% if new_nofo %}
-        <a role="button" class="usa-button" href="{% url 'guides:guide_compare_result' guide.id new_nofo.id %}">Return to comparison</a>
+        <a class="usa-button" href="{% url 'guides:guide_compare_result' guide.id new_nofo.id %}">Return to comparison</a>
       {% else %}
-        <a role="button" class="usa-button" href="{% url 'guides:guide_compare_upload' guide.id %}">Upload NOFO to compare</a>
+        <a class="usa-button" href="{% url 'guides:guide_compare_upload' guide.id %}">Upload NOFO to compare</a>
       {% endif %}
     </li>
     <li class="usa-button-group__item">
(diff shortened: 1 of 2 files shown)


nofos/guides/templates/guides/guide_edit.html
```diff
@@ -44,13 +44,13 @@ <h1 class="font-heading-md margin-top-1">{{ guide.title }}</h1>
...
-        <a class="usa-button" href="{% url 'guides:guide_compare_result' guide.id new_nofo.id %}">Return to comparison</a>
+        <a role="button" class="usa-button" href="{% url 'guides:guide_compare_result' guide.id new_nofo.id %}">Return to comparison</a>
       {% else %}
-        <a class="usa-button" href="{% url 'guides:guide_compare_upload' guide.id %}">Upload NOFO to compare</a>
+        <a role="button" class="usa-button" href="{% url 'guides:guide_compare_upload' guide.id %}">Upload NOFO to compare</a>
       {% endif %}
     </li>
     <li class="usa-button-group__item">
-      <a class="usa-button usa-button--outline" href="{% url 'guides:guide_import' %}">Replace this content guide</a>
+      <a role="button" class="usa-button usa-button--outline" href="{% url 'guides:guide_import' %}">Replace this content guide</a>
```
pcraig3, line 53: I don't think we need these here. 

While these are _visually_ styled like buttons, they are _functionally_ just links. You click on them and they take you to a different page. Usually buttons are for processing or submitting something (eg, upload your nofo), which indicates there is some kind of data thing happening on the backend. 

These are just links that we want to make more prominent with visual button styling. But if you are using a screenreader, it would actually be misleading to be like "you a clicking a button right now", rather than just saying: "hey, this is a link".

pcraig3 (approved):
Hello! I left a comment about removing `role="button"` — I don't think we need them for these links, but otherwise it looks good 👍

PR #436, opened 2025-08-11
Make the criteria to "merge" subsections harder

pcraig3:
## Summary

This PR solves a bug where two subsections in our diff were getting matched that were not the same at all. It essentially changes the bar for "matching subsections" to include 3 contiguous characters, where before it was only 1 character in common.

### Details 

Previously, we would merge subsections (as in, classify a separate `ADD` and `DELETE` as an `UPDATE`) if there was any matching text at all in the headings.

This failed recently on a couple of very unlike subsections.

Subsection 1: `Cost-sharing commitments`
Subsection 2: `Program focus`
Diff: `<del>Cost-sharing commitment</del><ins>Program focu</ins>s`

Because there was a shared "s", they were grouped together, leading to a very confusing diff.

In reality, these subsections should _not_ match because they are totally different and their body text is also totally different.

The new rule to this effect now changes the match criteria to look for 3 contiguous characters in the heading, otherwise they don't match.

### Screensaver

| before | after |
|--------|-------|
|  These two subsections were being matched together      |     Program focus and Cost-sharing commitments are now separated  |
|  Screenshot 2025-08-09 at 19 11 18      |   Screenshot 2025-08-09 at 19 12 00    |

### DIFF
--- a/nofos/guides/views.py
+++ b/nofos/guides/views.py
@@ -387,8 +387,12 @@ class ContentGuideCompareView(GroupAccessObjectMixin, LoginRequiredMixin, View):
 
             # Filter out sections that contain ONLY "ADDs".
             comparison = [
-                section for section in comparison
-                if not all(sub.comparison_type == "body" and sub.status == "ADD" for sub in section["subsections"])
+                section
+                for section in comparison
+                if not all(
+                    sub.comparison_type == "body" and sub.status == "ADD"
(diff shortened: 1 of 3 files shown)


pcraig3:
I'm actually just going to merge this because we are going into testing at 10:30 and this is a bug that the last user noticed.

PR #451, opened 2025-08-26
Group multiple subsection deletions in diff before later matching subsections

pcraig3:
## Summary

This PR changes the order of how multiple deletions in a row are shown in our diff interface. Grouping them together is more intuitive and solves a problem we observed in user testing. 

### Details

Previously, we would show the deleted subsections interleaved with matched subsections because our ordering algorithm would basically do: 

- check old document subsection
- check new document subsection
- check old document subsection+1
- check new document subsection+1
- old+2
- new+2
- etc. 

So if we had multiple deletions in a row they would get interleaved with Matched subsections from the new document. 

So, for example, if we had this old document:

```
Section 1
Section 1.1
Section 1.2
Section 1.3
Section 2
Section 3
Section 4
Section 5
```

And this new document (Sections 1.1-1.3 deleted):

```
Section 1
Section 2
Section 3
Section 4
Section 5
```

We would end up with this as the result:

```
MATCH: Section 1
MATCH: Section 2
DELETE: Section 1.1
MATCH: Section 3
DELETE: Section 1.2
MATCH: Section 4
DELETE: Section 1.3
MATCH: Section 5
```

This is unintuitive and confused multiple people during our recent round of user testing. Here's what we _should_ present instead:

```
MATCH: Section 1
DELETE: Section 1.1
DELETE: Section 1.2
DELETE: Section 1.3
MATCH: Section 2
MATCH: Section 3
MATCH: Section 4
MATCH: Section 5
```

In this version ☝️ , all of the deletions are grouped _before_ the next match rather than being interspersed among them. 

This new logic implements this change so hopefully it solves this confusing issue without causing new issues 😬 . 

And finally, I'll just note that multiple `ADD`s did not have this same problem, so they are unaffected. They have always been working as you would expect.

#### Original Word documents

| Old document | New document |
|--------|-------|
|  Includes "1.1", "1.2", "1.3"      |  Goes from 1 to 2, no increments     |
|  Screenshot 2025-08-26 at 13 45 01      |   Screenshot 2025-08-26 at 13 45 29    |

#### Diff: OLD document vs NEW document (Subsections "1.1", "1.2", "1.3" are DELETED)

| before | after |
|--------|-------|
|  deletions are interleaved with matches      |    deletions are all grouped before the next match   |
|    Screenshot 2025-08-26 at 13 44 21    |   Screenshot 2025-08-26 at 13 42 31    |

#### Diff: NEW document vs OLD document (Subsections "1.1", "1.2", "1.3" are ADDED now)

| before | after |
|--------|-------|
|   No change!     |    No change!   |
|     Screenshot 2025-08-26 at 13 46 04   |   Screenshot 2025-08-26 at 13 46 23    |

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,8 @@ Versioning since version 1.0.0.
 
 ### Fixed
 
+- Show deletions in compare page in order
+
 ## [3.8.0] - 2025-08-21
 
 ### Added
--- a/nofos/nofos/nofo_compare.py
+++ b/nofos/nofos/nofo_compare.py
@@ -163,48 +163,60 @@ def compare_sections(old_section, new_section):
     # Get all subsections for comparison
     new_subsections = list(new_section.subsections.all())
     old_subsections = list(old_section.subsections.all()) if old_section else []
-    max_length = max(len(new_subsections), len(old_subsections))
-    matched_subsections = set()
 
+    matched_subsections = set()
     subsections = []
 
-    for index in range(max_length):
-        new_sub = new_subsections[index] if index < len(new_subsections) else None
-        old_sub = old_subsections[index] if index < len(old_subsections) else None
-        # First, check the new subsection for a match
-        if new_sub:
-            matched_old_sub = find_matching_subsection(
-                new_sub, old_subsections, matched_subsections
-            )
-            if matched_old_sub:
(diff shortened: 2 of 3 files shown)


pcraig3:
Merging this and @admoorgit can check it out later.

Issue #446, opened 2025-08-20
The subsection content does not load on the admin page for a subsection


pcraig3:
Instead we get a JS error. 

It has something to do with JS, and it appeared when we updated Martor, so I will have to look into it.

<img width="910" height="693" alt="Image" src="https://github.com/user-attachments/assets/934f2d38-95e0-41eb-b166-b0d342c40782" />

pcraig3:
This is closed since #452 went in

PR #452, opened 2025-08-27
Update poetry dependencies + fix martor bug 🐞 

pcraig3:
## Summary

This PR updates all of our poetry libraries (except for `mammoth`, which has a bug (see #445)), and solves a bug we had with the latest version of martor.

### Details 

Part of martor's recent updates removed a JS-based resizing functionality for a lighter, simpler, CSS resizing. That's great, we love that. However, it was leading to a bug in the admin console telling me: `Uncaught TypeError: e(...).resizable is not a function`. 

This meant that we couldn't edit subsection content in the admin view, because the content would never populate: https://github.com/HHS/simpler-grants-pdf-builder/issues/446 

The issue turned out to be with my static folder, rather than something martor was doing wrong. 

When I first installed martor, I manually copied its static folder into mine, and we have been using those ever since. As part of martor's new update, some of the static files have been rewritten to no longer use `resizable`, but _my_ static files were still the old ones and were still expecting it to exist. Eventually I figured out that I needed to also copy the contents of `martor/static` into my static folder at `nofos/bloom_nofos/static` and then everything started to work. 

As part of this change, we can also remove the `resizable.min.js` include statements we had for the subsection edit page and the subsection create page. 

Updating martor also meant we were able to update `markdown`, which had been a problem for us in the past (see #257).

Overall, solid update.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -260,103 +260,90 @@ files = [
 
 [[package]]
 name = "charset-normalizer"
-version = "3.4.2"
+version = "3.4.3"
 description = "The Real First Universal Charset Detector. Open, modern and actively maintained alternative to Chardet."
 optional = false
 python-versions = ">=3.7"
 files = [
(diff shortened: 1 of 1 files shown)


pcraig3:
Merging this issue solves https://github.com/HHS/simpler-grants-pdf-builder/issues/446

Issue #489, opened 2025-10-27
Add a "Preview" screen for NOFO Composer


pcraig3:
After setting up your NOFO, you should be able to preview it. 

There is a mock for this: basically we can add one more item to the nav menu and use our HTML formatting to lay out the content guide the way it looks in our HTML view. 

<img width="519" height="673" alt="Image" src="https://github.com/user-attachments/assets/5daf7ce4-293e-4eda-8e3d-fe5a50e9dc15" />

Issue #486, opened 2025-10-27
Add body text formatting


pcraig3:
This story is fairly straightforward (hopefully!).

In the Word version of Content Guides, different (Sub)sections are designated as editable or not editable using an obscure Word feature called “Content Controls”. 

However, if you have 2 identically-formatted sentences where one sentence is editable and one is not editable, there is no way for a writer to tell this from just looking at it, which creates an incredibly frustrating experience for NOFO writers. To accommodate for this, we have come up with standard formatting for content guides that signals to writers if text is editable, and, if so, how. 

What we generally use us:

- Black text: Text is not editable
- Green text: text is completely editable. Sometimes this means that you can rewrite the subsection content, sometimes this means that you can delete the subsection if it is not relevant to you, but you can’t change the content, so this is oversignified. (The distinction is always explained in the instructions, but overall it is not ideal)
- Yellow highlighted text: variables. The yellow highlighted text is editable, whereas the surrounding text isn’t. Sometimes this is free text, sometimes select boxes, but we are going to assume in our app that it is always free text.

| black text | green text | yellow text | mixed formatting |
|--------|-------|-------|-------|
| non-editable text. Leave it alone.       |   Fully editable text. Change whatever you want (sometimes).    | "Variables" that you can enter in specific areas |  Multiple types of formatting. We are not planning to support this. |
|    <img width="665" height="282" alt="Image" src="https://github.com/user-attachments/assets/9db502ee-f31c-45c4-9200-db730548da0f" />    |    <img width="665" height="181" alt="Image" src="https://github.com/user-attachments/assets/f7726754-e949-435a-897b-fe8084314e94" />   |    <img width="665" height="212" alt="Image" src="https://github.com/user-attachments/assets/68040b14-d9d7-4e9c-9887-5064da7bd02c" />    |    <img width="665" height="174" alt="Image" src="https://github.com/user-attachments/assets/4959c574-2793-4811-b3fe-d1de91f8ba20" />    |

To extend the continuity from the formatting people have come to expect, we should implement this same formatting inside of our ContentGuideSection overview pages. We have 3 places where we have to do this (for System admins):

- In section Overview page (see screenshot below)
- In Editing page in “preview” pane
- In “Preview” mode (not yet built)

Here is an example of this in the Figma mocks (we haven't implemented this yet):

[...]

PR #495, opened 2025-10-29
Create NOFO Composer preview pane

pcraig3:
# Summary 

This PR puts up a "preview" page so that we can see (more or less) what the NOFO would look like as a Word document. This is as close as we are gonna get to the "export" feature.

# Description 

We have a couple of views already where we are showing the NOFO in a view that are meant to kind of look like something else. 

- NOFO HTML view (Looks like the PDF)
- NOFO Compare view (Looks like Word)

And now 

- NOFO Composer preview  (Looks like Word)

| NOFO HTML view | NOFO Compare view | NOFO Composer preview  |
|--------|-------|-------|
|  (It's a little messed up because it's a content guide not a NOFO)      |   NOFO compare panel    |    ✨ New!! ✨   |
|    Screenshot 2025-10-29 at 17 09 39    |    Screenshot 2025-10-29 at 17 09 23   |     Screenshot 2025-10-29 at 17 08 24  |

It's not completely done because there are some buttons that we should eventually put there, and we also eventually have to do body text formatting one day (#486), but it is good enough to ship!

- Added a new View, Url, and page template for this
- Added simple tests
- Added link to preview page from other sections

As far as testing this PR goes, I would say just see that it works and scroll through for any weirdness, but there's not really any logic going on right now, so there's no complex interaction worth testing.

Merging this resolves issue #489

### DIFF
--- a/nofos/composer/templates/composer/composer_preview.html
+++ b/nofos/composer/templates/composer/composer_preview.html
@@ -64,6 +64,11 @@
 
             <div class="section--content">
               {% for subsection in section.ordered_subsections %}
+                  {% if subsection.instructions %}
+                    <div>
+                      {{ subsection.instructions|safe_markdown|add_classes_to_paragraphs|add_classes_to_lists|convert_paragraphs_to_hrs }}
+                    </div>
+                  {% endif %}
                   {% with tag=subsection.tag content=subsection.name id=subsection.html_id class=subsection.html_class %}
                     {% include "includes/heading.html" with tag=tag content=content id=id class=class only %}
                   {% endwith %}


nofos/composer/templates/composer/composer_preview.html
```diff
@@ -0,0 +1,80 @@
...
+      <hr />
+
+      <div class="nofo_view">
+        <!-- Render each section with all its subsections -->
+        {% for section in sections %}
+          <section id="section--{{ section.html_id }}" class="section section--{{ forloop.counter }} {% if forloop.counter > 7 %}section--appendix {% endif %}section--{{ section.html_id }}{% if section.html_class %} {{ section.html_class }}{% endif %}">
+            <h2 id="{{ section.html_id }}" >{{ section.name }}</h2>
+
+            <div class="section--content">
+              {% for subsection in section.ordered_subsections %}
```
jlhogantech: Should this preview include subsection instructions?

pcraig3: Yes, it should. Good point.

Issue #484, opened 2025-10-27
Add 'instructions' to ContentGuideSubsections on import


pcraig3:
Real-world NOFOs typically start life as Content Guides. Content Guides are basically templated documents which NOFO writers use as a starting point to begin writing the full text of their NOFO. Because NOFOs follow a roughly standardized order (Step 1: Review the Opportunity, Step 2: Get Ready to Apply, etc), and because some of the sections in NOFOs are legally mandated standardized language, there is a strong skeleton already in place when your start writing your NOFO. As a writer, you are putting meat on bones, not making a jello.

Okay.

What follows from this is that a Content Guide and a NOFO are _very_ similar in practice. They are large-ish documents that can be usefully subdivided into “Steps” (we call them “Sections”) and those Sections have headings that we use to subdivide them into “Sections” (we call them “Subsections”).   At a high level, this data structure describes both a NOFO and a Content Guide.

```
-> Nofo
  -> (has 1 or more) Sections
    -> (has 1 or more) Subsections
```

Because of this, our Composer product/feature/epic reuses the data model and logic from importing NOFOs when we import Content Guides.

For example, in the `nofos` app, we have the following:

- [`class BaseNofo`](https://github.com/HHS/simpler-grants-pdf-builder/blob/main/nofos/nofos/models.py#L89)
- [`class Nofo(BaseNofo)`](https://github.com/HHS/simpler-grants-pdf-builder/blob/main/nofos/nofos/models.py#L217)

And in our (new) `composer` app, we have:

- [`class ContentGuide(BaseNofo)`](https://github.com/HHS/simpler-grants-pdf-builder/blob/main/nofos/composer/models.py#L16)

For the most part it works, but one part that is legitimately different between each type of document is the concept of “Instructions for Writers”.

The instructions for writers are (as you would expect) editorial instructions telling the writers what to do in each Subsection. So they are like: “Hey writers, add these pieces of information in this section. Do not meddle with this language though, etc”. They are included in Content Guides because System Admins want the Writers to read them before filling out the NOFO. Once the writers have filled out the NOFO, the instructions should go away, since they are not meant for applicants. 

Here is an example from a Content Guide of a Subsection preceded by instructions.

<img width="1212" height="336" alt="Image" src="https://github.com/user-attachments/assets/feb0cb58-33a3-4dfb-a592-f502d8cc20b3" />

[...]

PR #498, opened 2025-10-30
Part 2: add ability to edit instructions

jlhogantech:
resolves https://github.com/HHS/simpler-grants-pdf-builder/issues/484

- Fix styles for edit view instructions, and add ability to edit
- add a fix for matching instructions to subsections without a name, and a test for the case

| Composer edit view with instructions | Composer instructions edit view | 
| --- | --- | 
| Screenshot 2025-10-30 194835 | Screenshot 2025-10-30 194848 |

### DIFF
--- a/nofos/nofos/test_nofo.py
+++ b/nofos/nofos/test_nofo.py
@@ -6840,6 +6840,7 @@ class DecomposeInstructionsTablesTest(TestCase):
         self.assertEqual(len(removed_tables), 1)
         self.assertIn("Instructions for NOFO writers", removed_tables[0].get_text())
 
+
 class AddInstructionsToSubsectionsTests(TestCase):
     def test_add_instructions_to_subsections(self):
         sections = [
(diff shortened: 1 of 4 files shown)


nofos/nofos/test_nofo.py
jlhogantech: Most of these changes are just whitespace/formatting, sorry! Turns out I did not actually run the precommit stuff on my prev PR 😞 did add one net new test for the no-name subsection, which I will call out with a comment below

nofos/nofos/test_nofo.py
```diff
@@ -6895,31 +6901,65 @@ def test_add_instructions_to_subsections_matches_subsequent_duplicates(self):
...
+        assert (
+            "Instructions for Section B"
+            in sections[0]["subsections"][1]["instructions"]
+        )
+        assert (
+            "Instructions for Section A"
+            in sections[0]["subsections"][2]["instructions"]
+        )
+
+    def test_add_instructions_to_subsections_skips_subsections_with_no_title(self):
```
jlhogantech: this is the actual change -- new test added for additional edge case

pcraig3:
> Currently, the instructions are HTML, which is formatted nicely by the martor preview stuff but maybe harder for content guide writers to edit. Should the instructions HTML be parsed into markdown before exposing to users?

Yes, HTML is a good markup language for rendering content but it is horrible for editing. 

We have this same issue with the subsection `body`, which comes in as HTML and we convert to markdown. 

The code where we handle this is here: https://github.com/HHS/simpler-grants-pdf-builder/blob/main/nofos/nofos/nofo.py#L374-L387

So let's do the same thing for instructions

The conversion back from markdown should already be working because of the `safe_markdown` tag we are using in the template.

pcraig3:
> Currently, the instructions are HTML, which is formatted nicely by the martor preview stuff but maybe harder for content guide writers to edit. Should the instructions HTML be parsed into markdown before exposing to users?

Actually, I didn't realize this before, but these instructions are being saved as tables because that's how they are structured in the Word document (they are a 1-cell table in the Word docs).

I think we should be stripping the table semantics away, because it is not necessary and will make it confusing to edit in markdown. A 1-cell table is a shortcut we use in Word so that it is easy for the NOFO Builder to find these instructions and remove them. It has no semantic meaning and it shouldn't be a table.

//

Secondly, I am wondering if we should also delete the first line of the Instructions and just construct the "Instructions for NOFO Writers: {subsection.name}" ourselves using the subsection name. There are definitely some edge cases here, however, as not all subsections have names, and some Instructions titles include more than 1 subsection, so we would be changing the intent if the heading changed. On this aspect of things I am not sure what the right way to play it is, I could be swayed in either direction.

pcraig3 (changes requested):
Not that much from me, but I have a comment outside of this review about stripping the table semantics out from the instructions and then converting its contents to markdown.

jlhogantech:
Most recent change:
1. Adds instructions_edit to css rules for correct font family 
2. Extracts instructions as markdown, removing the wrapping one-cell table

does NOT extract instructions title as separate, uneditable field because I'm pretty sure that would require a migration and is a change to the existing ContentGuideSubsection model, and i wanted to confirm that's what we want to do

pcraig3 (approved):
It's looking good to me! The only question/comment I had was around instructions titles, so I left a thorough comment about it above. ☝️ 

In the end, I am okay with whatever the call is on this, so merge when you are ready.

PR #499, opened 2025-10-31
Better labels, better errors

pcraig3:
## Summary

This PR makes 4 changes, all minor:

1. Changes the labels on the ContentGuideSubsection accordions to match Annie's designs
2. Defaults the ContentGuideSubsection.edit_mode to "locked"
3. Add success and error messages when importing or deleting a content guide
4. Don't hide subsections in groups if they have instructions

#### 1. Changes the labels on the ContentGuideSubsection accordions to match Annie's designs

Annie mentioned that the labels were all different, so we are making them match. It's probably good to have some separation between the database value and the label in any case.

| before | after |
|--------|-------|
|      Screenshot 2025-10-31 at 17 43 23  |     Screenshot 2025-10-31 at 17 42 49  |

#### 2. Defaults the ContentGuideSubsection.edit_mode to "locked"

This is more of a "sensible defaults" question. The intent of the Composer is to provide a minimal set of editing options for writers who then export the doc and edit it as a Word doc with their team. Setting the default to "full text" editing makes it seem like we expect writers to write the whole NOFO in the browser, which is NOT what we want. So hopefully setting everything to "locked" by default has the effect of implying that.

#### 3. Add success and error messages when importing or deleting a content guide

Right now, we create an red 'destructive' `message` when we delete a Content Guide in the composer, but we don't actually look for that message on our index page. So the message gets shown on the "import" page as a validation message, which is very weird. 

So, this commit adds a success message for a successful import and an answer message for a deleted document at the top of the composer index page.

| before (after delete) | after (importing) | after (deleting) |
|--------|-------|----|
|  deletion message shows up as a validation message on import screen       |   success message for importing content guide    |  deletion message for deleting a content guide    |
|    Screenshot 2025-10-31 at 17 48 11    |  Screenshot 2025-10-31 at 17 48 41     |     Screenshot 2025-10-31 at 17 48 47  |

5. Don't hide subsections in groups if they have instructions

Previously, we would hide subsections on the composer section view page if the subsection met 2 conditions:
- same name as grouping
- no body text

However, now that instructions are a thing, we should _also_ check if they have instructions. If they do, we should keep them, even if they meet the other 2 conditions.

| before | after |
|--------|-------|
|   "help with systems" subsection was being dropped     |     "help with systems" subsection is not visible because it has instructions |
|   Screenshot 2025-10-31 at 17 50 25     |    Screenshot 2025-10-31 at 17 49 59   |

### DIFF
--- a/nofos/composer/views.py
+++ b/nofos/composer/views.py
@@ -256,7 +256,7 @@ class ComposerSectionView(GroupAccessObjectMixin, DetailView):
 
             # if first item
             if len(subsection_groups[current_idx]["items"]) == 0:
-                # skip if subsection heading matches group name and subsection.body is empty
+                # skip if subsection heading matches group name, subsection.body and subsection.instructions are empty
                 if (
                     normalize_name(subsection.name)
                     == normalize_name(subsection_groups[current_idx]["heading"])
--- a/nofos/composer/tests/test_views.py
+++ b/nofos/composer/tests/test_views.py
@@ -525,7 +525,7 @@ class GroupSubsectionsTests(TestCase):
         self.assertEqual([x.pk for x in groups[0]["items"]], [s1.pk, s2.pk])
 
     def test_header_item_with_nonempty_instructions_is_included(self):
-        # s1 is a header (preset name) with real body → should appear in items
+        # s1 is a header (preset name) with non-empty instructions → should appear in items
         s1 = self._make_subsection(
             "Funding details",
             tag="h4",


nofos/composer/views.py
```diff
@@ -253,6 +261,7 @@ def normalize_name(s: str) -> str:
                     normalize_name(subsection.name)
                     == normalize_name(subsection_groups[current_idx]["heading"])
                     and not subsection.body
+                    and not subsection.instructions
```
jlhogantech: **nit**: Update comment on line 259
```
                # skip if subsection heading matches group name, subsection.body is empty, and subsections does not have instructions
```

pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "nofos"
-version = "3.17.0"
+version = "3.18.0"
```
jlhogantech: for my own learning, version update required because of DB migration?

jlhogantech (approved):

Issue #500, opened 2025-11-03
Add and delete subsections (and build the UI for it)


pcraig3:
In the NOFO Builder, we let users create and delete subsections (we also technically allow reordering but only for superusers). 

For the NOFO Composer, we want to let SysAdmins create and delete subsections also.

To match the NOFO Builder functionality, we want to create 2 different UIs for this:

1. Add an "other actions" box at the bottom of the "edit" screen
2. Add a "configure section" link and page 

### 1. Other actions box on "edit" page

Pretty simple, let's add a box with 2 buttons, same as we have in the NOFO Builder. 

| "Other actions" box in NOFO Builder | "Other actions" box in our Figma designs |
|--------|-------|
|  2 buttons: add and delete      |  2 buttons also, add and delete     |
|     <img width="1002" height="718" alt="Image" src="https://github.com/user-attachments/assets/8f052b06-8e91-422d-b373-c7ad57452297" />   |   <img width="782" height="733" alt="Image" src="https://github.com/user-attachments/assets/684b3fa4-63ca-48fc-a67c-35146bf6996e" />    |

### 2. "Configure section" link and page 

For the NOFO Builder, we have this "configure section" link which brings you to a page that has "add" and "delete" buttons. It gives you a high-level view of your section to be able to add and delete sections. We want to do something similar for the Composer. 

We need to build a page for this, but we also need to add a link.

| "Configure section" link in NOFO Builder | "Configure section" link in NOFO Composer designs  |
|--------|-------|
|   The link in the NOFO Builder is in the section header     |   Right now, the link is at the bottom of the section. This seems like not the best place for it, but we can move the link around later if we need.    |
|    <img width="1002" height="548" alt="Image" src="https://github.com/user-attachments/assets/8c6bb9ca-a40e-4073-b458-372b7d86ed2b" />    |    <img width="751" height="402" alt="Image" src="https://github.com/user-attachments/assets/69428630-0c91-40b8-ad3f-2323efaeecaf" />   |

| "Configure section" page in NOFO Builder | "Configure section" page in NOFO Composer designs  |
|--------|-------|
|    Same layout with "Add subsection" button on each `hr` and a delete button next to it   |  It's a variant of our preview page, with an `hr` between sections and a `delete` link on the right column. Maybe a table based layout is the best here     |
[...]

PR #501, opened 2025-11-06
500 add remove subsections

jlhogantech:
This is part one of the changes required for https://github.com/HHS/simpler-grants-pdf-builder/issues/500

This PR:
- Adds buttons to the bottom of content guide subsection edit page to "add" and "delete" subsection
- Adds Content guide subsection Create view, with ability to create a new content guide subsection with all the same features as nofo build add subsection, PLUS: edit mode and instructions
- Adds content guide subsection delete confirmation view

TODO: 
- Create content guide section edit ("configure section" view) with Add and delete subsection functionality

Screenshots:
| Buttons on ContentGuide subsection edit page | Create ContentGuide page | With instructions expanded | After create with new subsection | Delete confirmation page | after delete without subsection|
| --- | --- | --- | --- | --- | --- | 
| Screenshot 2025-11-05 185111 | Screenshot 2025-11-05 184607 | Screenshot 2025-11-05 184621 | Screenshot 2025-11-05 184929 | Screenshot 2025-11-05 184951 | Screenshot 2025-11-05 185003|
|   |   |   NOTE: black outline is just browser interaction, not actual styles | NOTE: green success message does display but I didn't act fast enough to screenshot |  | NOTE: red deleted message does display but I didn't act fast enough to screenshot |

### DIFF
--- a/nofos/composer/views.py
+++ b/nofos/composer/views.py
@@ -441,7 +441,8 @@ class ComposerSubsectionCreateView(GroupAccessObjectMixin, CreateView):
         url = reverse_lazy(
             "composer:section_view", args=[self.document.pk, self.section.pk]
         )
-        return "{}#{}".format(url, self.object.html_id)
+        anchor = getattr(self.object, "html_id", "")
+        return "{}?anchor={}#{}".format(url, anchor, anchor) if anchor else url
 
 
 class ComposerSubsectionEditView(GroupAccessObjectMixin, UpdateView):
--- a/nofos/composer/templates/composer/subsection_edit.html
+++ b/nofos/composer/templates/composer/subsection_edit.html
@@ -134,13 +134,13 @@
         <ul class="usa-list usa-list--unstyled">
           <li>
             <a class="usa-button usa-button--green" href="{% url 'composer:subsection_create' document.pk section.pk %}?prev_subsection={{subsection.id}}">
-              Add subsection
-            </a> — add a new subsection after this one
+              Add section
+            </a> — add a new section after this one
           </li>
           <li>
             <a class="usa-button usa-button--secondary" href="{%url 'composer:subsection_confirm_delete' document.pk section.pk subsection.pk %}">
-              Delete subsection
-            </a> — remove this subsection completely
+              Delete section
+            </a> — remove this section completely
           </li>
         </ul>
       </div>


nofos/composer/views.py
```diff
@@ -404,6 +495,49 @@ def get_success_url(self):
         return "{}?anchor={}#{}".format(url, anchor, anchor) if anchor else url
 
 
+class ComposerSubsectionDeleteView(GroupAccessObjectMixin, DeleteView):
+    model = ContentGuideSubsection
+    pk_url_kwarg = "subsection_pk"
```
jlhogantech: `pk_url_kwarg` i just coped from NofoSubsectionDeleteView, not sure if needed

pcraig3: Okay, so I had to look this up, but basically, the answer is `get_object`: https://docs.djangoproject.com/en/5.2/ref/class-based-views/mixins-single-object/#django.views.generic.detail.SingleObjectMixin.get_object

Django uses `get_object` to grab the instance being referred to, but it assumes that the object will be passed in via a param called `pk` — and if you don't have that param, you can reassign this `pk_url_kwarg` variable to be the name of the parameter for the id of the object we are looking for. 

If you comment this line out, you get a 404: object not found.

Since we are driving down the hierarchy, we are doing this:

`<uuid:pk>/section/<uuid:section_pk>/subsection/<uuid:subsection_pk>/delete`

We _could_ flip this to something like this:

`<uuid:document_pk>/section/<uuid:section_pk>/subsection/<uuid:pk>/delete`

But it breaks the model we have for all our other urls. So in the end, we change the variable name and it works. Bing bang boom.

nofos/composer/templates/composer/subsection_create.html
```diff
@@ -0,0 +1,245 @@
...
+            <span class="usa-checkbox__label-description">Callout boxes <span aria-hidden="true">(📦)</span> use an accent color to call attention to important content.</span>
+          </label>
+        </div>
+      </div>
+
+      <!-- Add instructions -->
+      <details class="form-group margin-top-4">
+        <summary><span class="text-bold">Add instructions ?</span></summary>
+        <div class="form-group margin-top-4">
+          <label
```
pcraig3: Let's throw in some `usa-prose` here that copies the text we have on the "Edit instructions" page:

> - Edit instructions that OpsDivs and writers will review.
> - This content will automatically be locked and writers will not be able to make changes.

pcraig3:
Oh, and finally: we traditionally haven't been testing views.py for the `nofos` app, but we have started introducing view level tests for the `compare` and `composer` apps.

So let's add some basic tests for the new composer views (similar to the other ones, just make sure they load and can do what we want them to).

pcraig3 (approved):
Added two small comments but pre-approving this PR because they are very small and everything else here looks good.

Thanks @jlhogantech! Keen to get this in, it's shaping up!

PR #502, opened 2025-11-07
Add syntax highlighting for {variables} surrounded by curly braces

pcraig3:
## Summary

This PR adds a new syntax highlighting logic to the Ace editor (bundled with martor) to highlight text wrapped with curly braces (eg `{variable}`) as a variable. 

It also adds a markdown extension that adds a `<span>` around markdown formatting strings that include curly-brace surrounded variables (eg `{variable}`), so that we can apply the yellow highlight that writers are currently used to seeing in the Word doc.

It also tracks currently identified variables in the radio button for a "Select certain text only", which updates as the user types (see gif below).

WONTDO: This PR _will not_ add highlighting of variables to the section page overview or preview page, as these are handled in the #486 issue.

### Screenshots 

**Ace editor in "edit" mode**

| before | after |
|--------|-------|
|   variables not highlighted (default behaviour)     |   variables have a $\color{magenta}{\textsf{pink}}$ highlight colour     |
|    Screenshot 2025-11-07 at 17 19 27    |     Screenshot 2025-11-07 at 17 19 55  |

**Ace editor in "preview" mode**

| before | after |
|--------|-------|
|  variables not highlighted (default behaviour)       |    variables have a yellow background colour, like they do in the Word doc   |
|   Screenshot 2025-11-07 at 17 22 22     |    Screenshot 2025-11-07 at 17 20 28   |

**Tracking variables in radio button label (gif)**

Screen Recording 2025-11-10 at 15 30 51

### To test this

- View "edit_subsection" page in the composer
- Try to select "Edit certain text only" without any variables: you should see a validation message
- Add variables surrounded by curly braces (`This is a variable: {variable}`): they should show up highlighted in pink
  - Currently active variables should show up in the radio button description for "Edit certain text"
- Click into the "preview" tab: you should see the variables highlighted in yellow
- Click "Save"
- Use Devtools inspector to highlight variables on the section_overview screen with the accordion box open: you should see them unstyled but wrapped in a span, like this: `<span class="md-curly-variable">{variable}</span>`

### Details 

Mechanically, there are 2 new pieces of configuration logic we needed to enable this:

1. A Javascript file that adds [syntax highlighting](https://ace.c9.io/#nav=higlighter) to Ace editor: `nofos/bloom_nofos/static/js/composer/martor-curly-variables.js`
2. A [custom markdown extension](https://django-markdown-editor.readthedocs.io/en/latest/faq.html#how-do-i-add-custom-markdown-extensions) that will identify variables when we output markdown text as HTML: `nofos/bloom_nofos/markdown_extensions/curly_variables.py`

#### JavaScript syntax highlighter

This part of the code was the hardest because Ace is compiled and so we can't exactly follow the instructions on their documentation page. Luckily for us, Ace is old enough that it exposes a global object on the `window` which means we can get in there and patch in a new syntax highlighting rule, which means that all existing formatting works in addition to our new formatting. It's kind of weird how it works, but it does work and it's reasonably clean. Lots of comments in the JS file.

This turns the `{variable}` strings $\color{magenta}{\textsf{pink}}$ in the editor itself. 

#### Python markdown extension

Unlike the Ace stuff, martor does give an example on their documentation of adding in a custom markdown extension: https://django-markdown-editor.readthedocs.io/en/latest/faq.html#how-do-i-add-custom-markdown-extensions

This becomes a global extension that is applied app-wide, which means it will affect all markdown output (meaning in the Composer, the Compare tool, and the regular NOFO Builder). This should be fine, since curly braces aren't common in NOFOs. However, it does mean that we should be careful to only apply 'yellow highlight' styling in the composer (we can use body class names for this) so that we don't start highlighting things in published NOFOs by accident. 

#### Regex strings

We identify variables by doing regex comparisons. We do this in various places in this PR, which means that it would be nice to be able to use the same exact comparison string. However, because some comparisons are server-side (python) and some are client-side (JS), we have to have 2 different Regex patterns. I can't think of a great way around this, I figure it's fine to just have 2 of them.

Python: `CURLY_VARIABLE_PATTERN = r"(?<!\\)\{([^{}]*\S[^{}]*)\}"`
JS: `const CURLY_VARIABLE_PATTERN_JS = /(?<!\\)\{[^{}]*\S[^{}]*\}/g;`

We don't have any JS tests in the repo, but we have lots of Python tests, so the Python regex is tested both for markdown output _and_ in the `"extract_variables"` string in the model. 

There's a pretty good list of test cases in there and the tests make sure that what markdown highlights is the same as what we understand the "variables" to be at the model level. Here is the list of strings that we test:

```
"This sentence has no variables."
"This sentence has one {variable}."
"This has { variable with spaces }."
"Enter {Project name} and confirm {Project name}."
"This has {first} and {second} variables."
r"This is not a \{variable\}."
r"This is not a \{variable1} but this is a {variable2}."
"This {outer {inner} text} has nesting."
"This has {} empty braces."
"This has {   } empty braces with whitespace."
"This has { unmatched brace."
"Choose from {List: available options}."
"**Bold** and {variable} and _italic_."
```

This is still an MVP so we don't have live users and in particular, we don't have users who are realistically trying to game this to get it to break, as they would just be screwing things up for their colleague. 

All to say, we are not trying to cover literally every edge case, but this gives us a foundation to build on in case we notice weird behaviour in the future.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,14 @@ Versioning since version 1.0.0.
 
 ### Added
 
+- Add functionality for adding/deleting a Content Guide subsection
+- Add "preview" page for Composer
+- Added syntax highlighting to Ace editor for {variables}
+- Add django-markdown-editor markdown extension for wrapping {variables} with spans
+
 ### Changed
+
+- Allow "instructions" to be edited like a subsection body
 
 ### Fixed
 
@@ -30,7 +37,7 @@ Versioning since version 1.0.0.
 
 ### Fixed
 
-- Show "destructive" alert message when deleting a NOFO
+- Show "destructive" alert message when deleting a Content Guide
 
 ### Migrations
 
--- a/nofos/bloom_nofos/static/js/composer/martor-curly-variables.js
+++ b/nofos/bloom_nofos/static/js/composer/martor-curly-variables.js
@@ -92,7 +92,7 @@
           value: token.value.substring(lastIndex, m.index),
         });
       }
-      // Add the curly brace itself as a special "variable.curly" token
+      // Add the variable with included curly braces as a special "variable.curly" token
       // This is what CSS will target with .ace_variable.ace_curly
       newTokens.push({ type: "variable.curly", value: m.text });
       lastIndex = m.index + m.text.length;
--- a/nofos/composer/models.py
+++ b/nofos/composer/models.py
@@ -134,13 +134,13 @@ class ContentGuideSubsection(BaseSubsection):
         used_keys = set()
 
         for m in self._VAR_PATTERN.finditer(text):
-            raw = m.group(1).strip()  # group(1) instead of group("raw")
+            variable = m.group(1).strip()
 
             var_type = "string"
-            label = raw
+            label = variable
 
-            if ":" in raw:
-                head, tail = raw.split(":", 1)
+            if ":" in variable:
+                head, tail = variable.split(":", 1)
                 if head.strip().lower() == "list":
                     var_type = "list"
                     label = tail.strip()


nofos/bloom_nofos/markdown_extensions/curly_variables.py
```diff
@@ -0,0 +1,79 @@
...
+Matches:
+    {variable}
+    { variable with spaces }
+    {List: items}
+    {Prompt text here}
+
+Ignores:
+    \{escaped\}         ← escaped braces
+    {   }               ← only whitespace inside
+    {nested {braces} ignored}   ← nested braces are not supported (variable is "{nested {braces}" )
```
jlhogantech: **nit**: not totally clear on what this comment means -- nested braces are not supported and so are ignored? Or are not supported so the inner braces become part of the variable name e.g. the variable is `"nested {braces}"`? I think the second but the first outer `{` in the example given in the comment makes me unsure.

nofos/bloom_nofos/markdown_extensions/curly_variables.py
```diff
@@ -0,0 +1,79 @@
...
+        el.text = m.group(0)
+        # Return element plus start/end positions for Markdown processing
+        return el, m.start(0), m.end(0)
+
+
+class CurlyVarExtension(Extension):
+    """
+    Markdown extension that registers the CurlyVarInlineProcessor.
+
+    The processor runs fairly early (priority 185) so that variable
```
jlhogantech: **nit**: what are the relative priorities? Is 1 the highest or the lowest?

pcraig3: Higher priorities go first! (Here is a passive-aggressive comment about it: https://github.com/Python-Markdown/markdown/issues/755)

The example extension in the markdown repo is at priority `175`: https://python-markdown.github.io/extensions/api/#extendmarkdown

The priorities of some other things are listed in the source code: https://github.com/Python-Markdown/markdown/blob/master/markdown/inlinepatterns.py#L74

At 185, it runs after code spans (190), so if you do ```{variable}```, it shouldn't work (this would probably never be in a NOFO though). 

I don't have a strong reason for this, just that it's close to the example and runs with a reasonably high priority. We might find there is some edge case and would have to revisit this.

pcraig3:
> Would love to hear how you came to this solution of wrapping the martor tokenizer

Ace editor has some documentation on creating a syntax highlighter: https://ace.c9.io/#nav=higlighter

It shows how you would add rules: 

```
var MyNewHighlightRules = function() {
   this.$rules = {
        "start" : [
            {
                token: token, // String, Array, or Function: the CSS token to apply
                regex: regex, // String or RegExp: the regexp to match
                next:  next   // [Optional] String: next state to enter
            }
        ]
    };
};
```

So that's great, but once we access the ACE object from the window, it is too late to add rules this way. The tokenizer is already instantiated with the old rules and doesn't like the new one.

However, we can change tokens AFTER they're created, which means we let the original rules run and then we intercept them and post-process them to create a new highlighting grouping, and then tell Ace to re-render and it works as we would expect. So, not exactly perfect but perfectly sufficient.

PR #506, opened 2025-11-11
503 composer basic audit history

### LINKED ISSUE #503: Add (basic) audit history page for NOFO Composer
In the NOFO Builder, we have an audit history page which shows a list of edits to the NOFO object. 

| before | after |
|--------|-------|
|    click this link to see list of updates    |   page of updates!    |
|   Image     |    Image   |

We are using a library called '[django-easy-audit](https://pypi.org/project/django-easy-audit/)' which automatically logs all updates to all known objects. This means it makes an entry when we add a user, edit a NOFO, delete a subsection, etc. It's pretty messy, but it is nice because it works without any real configuration or having to integrate custom code into our first-party apps. 

The audit page is built by querying django easy audit for events related to a NOFO, Section, or Subsection. The logic for all that stuff is here: https://github.com/HHS/simpler-grants-pdf-builder/blob/main/nofos/nofos/audits.py

It is a pretty inefficient method because we are doing these unbounded queries across different models, which is simple but not particularly smart. This page can take 10-15 seconds to load in prod, but that's not something we need to fix for this issue, we can just eat it for now. (Note: if you want to fix this, be my guest — but let's chat about it first.)

We also have a "updated_by" field on the NOFO object which saves the last user to make a change, so at least that one is instant: https://github.com/HHS/simpler-grants-pdf-builder/pull/474

For this story, we want to add the "See all updates" link on the section overview page: https://www.figma.com/design/J41kYT5DvfnxYgqzaB6FKP/NOFO-Composer?node-id=754-15307&t=F9dp2JqHVS4oynGb-0

And we want to add a new page for the audit history: https://www.figma.com/design/J41kYT5DvfnxYgqzaB6FKP/NOFO-Composer?node-id=1139-7296&t=F9dp2JqHVS4oynGb-0

The only difference between what we do with the NOFO Builder and the Composer is that we want to call them "Steps" instead of "Sections" and "Sections" instead of "Subsections".

I have a strong suspicion that people will want to see a before/after of the changes that are made to subsections when other people make changes, but for this story we are _not_ going to implement that page: only a generic "history" page, same as in the NOFO Builder.

jlhogantech:
Adds a basic audit history for composer view by 
- generalizing the existing audit helpers to take model name params
- moving core audit event loading and pagination into a base view
- inheriting from the base view for preexisting NOFO history 
- inheriting from the base view for new composer history

Screenshot 2025-11-11 180609

closes #503

### DIFF
--- a/nofos/nofos/audits.py
+++ b/nofos/nofos/audits.py
@@ -32,13 +32,14 @@ def deduplicate_audit_events_by_day_and_object(events):
     return list(deduplicated.values())
 
 
-def format_audit_event(event, formatting_options={}):
+def format_audit_event(event, formatting_options=None):
     """
     Takes a CRUDEvent and returns a formatted dictionary for display in the UI.
     Includes enhanced labels and object details.
     """
     BASE_DOCUMENT_TYPES = ["nofo", "contentguide"]
 
+    formatting_options = formatting_options or {}
     SubsectionModel = formatting_options.get("SubsectionModel", Subsection)
     document_display_prefix = formatting_options.get("document_display_prefix", "NOFO")
 
@@ -60,7 +61,7 @@ def format_audit_event(event, formatting_options={}):
         event_details["object_html_id"] = ""
 
     # Improve object description for subsection edits
-    if "subsection" in event.content_type.model:
+    if event.content_type.model.endswith("subsection"):
         try:
             subsection = SubsectionModel.objects.get(id=event.object_id)
             name = subsection.name or "#{}".format(subsection.order)
--- a/nofos/nofos/views.py
+++ b/nofos/nofos/views.py
@@ -1093,20 +1093,19 @@ class NofoEditStatusView(BaseNofoEditView):
 
 class BaseNofoHistoryView(DetailView):
     """
-    Mixin to provide common functionality for NOFO history views.
+    Mixin to provide common functionality for BaseNOFO-object history views.
     """
 
     events_per_page = 25  # Show 25 events per batch
 
     # Must be set by child classes
     model = None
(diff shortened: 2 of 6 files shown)


nofos/nofos/audits.py
```diff
@@ -32,11 +32,16 @@ def deduplicate_audit_events_by_day_and_object(events):
     return list(deduplicated.values())
 
 
-def format_audit_event(event):
+def format_audit_event(event, formatting_options={}):
```
jlhogantech: don't love where this landed with passing in the `formatting_options`, but those need to be defined by / scoped to the calling view. Another option could be to move this `format_audit_event` helper util into the `BaseHistoryView`, but it is also needed by the nofo-specific `NofoModificationsHIstoryView`. So just making view-specific dict of props felt like the easiest way forward.

nofos/nofos/views.py
```diff
@@ -1090,32 +1091,82 @@ class NofoEditStatusView(BaseNofoEditView):
...
+    Mixin to provide common functionality for NOFO history views.
+    """
+
     events_per_page = 25  # Show 25 events per batch
 
-    archived_error_message = "Archived NOFOs don’t have an audit history."
-    cancelled_error_message = None  # setting None allows us to view for cancelled NOFOs
+    # Must be set by child classes
+    model = None
+    event_formatting_options = {}
```
pcraig3: Can we do this?

```
    def get_event_formatting_options(self):
        return {}
```

pcraig3 (approved):
Awesome, thanks for this! Going to merge so it is ready for later today.

PR #514, opened 2025-11-14
Make {variables} regex stricter to fix custom attributes

pcraig3:
## Summary

This PR fixes markdown [attribute lists](https://python-markdown.github.io/extensions/attr_list/) my making our "variable" regex more strict. 

Specifically, we no longer consider something a variable if it starts with: `{:`, `{.`, or `{#`.

### Screenshots

With markdown attribute lists, we could add classnames to table headings like this: `{: .w-33}`, but our new extension messed this up.

| present (bug) | after (fixed) |
|--------|-------|
|  classname string is visible, not actually being applied (tables are varying widths)      |  classname string is gone and applied correctly (table columns are equal width)     |
|   Screenshot 2025-11-14 at 18 45 36     |    Screenshot 2025-11-14 at 18 44 46   |

### Details

In a recent previous PR (#502), we added a new markdown extension that looks for variables, which are wrapped in curly braces.

However, we also sometimes use our "[markdown attribute lists](https://python-markdown.github.io/extensions/attr_list/)" extension, which allows us to add classes, ids, etc into markdown with a custom syntax that _also_ uses curly braces.

So, for example, here is some of that syntax:

- `{: #someid .someclass somekey='some value' }`
- `{: #id1 .class1 id=id2 class="class2 class3" .class4 }`

Our new extension was messing up parsing custom attributes, so I made the regex stricter.

Now, we _no longer_ consider something a custom variable if it starts with either of these strings:

- `{:`
- `{.`
- `{#`

So that should restore functionality of custom attributes, although it _requires_ that we add the colon, period, or hash, which is not strictly required in the custom attribute documentation.

### DIFF
--- a/nofos/bloom_nofos/markdown_extensions/curly_variables.py
+++ b/nofos/bloom_nofos/markdown_extensions/curly_variables.py
@@ -11,15 +11,29 @@ in a <span> element so they can be styled in the rendered HTML.
 
 ----------------------------------------------------------------
 Regex pattern used:
-    (?<!\\)\{([^{}]*\S[^{}]*)\}
+    (?<!\\)\{(?![:.#])([^{}]*\S[^{}]*)\}
 
 Explanation:
-    (?<!\\)     Negative lookbehind — ensure the "{" is not escaped (e.g., "\{" is ignored)
-    \{          Match literal opening brace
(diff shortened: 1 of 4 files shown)


pcraig3:
Merging because this is a bugfix PR.

PR #515, opened 2025-11-16
507: Update appearance of edit mode labels

jlhogantech:
updates edit_mode labels to:
- not exist for "full"
- have specified colors
- have specified text

NOTE: I just used `bg-primary-lighter` for the blue, because as far as I can tell this specific blue does not yet exist anywhere in the app yet. Can add if we need but felt like this might not be the right place to introduce a totally new color? 

Screenshot 2025-11-16 143736


nofos/bloom_nofos/templates/includes/subsection_label.html
```diff
@@ -0,0 +1,9 @@
+{% load edit_mode_label edit_mode_label_class %}
+
+{% with edit_mode|edit_mode_label as label %}
+  {% with edit_mode|edit_mode_label_class as class %}
+    {% if label %}
+      <span class="usa-tag maxh-fit-content text-black {{class}}">{{ label }}</span>
+    {% endif %}
```
pcraig3: Let's also make `font-weight: 400` and remove the `text-transform: none` to match the mocks a bit better.

PR #516, opened 2025-11-16
508-scroll-to-top macro

jlhogantech:
Adds an optional `scroll to top` button to the `base.html` template, which can be "turned on" by adding "include_scroll_to_top" = True in the view context data. 

also as ride-alongs:
- adds a little missing padding to the top of the 'configure section' page
- adds subsection body styling to the 'configure section' page -- is this right? 

scroll_to_top

### DIFF
--- a/nofos/bloom_nofos/static/styles.css
+++ b/nofos/bloom_nofos/static/styles.css
@@ -155,7 +155,6 @@ footer .usa-footer__primary-link:visited {
   background-position: center center;
   background-size: cover;
   display: inline-flex;
-
 }
 
 .usa-button--icon--only {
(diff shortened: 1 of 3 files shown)


pcraig3 (changes requested):
This is great!

Let's do a couple changes before merging:

1. Put the JS into a `<script>` tag in the macro itself
2. Remove the current “back to top” link from the NOFO edit page (`nofos/templates/nofos/nofo_edit.html`) and use this macro instead. 

In the `nofo_edit` template, you would (a) remove the link at the bottom of the page, (b) remove the “sentinel” div, and (c) go into the "nofo_edit.js" file (under "bloom_nofos/static/js") and [delete the portion of the JS](https://github.com/HHS/simpler-grants-pdf-builder/blob/main/nofos/bloom_nofos/static/js/nofos/nofo_edit.js#L187) which targets the back to top link.

PR #517, opened 2025-11-17
Yes/No questions for System Admin user

pcraig3:
## Summary

This PR fulfils the requirement to show “Conditional: Yes|No” labels on the accordion headers and also adds a Note that says “Hey, this is the “NO” answer to {{ question }}”. 

It is a big PR because it introduces a new data format for the JSON questions and also a little python loader for pulling them into Python code. There’s a fulsome discussion of this PR in this issue: #518 

#### Section accordions 

| before | after |
|--------|-------|
|    only says "locked, etc", no conditional label    |  adds "Conditional: Yes" and "Conditional: No" labels to accordions     |
|    Screenshot 2025-11-18 at 09 25 20    |   Screenshot 2025-11-18 at 09 22 46    |

#### Cost sharing: NO 

| before | after |
|--------|-------|
|  no label, no note      |  label: "Conditional: No" + note above instructions     |
|  Screenshot 2025-11-18 at 09 25 02    |    Screenshot 2025-11-18 at 09 23 07   |

#### Cost sharing: YES 

| before | after |
|--------|-------|
|  no label, no note      |  label: "Conditional: Yes" + note above instructions     |
|  Screenshot 2025-11-18 at 09 24 27   |  Screenshot 2025-11-18 at 09 23 36   |

### Details

Content Guides include several policy-driven conditional questions (e.g., cost sharing, maintenance of effort).

Up until now, these conditional relationships were not surfaced in the Composer UI, and subsections did not indicate when they were tied to a conditional question. Now, they are.

This PR introduces a JSON-formatted conditional question registry and UI updates for the System Admin flow that display conditional metadata directly within Composer.

New stuff:

1. JSON-driven conditional question definitions

- New JSON file: `conditional_questions.json`, containing question keys, labels, and subsection name mappings.
- New Python file: `conditional_questions.py` to dynamically load and match these definitions.

2. Template filters + UI elements

- New template tags:
  - `conditional_questions_label` — returns badge text for subsections tied to a conditional question.
  - `conditional_questions_note` — returns the full conditional question prompt for display.
- New UI:
  - `subsection_label.html` — renders a conditional badge beside edit mode tags.
  - `composer_section.html` — displays explanatory conditional notes below subsection headings.

3. Subsection-level YES/NO parsing

- Added model logic to detect (YES) or (NO) markers within a subsection’s instructions.
- Exposes boolean helpers (`conditional_answer`, `is_conditional`) to support conditional workflows.

Subsections connected to conditional questions are now clearly labeled within the Composer UI. Writers see inline guidance explaining why a subsection appears and how it relates to a policy question. No database schema changes required; all behaviour is driven by configuration and template logic.

### How to test

It's easiest if you have an updated Content Guide (`CDC_Content_Guide_2025-11-18`).

1. Import content guide
2. Click into it
3. Scroll down to "Eligibility -> Cost sharing"
4. Tags should be visible. You can open and close these sections to see the note.

### DIFF
--- a/nofos/composer/models.py
+++ b/nofos/composer/models.py
@@ -121,18 +121,18 @@ class ContentGuideSubsection(BaseSubsection):
 
     def _find_yes_no_string(self) -> str | None:
         """
-        Parse the instructions for a (YES) or (NO) marker.
+        Find the exact strings "(YES)" or "(NO)" inside the instructions.
 
         Returns "YES", "NO", or None if nothing found.
         """
         if not self.instructions:
             return None
 
-        m = self._YES_NO_PATTERN.search(self.instructions)
-        if not m:
+        token = self._YES_NO_PATTERN.search(self.instructions)
+        if not token:
             return None
 
-        return m.group(1).upper()
+        return token.group(1).upper()
 
     @property
     def conditional_answer(self) -> bool | None:
@@ -140,9 +140,9 @@ class ContentGuideSubsection(BaseSubsection):
         Boolean representation of the (YES)/(NO) token in instructions.
 
         Returns:
-            True  -> (YES)
-            False -> (NO)
-            None  -> no token present
+            (YES) -> True
+            (NO) -> False
+            No token -> None
         """
         token = self._find_yes_no_string()
         if token == "YES":
--- a/nofos/composer/utils.py
+++ b/nofos/composer/utils.py
@@ -48,13 +48,12 @@ def get_edit_mode_label_class(value: str) -> str:
     return EDIT_MODE_LABEL_CLASSES.get(value) or ""
 
 
-def get_conditional_questions_label(subsection: ContentGuideSubsection) -> str:
+def get_conditional_questions_label(conditional_answer=None) -> str:
     """Return a label for an ContentGuideSubsection that is a conditional question."""
-    if subsection and subsection.is_conditional:
-        if subsection.conditional_answer:
-            return "Conditional: Yes"
-        else:
-            return "Conditional: No"
+    if conditional_answer is True:
+        return "Conditional: Yes"
+    if conditional_answer is False:
+        return "Conditional: No"
 
     return ""
 
--- a/nofos/bloom_nofos/templates/includes/subsection_labels.html
+++ b/nofos/bloom_nofos/templates/includes/subsection_labels.html
@@ -1,5 +1,13 @@
 {% load edit_mode_label conditional_questions_label %}
 
+{% if conditional_answer is True or conditional_answer is False %}
+  {% with conditional_answer|conditional_questions_label as conditional_label %}
+      {% if conditional_label %}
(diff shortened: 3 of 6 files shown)


nofos/bloom_nofos/templates/includes/subsection_label.html
```diff
@@ -1,9 +1,19 @@
...
+    {% with edit_mode|edit_mode_label_class as class %}
+      {% if label %}
+        <span class="usa-tag maxh-fit-content text-black text-normal text-no-uppercase {{class}}">{{ label }}</span>
+      {% endif %}
+    {% endwith %}
   {% endwith %}
-{% endwith %}
+{% endif %}
+
+{% if subsection %}
```
jlhogantech: given how this is implemented, could you pass in both `edit_mode` and `subsection` to a single subsection label inclusion macro? If that's not the intended use (as in composer_section.html where you have two separate tag templates), then should this maybe be an else if to make that explicit?

nofos/composer/models.py
```diff
@@ -115,6 +115,46 @@ class Meta:
...
+    def _find_yes_no_string(self) -> str | None:
+        """
+        Parse the instructions for a (YES) or (NO) marker.
+
+        Returns "YES", "NO", or None if nothing found.
+        """
+        if not self.instructions:
+            return None
+
+        m = self._YES_NO_PATTERN.search(self.instructions)
```
jlhogantech: **nit**: prefer to avoid single-char variables

nofos/composer/models.py
```diff
@@ -115,6 +115,46 @@ class Meta:
...
+
+        return m.group(1).upper()
+
+    @property
+    def conditional_answer(self) -> bool | None:
+        """
+        Boolean representation of the (YES)/(NO) token in instructions.
+
+        Returns:
+            True  -> (YES)
```
jlhogantech: **nit**: this reads as backwards to me, my brain expects `(YES) -> True` etc

nofos/composer/models.py
```diff
@@ -115,6 +115,46 @@ class Meta:
         help_text="Whether to show this section in the content guide.",
     )
 
+    # ---------- Conditional answer helpers ---------- #
+
+    _YES_NO_PATTERN = re.compile(r"\((YES|NO)\)")
+
+    def _find_yes_no_string(self) -> str | None:
```
jlhogantech: **nit**: Include in function name that matching is not just for "yes" or "no" string, but rather the `()`-wrapped marker or token. You use `marker` here, and `token` below in docstring for `conditional_answer`, i would stick with one and I think either is fine

nofos/composer/utils.py
```diff
@@ -48,6 +48,36 @@ def get_edit_mode_label_class(value: str) -> str:
     return EDIT_MODE_LABEL_CLASSES.get(value) or ""
 
 
+def get_conditional_questions_label(subsection: ContentGuideSubsection) -> str:
```
jlhogantech: **nit**: this could be achieved with only `conditional_answer`, could just pass in that single necessary field to the label template macro for consistency with the `edit_mode` functionality. but certainly not necessary at all

PR #524, opened 2025-11-20
511: Remove yes/no edit_mode, and add separate optional flag

### LINKED ISSUE #511: Add "optional" attribute to ContentGuideSubsection
After doing a deep dive into an Actually Existing Content Guide, I have learned that some sections are optional _and_ editable, which is a combination we don't currently we support. 

| optional AND editable section |
|--------|
|   Section can be removed if not relevant. However, if it is relevant and you are keeping it, you have to be able to change the text.     |
|   Image     |

Instead of trying to add more `edit_mode` combinations, we should pull out the “optional/required” decision into its own question. 

When we start on the writer’s flow, we will need a way for writers to indicate their answer to this question, but for the time being, we just need to mark whether this is an optional question. 

I think “optional” is the more descriptive here, so I would say we should add a boolean value for “optional” to `models.py` and then add it to the subsection edit form, which means the “edit” page and the “new” page.  `optional` is false by default.

We should _also_ remove the `yes_no` edit type, and the label for it.

Here is a mockup of what we are looking for ([link](https://www.figma.com/design/J41kYT5DvfnxYgqzaB6FKP/NOFO-Composer?node-id=754-15222&t=e8BVK6fPzZBjNAYr-0)):

| before  | after |
|--------|-------|
|   4 "edit mode" options for a subsection    |  3 "edit mode" options and a new question about whether this subsection is required or optional      |
|     Image  |     Image  |

This new design would apply on the Subsection edit page and also the create new Subsection page.

If a subsection is “optional”, that means we will give it a tag on the accordion header like we do with other statuses. But because it is now independent of edit_mode, it means we can have 2 tags on a subsection (eg, "optional" and "locked" at the same time), which then means that we need to probably create a little flex container div (or inline block) so that we can list tags horizontally for when there are more than one on a section. 

So, to sum up:

- Add `optional` to the subsection model: defaults to false 
- Update edit page and new subsection page with new radio button label
- Add an optional label to the section overview page (ie, with the accordion headers)

jlhogantech:
closes #511 

| Subsection edit page |
| -- |
|Screenshot 2025-11-19 195052 | 
| Section view |
|Screenshot 2025-11-19 195109 |

### DIFF
--- a/nofos/composer/forms.py
+++ b/nofos/composer/forms.py
@@ -16,7 +16,15 @@ class ComposerSubsectionCreateForm(forms.ModelForm):
 
     class Meta:
         model = ContentGuideSubsection
-        fields = ["name", "tag", "callout_box", "edit_mode", "body", "instructions"]
+        fields = [
+            "name",
+            "tag",
+            "callout_box",
+            "optional",
+            "edit_mode",
+            "body",
+            "instructions",
+        ]
         widgets = {
             "name": forms.TextInput(),
         }
--- a/nofos/composer/templates/composer/subsection_create.html
+++ b/nofos/composer/templates/composer/subsection_create.html
@@ -126,6 +126,45 @@
           </div>
         </div>
       </details>
+
+      {# --- Optional or Required (USWDS tiles --- #}
+      <fieldset class="usa-fieldset margin-top-2 width-mobile">
+        <legend class="usa-label padding-bottom-1 text-bold">Is this section required or optional for writers to complete?</legend>
+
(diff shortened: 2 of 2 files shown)


pcraig3:
Added 1 more thing, which is the "required/optional" radios on the "new subsection" page.

localhost_8000_composer_a5b55b42-b255-4ced-a659-722b22956266_section_a248d2c8-1113-4f41-a1d4-151514ed0b7b_subsection_create_prev_subsection=d6442db1-1b9e-4879-b67e-e760cce5d709

PR #525, opened 2025-11-20
512 update homepage

jlhogantech:
With only draft ContentGuides:
image
<details>
<summary> outdated screenshot </summary>
Screenshot 2025-11-20 111304
</details>
With no ContentGuides:
(this prob needs some styles for the "No content guides available" text)
Screenshot 2025-11-20 111039

I'm not sure how to mark a ContentGuide as "active"

### DIFF
--- a/nofos/nofos/templates/nofos/nofo_search.html
+++ b/nofos/nofos/templates/nofos/nofo_search.html
@@ -57,7 +57,7 @@
                 {% timezone "America/New_York" %}
                   <td data-sort-value="{{ nofo.updated|date:"YmdHM" }}">
                     <a href="{% url 'nofos:nofo_history' nofo.id %}" title="View audit history">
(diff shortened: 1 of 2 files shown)


pcraig3 (commented):
Looks good! 

I have a few suggestions for cleaning up the table styling a bit, based on some of the stuff I've done on the `nofo_index` page where I was trying to save space.

- Let's just show the user email, not email + name, which is default for `draft.updated_by`
- Let's cut the left+right padding for the delete column, because we don't need so much whitespace there
- Slightly more complicated, but for NOFOs that are more than a day old, I hide the hours and minutes (who cares what time of day you last updated a month ago?)
  - the code for that is kind of messy: we get the "today" value in [`views.py`](https://github.com/HHS/simpler-grants-pdf-builder/blob/main/nofos/nofos/views.py#L238) and then have this check in the template [`nofo_index.html`](https://github.com/HHS/simpler-grants-pdf-builder/blob/main/nofos/nofos/templates/nofos/nofo_index.html#L92)

Here is what the before/after looks like for me

| before | after |
|--------|-------|
|        |  remove username, don't show times if you don't have to, cut padding for last column     |
|   Image     |   Image    |

PR #527, opened 2025-11-21
522 label legends

### LINKED ISSUE #522: Add label tags to top of "section overview" page and also to radio labels
This ticket is a little weird because the designs are not quite updated, but I synced with Annie and @benbrasso-agile6 today and we hashed out what we think the designs should be.

Basically, we noticed this happen in user testing:

1. people initially didn't know what the labels referred to ("some text", "locked", etc)
2. one person pointed out that the radio button text was different than the label, so we should associate them with each other. 

So we have 2 more design interventions to should solve solve both of these:

1. Add a little explainer box to the top of the section overview page
2. Add the labels to the relevant radio buttons

### 1. Add a little explainer box to the top of the section overview page

So, this is a mock-up of the content that we want to have at the top of the section overview page ([link](https://www.figma.com/design/J41kYT5DvfnxYgqzaB6FKP/NOFO-Composer?node-id=754-15307&t=aqvx5pir5QQOLcza-0):

| where it lives | close-up |
|--------|-------|
|  Insitu      |   By itself    |
|   Image     |    Image   |

That would have been nice and easy for us, but @benbrasso-agile6 pointed out that it now looks exactly like all the other accordion headers, so we should have a new design for this so that it is visually distinct (ugh). He's right though.

Okay, so here are the new designs, which live in the "Writer's flow" ([link](https://www.figma.com/design/J41kYT5DvfnxYgqzaB6FKP/NOFO-Composer?node-id=694-14011&t=aqvx5pir5QQOLcza-0)):

| open | closed | similar design in the NOFO Builder | 
|--------|-------|-------|
|     Image   |    Image   | Image | 

☝️ So this is technically a new element we don't have yet, but it is using the same colour as the warning tabs we have in the NOFO Builder. 

Those tabs are too tabby for reuse, but we can make this into a [details/summary element](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/details) and style it up to look like a button that expands to become a big box with a border. 

So that's what I think the best way forward here is. 

```html
<details>
  <summary>Guide to the tags</summary>
  <div>
     <!-- content -->
  </div>
</details>
```
 
I also think it's okay if we put the "Conditional" labels side by side instead of figuring out how to do that stacked layout.

### 2. Add the labels to the relevant radio buttons

This one is simpler, luckily!

At least, simpler to explain.

A good way of associating the labels with the buttons is by literally sticking them in there.
It also clears up which radio buttons are _not_ associated with a label. 

Image

So this story is to get labels into the radio button labels on the "edit subsection" pages. 

Do we need it on the "new" page? If it's easy, why not? If it's annoying, I think we can skip it.

jlhogantech:
closes #522 
| Radio button labels | Tag explainer closed | Tag explainer expanded |
| -- | --- | --- |
| Screenshot 2025-11-21 094247 |Screenshot 2025-11-21 093918 | Screenshot 2025-11-21 094018 |

### DIFF
--- a/nofos/bloom_nofos/static/styles.css
+++ b/nofos/bloom_nofos/static/styles.css
@@ -81,6 +81,23 @@ blockquote {
   background-color: var(--color--optional-purple)
 }
 
+details.explainer-box summary {
+  border-radius: 2px;
+  width: fit-content;
+}
+
+details.explainer-box summary:hover {
+  cursor: pointer;
+}
+
+details[open].explainer-box summary {
+  border-radius: 2px 2px 0 0;
+}
+
+.explainer-box--content {
+  border-radius: 0 2px 2px 2px;
+}
+
 /* General USWDS CSS */
 
 .usa-table thead td,
--- a/nofos/composer/templates/composer/composer_section.html
+++ b/nofos/composer/templates/composer/composer_section.html
@@ -54,9 +54,9 @@
         <li>Select “Edit” to set how much control OpDivs and writers have.</li>
         <li>Tags (e.g., Locked, Certain text, Optional) show those permissions and can be changed when editing.</li>
       </ul>
-      <details>
-        <summary class="accent-warm-dark width-card radius-sm padding-1">Guide to the tags</summary>
-        <div class="border border-accent-warm-dark border-radius-sm">
+      <details class="explainer-box">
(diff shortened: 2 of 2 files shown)


pcraig3:
> 522 label legends

Sounds like an off-brand Switch game

PR #529, opened 2025-11-21
521: detect presence of variables on content guide ingestion

jlhogantech:
The existing content guide, immediately after import:
Screenshot 2025-11-21 165654


pcraig3 (approved):
Works great! 😚 👌

PR #536, opened 2025-11-26
Create a ContentGuideInstance

pcraig3:
## Summary

This PR puts in place the first few views we need to get a ContentGuideInstance off the ground. It's like the ContentGuideInstance the Wright Brothers would have built.

It creates the following writer views:

1. Writer dashboard
2. "Before you start" page
3. Pick a content guide to use page
4. Enter (some) details about your draft NOFO page
5. Archive your draft NOFO

Things that are not in this PR:

- Not all the details are present on the draft NOFO page
- No user permissions stuff exists, so currently any users can use both sides of composer
- 'Published' Content Guides are not a thing yet, so there is no logic for that so far

### Screenshots

Here is what it looks like to create a new ContentGuideInstance

| landing page | before you start page | start page | details page | landing page after creating draft NOFO | Admin page for draft NOFO | 
|--------|-------|----|----|----|----|
|  `/composer/writer`      |   `/composer/writer/before-start`    |  `/composer/writer/start`  | `/composer/writer/details`  | `/composer/writer` | `/admin/composer/contentguideinstance/`  |
|  No draft NOFOs, but you can create a new draft      |  "before you start" list of things to know     | 'start' page: pick a content guide as a basis for your draft NOFO  | Enter initial details about your draft NOFO | New draft NOFO created | Admin page with all draft NOFO details |
|   localhost_8000_composer_writer     |   localhost_8000_composer_writer_before-start    | localhost_8000_composer_writer_start |  localhost_8000_composer_writer_start_d9ee5eca-f245-42c3-8909-d49f54df3916_details | localhost_8000_composer_writer (1) | localhost_8000_admin_composer_contentguideinstance_a92c3355-4bfa-4bf4-bc6c-8b063daeaa91_change_ |

And as a gif

Screen Recording 2025-11-26 at 11 17 34

### How to test this

To test this, just create a Content Guide Instance (note, it requires at least one Content Guide):

1. Visit `/composer/writer`
2. Click the button to start a draft NOFO
3. Arrive on 'before-start' screen. Look at it and proceed
4. Arrive on 'start' screen. Pick a content guide to base your draft NOFO on. This is used to set `parent` but it doesn't do anything else right now
5. Arrive on 'details' screen. Clear out all values and try to submit to trigger error.
6. Enter a value for each field and submit
7. Arrive back on landing screen. You can see your new draft NOFO.
8. Archive your draft NOFO
9. Visit `/admin` and click into "Content guide instances"
10. Find your ContentGuideInstance
11. Make sure that all values are what you entered, and that it is also "archived"
12. Done!

### Details

This PR does a bunch of stuff to set up the foundation for the ContentGuideInstance object. 

#### Set up initial views

There are 5 initial views created, as mentioned above:

1. Writer dashboard
2. "Before you start" page
15. Pick a content guide to use page
16. Enter (some) details about your draft NOFO page
17. Archive your draft NOFO

This gives us a kind of MVP lifecycle for creating a ContentGuideInstance.

I wondered about splitting the writer views into another "views" file (maybe `composer/views/admin_views.py`, `composer/views/writer_views.py`, composer/views/shared_views.py`), and maybe this makes sense later, but it would mean conflicts for parallel work so I figured I'll just add them at the end of the regular views file.

I did add new 'writer' tests in a dedicated tests file though.

#### `created_display` and `updated_display` are timezone-aware

This was an issue previously, that we lost timezone support after we added these helpers. However, it was a pretty small thing to create "timezone-aware" date values and now the template helper works as you would expect.

100% of our users are in EST so we just hardcode it. This could be a problem eventually but for now it is not.

#### Error summary block

Because the "details" page is pretty complicated, I created a summary block for multiple errors (if all values are empty, it will trigger).

It's probably not something we need very often, but on pages with multiple values that require validation enter, it can be useful. 

#### Add admin.py configuration for ContentGuideInstance

At this point, the only way to actually see what is in a saved ContentGuideInstance is to use the superuser admin view, so I added some code that gives us a list view and an individual view. This way, we can see that our guides are saving correctly with all the values we expect, and also delete them fully if we want to (other than just archiving them).

#### Clean up icon colour CSS

The CSS for icons is pretty messy and bad, and one of the worst things was adding colours because we are using a CSS filter to change the SVG colour. 

This has meant that icon colours are non-obvious when you look at them in the CSS. So I went through and pulled most of the filter colours out of the previous blocks they are in and created dedicated classes for common icon colours. The idea is that you should be able to add classnames to your link or button to (a) give it an icon, (b) decide if it goes before or after the content, and now (c) choose its colour. 

It's not 100% amazing, but it seems like an improvement. 

The reason to do this at all was that I was building that "download" link present on both landing pages of Composer.

### DIFF
--- a/nofos/composer/tests/test_views_writer.py
+++ b/nofos/composer/tests/test_views_writer.py
@@ -63,13 +63,13 @@ class WriterDashboardViewTests(TestCase):
             title="ACF Content Guide",
             opdiv="acf",
             group="acf",
-            status="active",
+            status="draft",
         )
         hrsa_guide = ContentGuide.objects.create(
             title="HRSA Content Guide",
             opdiv="hrsa",
             group="hrsa",
-            status="active",
+            status="draft",
         )
 
         self.client.login(email="acf@example.com", password="testpass123")
@@ -105,13 +105,13 @@ class WriterDashboardViewTests(TestCase):
             title="ACF Content Guide",
             opdiv="acf",
             group="acf",
-            status="active",
+            status="draft",
         )
         hrsa_guide = ContentGuide.objects.create(
             title="HRSA Content Guide",
             opdiv="hrsa",
             group="hrsa",
-            status="active",
+            status="draft",
         )
 
         self.client.login(email="bloom@example.com", password="testpass123")
@@ -188,13 +188,13 @@ class WriterInstanceStartViewTests(TestCase):
             title="ACF Guide",
             opdiv="acf",
             group="acf",
-            status="active",
+            status="draft",
         )
         hrsa_guide = ContentGuide.objects.create(
             title="HRSA Guide",
             opdiv="hrsa",
             group="hrsa",
-            status="active",
+            status="draft",
         )
 
         self.client.login(email="acf@example.com", password="testpass123")
@@ -213,13 +213,13 @@ class WriterInstanceStartViewTests(TestCase):
             title="ACF Guide",
             opdiv="acf",
             group="acf",
-            status="active",
+            status="draft",
         )
         hrsa_guide = ContentGuide.objects.create(
             title="HRSA Guide",
             opdiv="hrsa",
             group="hrsa",
-            status="active",
+            status="draft",
         )
 
         self.client.login(email="bloom@example.com", password="testpass123")
@@ -238,7 +238,7 @@ class WriterInstanceStartViewTests(TestCase):
             title="ACF Guide",
             opdiv="acf",
             group="acf",
-            status="active",
+            status="draft",
         )
         self.client.login(email="acf@example.com", password="testpass123")
 
@@ -290,7 +290,7 @@ class WriterInstanceDetailsViewTests(TestCase):
             title="ACF Core Content Guide",
             opdiv="acf",
             group="acf",
-            status="active",
+            status="draft",
         )
 
         self.url = reverse(


nofos/composer/templates/composer/composer_section_edit.html
```diff
@@ -63,7 +63,7 @@ <h3 class="usa-accordion--heading margin-top-0">
             </div>
             <div class="grid-col-12 display-flex flex-align-center margin-y-3">
               <div class="flex-1 border-bottom border-base-lighter border-bottom-width-1px"></div>
-              <a class="usa-button usa-button--light usa-button--outline margin-right-0" href="{% url 'composer:subsection_create' document.pk section.pk %}?prev_subsection={{subsection.id}}">+ Add section</a>
+              <a class="usa-button usa-button--outline usa-button--icon--before usa-button--icon--usa-blue usa-button--icon--usa-blue--hover usa-button--add" type="button" href="{% url 'composer:subsection_create' document.pk section.pk %}?prev_subsection={{subsection.id}}">Add section</a>
```
jlhogantech: **non blocking** why the preference for template-specific css rules that require wrapping classes over the util class like `usa-button--light`?

pcraig3: You're right, I just added this in because I was adding the plus icon to the button and so I duplicated what we did with the other add button.

I've restructured it somewhat to remove the template-specific CSS rule since clearly we don't really care what page this button is on, so long as it looks like how we expect.

nofos/composer/forms.py
```diff
@@ -89,3 +95,88 @@ class ComposerSubsectionInstructionsEditForm(forms.ModelForm):
...
+            "short_name": "A name that makes it easier to find this NOFO later. It won’t be public.",
+            "title": "The official name for this NOFO.",
+            "number": "The opportunity number for this NOFO.",
+        }
+
+    def __init__(self, *args, user=None, **kwargs):
+        super().__init__(*args, **kwargs)
+
+        # Make opdiv + title required
+        if "opdiv" in self.fields:
```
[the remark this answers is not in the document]
pcraig3: You are right to call this out. I figured that you could manually delete the HTML fields in DevTools to submit only 1 field, but I just tried it and they all show up anyway. 

So it seems like they will always be here, which means we can ditch the "if" statements.

PR #540, opened 2025-11-28
Move configure section button, add expand all button

### LINKED ISSUE #533: Move “configure section” link and implement “expand all” button
So “configure section” is clearly in the wrong place and from now on it will live at the top of a section.

Here is what that looks like ((link)[https://www.figma.com/design/J41kYT5DvfnxYgqzaB6FKP/NOFO-Composer?node-id=754-23033&t=vgPruR8YfwAq0wRU-0]): 

Image

I think the reason it was at the bottom initially was because the “Expand all” button was in the same place. This is not something we implemented yet, so we can do both at the same time.

- The "Configure section" link is already working fine, we just want to move it.
- The “Expand all” should open _all_ the accordions. If at least one accordion on the page is _not_ open, it should say “Expand all”. Once all are open, it should change to say “Collapse all” (and the plus changes to a minus).

jlhogantech:
closes #533

Q: Should I pull out the javascript into a separate file and include it in the template?

expand_all

### DIFF
--- a/nofos/bloom_nofos/static/js/composer/composer_section.js
+++ b/nofos/bloom_nofos/static/js/composer/composer_section.js
@@ -1,0 +1,65 @@
+document.addEventListener('DOMContentLoaded', function() {
+  const expandCollapseBtn = document.getElementById('expand-collapse-all');
+  const accordionButtons = document.querySelectorAll('.usa-accordion__button');
+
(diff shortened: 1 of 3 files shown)


pcraig3 (approved):
👍  Approved!

pcraig3:
Gonna merge

PR #548, opened 2025-12-03
Add confirmation page to Content guide instance

pcraig3:
## Summary

This PR adds a new “Confirmation” page which plays back all your answers to the Details page, and both Conditional Questions pages. 

Note that the ContentGuideInstance object is actually when you "continue" on the Details page, so we have the object by this point. The confirmation page is giving you the chance to check your answers, but it doesn't create the new Instance object.

Also note that pressing the “edit” button on the confirmation page doesn’t plop you back on the confirmation page once you save, you will have to go through all remaining pages again. This is not ideal but I don’t really think we need to make this more complex right now. I am gonna make another ticket for this (low priority).

### Screenshot 

localhost_8000_composer_writer_ad9e513d-971d-4560-8099-dfc7abb29143_review

### How to test

(Create a ContentGuideInstance object)

1. Visit `/composer/writer` and start a draft NOFO
2. Complete `/before-start`
3. Complete `/start`
4. Complete `/details`
5. Complete both `/conditional-questions` pages
    1. Note that there is a new back button, so you can press "back" too and check it out
6. Okay! Now you are on the confirmation page
7. Check that all your answers are right
8. Use "edit" links, see that they go back to the right place (note that you have to step through all pages again)
9. "Get started!" takes you back to the dashboard page and you see the "created" alert

### Creating the confirmation page

On the confirmation page, we have to loop through all the questions and all the answers. For the “details” questions, we can get them from the form itself, because it lists all the keys for the fields we are entering. Using those keys, we can get the question labels, and then we can get the answers to them by asking for the key from the instance. 

For the Conditional Questions, we can get all questions for a given page, and then get the answers by using the question label and looking inside `instance.conditonal_quesitons[key]`. So it’s not too bad.

Note that I did have to extend the “details” view, because it didn’t have a previous way to hydrate that form if we already had an instance. Previously, we were always creating a new ContentGuideInstance so it would always be empty when we came to this page.

### Adding back links

I noticed that all the mocks include a back link beside “save and continue” (same in the USWDS example here), so I went ahead and added those. 

Note that I did something a little weird, which is that I used the flex: row-reverse to reveerse the order in which they appear in the markup. The reason for this is that I think the “tab” action should hit the “Save” button before it hits the back button because “save” is the action you are going to want to do 90% of the time.

### Refactoring writer tests setUp method 

Lots of our tests were doing very similar setup steps all the way down, and I figured there was a good opportunity here to do some refactoring.

### DIFF
--- a/nofos/composer/views.py
+++ b/nofos/composer/views.py
@@ -1230,6 +1230,8 @@ class WriterInstanceConfirmationView(LoginRequiredMixin, TemplateView):
       1) Details fields (opdiv, agency, title, short_name, number)
       2) Conditional questions for page 1
       3) Conditional questions for page 2
+
+    Note: this view assumes the max page number is 2
     """
 
     template_name = "composer/writer/writer_confirmation.html"
@@ -1293,7 +1295,6 @@ class WriterInstanceConfirmationView(LoginRequiredMixin, TemplateView):
     def get_context_data(self, **kwargs):
         context = super().get_context_data(**kwargs)
 
-        # 1. Details
         context["details_fields"] = self._build_details_fields()
         context["page1_questions"] = self._build_conditional_questions_for_page(1)
         context["page2_questions"] = self._build_conditional_questions_for_page(2)
--- a/nofos/composer/conditional/conditional_questions.py
+++ b/nofos/composer/conditional/conditional_questions.py
@@ -59,7 +59,11 @@ class ConditionalQuestionRegistry(dict[str, ConditionalQuestion]):
 
     @property
     def max_page(self) -> int:
-        """Highest page number in the registry (for navigation)."""
+        """
+        Highest page number in the registry (for navigation).
+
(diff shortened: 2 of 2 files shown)


nofos/composer/views.py
```diff
@@ -1164,13 +1212,113 @@ def form_valid(self, form):
...
+            )
+
+        return items
+
+    def get_context_data(self, **kwargs):
+        context = super().get_context_data(**kwargs)
+
+        # 1. Details
+        context["details_fields"] = self._build_details_fields()
+        context["page1_questions"] = self._build_conditional_questions_for_page(1)
```
jlhogantech: **nit**: this is the first time page count is hardcoded, but I think it's fine and not worth additional complexity to make dynamic. Just maybe worth calling out somewhere (in conditional_questions.json/.py?) so easy to remember this will need to change if more question pages are added

PR #559, opened 2025-12-04
549 writer subsection edit

### LINKED ISSUE #549: 1/4: Add basic edit subsection page for writers
This is one of 4 stories to complete the "edit" page for a NOFO writer.

The reason why this story is so small is so that we can get this view in for the basic functionality and then work somewhat independently on it.

Subsections can have 3 edit modes:

1. `full`: full editing allowed (✅  handled in this issue)
2. `variables`: only allow editing of the variables that are in the body content (❌  not handled in this issue)
3. `locked`: no editing allowed

### Adding instructions

If there are instructions, we want to show those instructions on the Edit page. [Link to Figma](https://www.figma.com/design/J41kYT5DvfnxYgqzaB6FKP/NOFO-Composer?node-id=1765-24997&t=aR6rMYTZ5rQACDF6-0) (requires access).

Image

Unlike with sys admins, we want writers to be able to see the instructions while filling out content, so the instructions should sit between the heading and the form element(s).

Looks like we want to add a horizontal spacer above and below the instructions. We have been using `<hr>` inside of a `usa-prose` div for this in the past few stores.

Also, padlock icon in the top corner.

### Edit mode: full

For this issue, we only want to add the edit mode for `full`, which means giving writers the exact same markdown editing window that we are giving to system admins. 

Writers should be able to update all the content, or even delete all of it if they want. No validation needed.

### All "edit subsection" issues

1. (this one) [Adding subsection page w/ instructions](https://github.com/HHS/simpler-grants-pdf-builder/issues/549)
2. [Adding variables](https://github.com/HHS/simpler-grants-pdf-builder/issues/550)
3. [Adding optional/required](https://github.com/HHS/simpler-grants-pdf-builder/issues/551)
4. [Adding not started/started/done](https://github.com/HHS/simpler-grants-pdf-builder/issues/552)

jlhogantech:
closes #549 

just leaves things pretty basic for edit_mode = "variables" or "locked", to be handled in later tickets
Screenshot 2025-12-04 182822

### DIFF
--- a/nofos/composer/views.py
+++ b/nofos/composer/views.py
@@ -1501,14 +1501,13 @@ class WriterInstanceSubsectionEditView(GroupAccessContentGuideMixin, UpdateView)
         context["instance"] = self.instance
         context["section"] = self.section
         context["subsection_variables"] = render_curly_variable_list_html_string(
-            self.get_object().extract_variables()
+            self.object.extract_variables()
         )
         return context
 
     def form_valid(self, form):
-        messages.add_message(
+        messages.success(
             self.request,
-            messages.SUCCESS,
             "Updated section: “<strong>{}</strong>” in ‘<strong>{}</strong>’".format(
                 self.object.name or "(#{})".format(self.object.order),
                 self.object.section.name,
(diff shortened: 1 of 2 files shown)


nofos/composer/views.py
```diff
@@ -645,6 +646,13 @@ def get_context_data(self, **kwargs):
             ),
             args=[document.pk],
         )
+
+        # Only name and not full URL because each subsection id is needed for full URL
+        context["edit_subsection_url_name"] = (
+            "composer:subsection_edit"
+            if document_field_name == "content_guide"
+            else "composer:writer_subsection_edit"
+        )
```
pcraig3: Right, that makes sense.

PR #583, opened 2025-12-11
Add redirect logic for not-fully-initiated ContentGuideInstance

### LINKED ISSUE #555: Return to a partially started ContentGuideInstance
Right now, if you start a ContentGuideInstance, but don't make it past the Confirmation page, your draft NOFO exists, but it is not possible to go back to it and continue filling it out. 

I think we should put you back on the last page you were on when you stopped filling in the draft NOFO, with the exception o the "details" page, because if you don't make it past the "Details" page, there is no draft NOFO yet.

So:

- if you were on the Conditional Questions 1 page, you go back there.
- if you were on the Conditional Questions 2 page, you go back there.
- if you were on the Confirmation Page, you go back there

The way we can figure this out is by looking at what exists on the ContentGuideInstance itself. 

If no conditional questions, go to page 1.
If half (maybe count them?) go to page 2.
If all of them, go to confirmation page

After this, the link on the index page for Composer Writers should take you there, and we can add a "Finish this draft NOFO later" link just above the buttons on the Confirmation page, [like we see in the mocks](https://www.figma.com/design/J41kYT5DvfnxYgqzaB6FKP/NOFO-Composer?node-id=711-16556&t=6IUUxQ8L9FIiCwg5-0) (Figma access required)

jlhogantech:
Closes #555 

- If any unanswered questions for page 1, direct user to page 1
- else if any unanswered questions for page 2, direct user to page 2
- else if not yet sections, direct user to 'confirmation' page
- else redirect to section view for section 1

### DIFF
--- a/nofos/composer/tests/test_views_writer.py
+++ b/nofos/composer/tests/test_views_writer.py
@@ -391,8 +391,8 @@ class WriterInstanceDetailsViewTests(BaseWriterViewTests):
         self.assertEqual(
             response.url,
             reverse(
-                "composer:writer_conditional_questions",
-                kwargs={"pk": str(instance.pk), "page": 1},
+                "composer:writer_instance_redirect",
+                kwargs={"pk": str(instance.pk)},
             ),
         )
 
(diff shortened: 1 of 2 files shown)


pcraig3:
I made a few extra changes to this PR:

1. Add a "Finish this NOFO later" link just above "Get started!" on the Writer Confirmation page
2. I realized we can use this redirect route for every draft NOFO page. If we are doing it the first time, we go to the next page to fill out. If we are coming back from the confirmation page, we go back to it. 👍  Seems to work great.
3. Also I got rid of the success message that had a TODO to get rid of it.

Issue #577, opened 2025-12-10
2/2: Look for instructions titles as the starting text of the subsection body if no subsection.name


pcraig3:
# Summary

This is a follow-on ticket from #576, and it is much smaller. 

If there is a subsection with no name, we should try to match the instructions title against the _start_ of the subsection body.

## Details

This is fairly specifically about the initial "Tagline" instruction. Since there is no name to this subsection, there is no way to find it. 

Incidentally, the "Tagline" instruction box _is_ getting matched to a subsection in the "Contacts and support" Step of the NOFO: Approach. This is not what we want.

Once we have an algorithm for isolating the title from an instructions box we should check the beginning of the body text of subsections without a name for a potential match. This should work for Tagline, and -- hopefully -- doesn't mess anything else up.

PR #585, opened 2025-12-11
577 match subsection body if no name

### LINKED ISSUE #576: 1/2: Use instructions titles for better matching
# Summary

The idea here is that we should figure out where the instructions "title" is rather than just matching across the entire text of the instructions.

## Screenshot 

Image

## Details 

Right now, we are using a pretty loose matching algorithm for instructions (inside [`add_instructions_to_subsections`](https://github.com/HHS/simpler-grants-pdf-builder/blob/c987c9cab65c8356f45f0905eae7c26ed5ef9ee3/nofos/nofos/nofo.py#L2418)), which is slightly problematic for another ticket we want to complete: [WIP] “use instructions text to figure out X about the subsection” idea. 

Currently we match by saying "look for the lowercased `subsection.name` in the instructions text, which it turns out, happens fairly often. 

To solve this, we should match on the name of the instructions block, not all the content,

### Match on the name of the instructions block, not all the content

Early in the NOFO (`Step 1` > `Eligibility`), we have a subsection called “Eligibility”, followed by a section called “Eligible applicants”. The instructions that start with “Instructions for NOFO Writers: Eligible applicants” are being attached to the first subsection (“Eligibility”) not the second one where the title is an exact match. 

In the body of the instructions, we see “Eligibility depends on the grant program’s statutory authority.”, so that is what is getting matched. 

Instead, we should do something like:

- Get first paragraph of instructions box.
- Split on first colon, drop the first part (ie, “Instructions for NOFO writers:”), take the remainder, strip whitespace 
- Lop off anything in brackets
- Get subsection name, also lop off brackets
- Now, do a stronger match for the subsection name and `instructions_title`
  - This can either be "exact match, insensitive" or "find in, with case sensitivity" (see 3 below)

It is more strict, but I think it works fine, after running through the content guide.

### Edge cases: 3 matches we want to make sure still work.

1. Extra stuff in instructions 

If we see something like “Cost sharing” and instructions are like “Cost sharing (YES)”, we want to keep that. This is why we should remove brackets from instructions boxes heading. 

2. Extra stuff in subsection.name

If we see “Instructions for NOFO Writers: Project narrative” and the subsection.name is “Project narrative (100 points)“, we want to make sure those still match. That is why we want to lop off brackets on the subsection titles.

3. Whole title matching

There is a subsection in Step 1 that is called “Approach”, followed by “Overview of approach”. The instructions look like “Instructions for NOFO Writers: Overview of approach”.   As currently written, we match the instructions to the “Approach” subsection, because it partially matches the title phrase. While some of the other mismatches are about finding the `subsection.name` in the body of the instructions content, this one is actually in the title portion of the instructions.

There are 2 ways of solving this:

- Case sensitive match ("Approach" would not match "Overview of approach")
- Match the whole name to the whole title (can be case-insensitive)

I don’t really mind which one, whichever seems the least problematic. 

### Test cases

I looked through the instructions we are matching incorrectly and here are the ones I came up with:

- Step 1
    - Eligibility ("Eligible applicants")
    - Program description ("Application contents and format")
    - Background (“Overview”) 
    - Overview (“Purpose”)
    - Purpose (“Program logic model”)
    - Approach (“Overview of approach”) 
- Step 2
    - Grants.gov (“Other required forms”)
- Step 3
    - Attachments (“Project narrative”)
    - Programmatic (“Cooperative agreement terms”)
- Step 5
    - Application (“Option 2”) 
- Contacts and support
    - Program (“Tagline”)

jlhogantech:
closes #576 , #577 

**Part 1: Update instruction matching logic**
Matching instructions steps for a given instructions table:

find instructions title as expected based on presence of "instructions for nofo writers" or "instructions for new nofo team"
extract the part of the title after the ":"
remove anything in trailing ()
Should this also check for square brackets or are we sure it will be parens?
For each subsection, remove any trailing parens from the name
compare instructions title lower to subsection name lower
If match, instructions are for that subsection

**Part 2: match on body if no subsection name**
Updates instruction matching logic to check the instructions title against the start of the subsection body if the subsection has no name

### DIFF
--- a/nofos/nofos/nofo.py
+++ b/nofos/nofos/nofo.py
@@ -2437,6 +2437,56 @@ def add_instructions_to_subsections(sections, instructions_tables):
     Returns:
         None: The function modifies the sections in place.
     """
+
+    def _extract_instructions_title(instructions_body):
+        """
+        Extract and return the instructions title (text after the first colon) or None.
+        """
+        logger = logging.getLogger(__name__)
+
+        try:
+            node = instructions_body.find(
+                string=re.compile(
+                    r"^(instructions for nofo writers|instructions for new nofo team)",
+                    re.IGNORECASE,
+                )
+            )
+
+            if not node:
+                raise ValueError("No instructions title line found")
+
+            title_text = str(node).strip()
+
+            if ":" not in title_text:
+                raise ValueError("Instructions title is missing a colon")
+
+            # Split on first colon
+            _, title = title_text.split(":", 1)
+            title = title.strip()
+
+            if not title:
+                raise ValueError("Instructions title is empty after colon")
+
+            return title
+
+        except Exception as e:
+            logger.warning(
+                "Failed to extract expected instructions title from instructions body.",
+                extra={
+                    "error": str(e),
+                    "instructions": instructions_body,
+                },
+            )
+            return None
+
+    def _normalize_title(title: str):
+        """Trim and remove trailing parentheses, then normalize case for comparison."""
+        REMOVE_TRAILING_PARENS_REGEX = r"\s*\(.*\)\s*$"
+
+        normalized = title.strip()
+        normalized = re.sub(REMOVE_TRAILING_PARENS_REGEX, "", normalized)
+        return normalized.casefold()
+
     logger = logging.getLogger(__name__)
     all_subsections = [
         subsection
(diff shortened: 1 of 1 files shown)


nofos/nofos/nofo.py
```diff
@@ -2450,21 +2450,75 @@ def add_instructions_to_subsections(sections, instructions_tables):
...
+            instructions_title = re.match(
+                r"[^:]*:\s*(.*)", instructions_title.get_text().strip()
+            ).group(1)
+        except AttributeError:
+            logger.warning(
+                "Failed to extract expected instructions title from table.",
+                extra={"table": instruction_table.prettify()},
+            )
+            # TODO: should we continue or skip here?
+            continue
```
pcraig3: A few comments here:

1. use `str` instead of `get_text`

For `instructions_title.get_text().strip()`

I think we can just cast this using str which is a bit safer, because `get_text` can return an error depending on what specifically is returned by `find()` on line 2464.

So we can do `str(instructions_title).strip()`

2. use `split` rather than regex

We can also use split on the first colon (which your regex also does, but I find the splits more readable).

```
title_text = str(instructions_title).strip()

if ":" not in title_text:
    logger.warning(
        "Instructions title is missing a colon.",
        extra={"title_text": title_text},
    )
    continue

_, instructions_title = title_text.split(":", 1)
instructions_title = instructions_title.strip()
```

3. Maybe create a private internal function to get the title?


For this “get the title from the instructions table” logic, what do you think of creating an embedded private method like this:  

``` 
def _extract_instructions_title(instructions_body, instruction_table):
    """
    Extract and return the instructions title (text after the first colon) or None
    """
    try:
        node = instructions_body.find(
            string=re.compile(
                r"^(instructions for nofo writers|instructions for new nofo team)",
                re.IGNORECASE,
            )
        )

        # If no matching line was found at all
        if not node:
            raise ValueError("No instructions title line found")

        title_text = str(node).strip()

        if ":" not in title_text:
            raise ValueError("Instructions title is missing a colon")

        # Split on first colon
        _, title = title_text.split(":", 1)
        title = title.strip()

        if not title:
            raise ValueError("Instructions title is empty after colon")

        return title

    except Exception as exc:
        logger.warning(
            "Failed to extract expected instructions title from table.",
            extra={
                "error": str(exc),
                "table": instruction_table.prettify(),
            },
        )
        return None 
```

And then in the body of the main function, we can do this:  

``` instructions_title = _extract_instructions_title(
    instructions_body, instruction_table
)
if not instructions_title:
    continue

```

PR #586, opened 2025-12-11
Detect optional from instructions text

### LINKED ISSUE #578: Detect optional subsections on import
# Summary

When we import a ContentGuide into the Composer, we just default them all to non-optional. However, the instructions text reliably indicates whether or not a subsection is optional, so we can add some logic to detect subsections that should be `optional`. 

### Details

After reviewing the content guide, I have determined that almost all (if not all) of the instructions boxes include one of these phrases as part of their first bullet point:
- “This section is required for all NOFOs...”
- “This section is only required if…“

Note that I have a slightly adjusted version of the Content Guide (`2025-12-10`) where I have corrected the couple of instructions that didn't follow this format. 

Before doing this story, we want to complete the two about better instructions matching (#576 and #577) so that we know the instructions that we have are actually getting assigned to the right subsections.

Once we have that, the rest is pretty simple. If this phrase appears in the instructions text at all “This section is only required if”, then we can mark that section as optional when we import the Content Guide as a SysAdmin in the NOFO Composer.

jlhogantech:
closes #578 

Updates subsection creation to detect optional subsections based on presence of "section is only required if" in instructions

### DIFF
--- a/nofos/nofos/nofo.py
+++ b/nofos/nofos/nofo.py
@@ -420,12 +420,18 @@ def _build_document(document, sections, SectionModel, SubsectionModel):
 
             if hasattr(SubsectionModel, "extract_variables"):
                 variables = subsection_obj.extract_variables()
+                # If variables are detected, edit_mode = "variables"
                 if variables:
                     subsection_obj.edit_mode = "variables"
                     serializable_variables = {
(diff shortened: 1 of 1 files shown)


pcraig3 (approved):
Awesome stuff!

PR #592, opened 2025-12-16
Use a pencil icon in ToC for "Build your application"

pcraig3:
## Summary

This is a very small update: we now use a pencil icon on the ToC and Section page for "Build your application"

Essentially, some of the step names have changed and now we need to update our icon macro.

### Screenshot 

| before | after |
|--------|-------|
|  Magnifying glass icon is wrong here (Step 1 already uses it)      |    Pencil icon is good ✏️    |
|    Screenshot 2025-12-16 at 18 26 21    |    Screenshot 2025-12-16 at 18 25 39   |

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,10 +11,13 @@ Versioning since version 1.0.0.
 
 ### Changed
 
+- Use pencil icon for "Build your application" step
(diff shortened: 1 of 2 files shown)


pcraig3:
Merging because Yasmine needs this for an upcoming NOFO.

PR #595, opened 2025-12-18
556: Add history page for contentguideinstance (Draft NOFO)

### LINKED ISSUE #556: [WIP] Change history for a ContentGuideInstance

jlhogantech:
closes #556 
- adds history page for ContentGuideInstance, with associated view
- updates SectionView to redirect to this page when 'writer' mode
- adds handling for 'created' events to `format_audit_event` for consistency in displayed object types

| without handling for 'created' events | with handling | 
| --- | --- | 
| Screenshot 2025-12-18 093511 | Screenshot 2025-12-18 093458 |

### DIFF
--- a/nofos/nofos/tests_nofos/test_history.py
+++ b/nofos/nofos/tests_nofos/test_history.py
@@ -5,6 +5,8 @@ from django.urls import reverse
 from django.utils import timezone
 from easyaudit.models import CRUDEvent
 
+from nofos.audits import remove_model_from_description
+
 from ..models import Nofo, Section, Subsection
 
 User = get_user_model()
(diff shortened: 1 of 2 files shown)


nofos/nofos/tests_nofos/test_history.py
```diff
@@ -64,7 +64,7 @@ def test_history_view_shows_regular_crud_events(self):
 
         self.assertEqual(response.status_code, 200)
         self.assertContains(response, event.get_event_type_display())
-        self.assertContains(response, str(self.nofo))
+        self.assertContains(response, "NOFO Created")
```
jlhogantech: Previously: `([nofo object pk) NOFO Created` 

so ultimately, this feels definitely OK and maybe even preferable, as I don't think the pks are actually something we expect folks to know about/use, right? 
Can make an exception or totally reverse this change if I'm wrong here, just let me know!

pcraig3 (commented):
There is a little inconsistency over our toString printing of the ContentGuide name that I mentioned above (you can manually type in the URL to change the title to see this).

But, more annoyingly, there seem to be 3 audit events created when you edit a subsection as NOFO writer.

| what you see on the audit history page | change 1: viewed | change 2: the actual change I made | change 3: not sure (maybe "done"?) |
|--------|-------|------|------|
|   Image     |    Image   |  Image | Image | 

What I would suggest is to make the `_filter_updated_events` method so that it doesn't allow events where changed fields:
- has only 1 key: `status`
- is null (we should have already added this there)

PR #602, opened 2025-12-30
593 subsection diff

### LINKED ISSUE #593: Create a diff page for Subsection updates (for all NOFO-like objects)
So in https://github.com/HHS/simpler-grants-pdf-builder/issues/556, we want to create the content guide audit page for ContentGuideInstances.

The follow-up that we heard from a few people in testing was "what changed?", which is the natural follow-up question to ask. 

Here is more or less what I am thinking this screen could look like:

| before | after |
|--------|-------|
|  "audit log" page       |   "diff" page for subsection update (remove the Bloomworks tag)    |
|   Image     |    Image   |

We already have a bunch of diffing logic in the [`nofo_compare.py`](https://github.com/HHS/simpler-grants-pdf-builder/blob/main/nofos/nofos/nofo_compare.py) file, which should be reusable. 

The actual comparison happens here: https://github.com/HHS/simpler-grants-pdf-builder/blob/main/nofos/nofos/nofo_compare.py#L208

The reason there is so much code there is because we are trying to match subsections to each other across NOFOs, which is a thing we don't have to manage with this story (luckily). 

There is another step in there where we generate "just adds" or "just deletes" and we put it in a little data class that's defined at the top: https://github.com/HHS/simpler-grants-pdf-builder/blob/main/nofos/nofos/nofo_compare.py#L381

That dataclass is meant for when you have 2 subsections you are comparing against each other, so in this case, we most likely don't need that. 

If you update a subsection, you can find the audit event for it in the "CRUD events" section of the admin UI, and then if you click the verb ("Update") you get a screen that looks like this:

Image

For our purposes, we want to run a diff on the "changed_fields" > "body" (body[0] is the old version, and body[1] is the new version). Once we have that combined diff, we can decompose the "ins" and "del" elements to get the "only deletions" render and the "only adds" render.

Note that I am running this UI past Annie, so it may change slightly, but can't imagine it changes substantially. 

Note that this issue does _not_ include diffs for anything related to the NOFO object or the Section objects (just Subsections).

jlhogantech:
closes #593 

- Adds a view to see the 'diff' for changes to a ContentGuide subsection. 
- Makes `Update` a clickable link to see the new 'diff' view for any "Contentguideinstance"  
- Currently works for changes to 'variables' and 'body', but the link is present for any update event for "Contentguideinstance" so it is possible a user could land on the 'diff' view without a viewable diff.
    - Should we: add explicit condition to the check before adding the link? Or add an empty state for other Update events that get clicked?
- In order to display variable diffs, extracts the `replace_variables_with_values` from templatetag into a util function that is used in the view in addition to the templatetag  

| Variables diff (add values) | Variables diff (change value) | Body diff | 
| --- | --- | --- |
| Screenshot 2025-12-29 214456 | Screenshot 2025-12-29 214510 | Screenshot 2025-12-29 214429 |

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ Versioning since version 1.0.0.
 - Added new theme: "SAMHSA Light"
 - Add "README" overview to the composer app
 - Add an audit history page for the Composer Writer app
+- Add a "compare" page for UPDATE audit actions in the Composer Writer
 
 ### Changed
 


nofos/composer/views.py
```diff
@@ -1779,3 +1785,69 @@ def get_section_model_name(self):
...
+                markdownify(subsection.body), parse_variables(new)
+            )
+            if has_diff(diff := html_diff(body_with_old, body_with_new)):
+                context["diff_old"] = extract_old_diff(diff)
+                context["diff_new"] = extract_new_diff(diff)
+
+        return render(request, "composer/writer/writer_history_compare.html", context)
+
+
+def safe_get_changed_fields(event):
```
jlhogantech: Should these helper functions be defined within the view, since they are only used there? Or maybe extracted into utils?

nofos/nofos/audits.py
```diff
@@ -152,7 +153,14 @@ def _filter_events(events):
                 changed after initial update (e.g. filename, conditional_questions).
         """
         EXCLUDED_FIELDS = set(
-            ["updated", "status", "conditional_questions", "filename"]
+            [
+                "updated",
+                "status",
+                "hidden",
+                "optional",
```
jlhogantech: Added `optional` and `hidden` to the list of excluded `changed_fields` -- is this OK? would need different UI treatment to display these changes

pcraig3 (commented):
I think this PR is working great! 

A couple notes:

## What do to about empty diffs

> the link is present for any update event for "Contentguideinstance" so it is possible a user could land on the 'diff' view without a viewable diff.
> Should we: add explicit condition to the check before adding the link? Or add an empty state for other Update events that get clicked?

I think we shouldn't do anything just now. I experimented with making diffe> rent types of non-body changes to the subsections (changing the status, changing "hidden" values, etc) and I wasn't able to find any empty diffs. Unless you can find any scenarios that result in a user seeing an empty diff page, I am good to assume that we are basically okay on this.

## handling tables and lists

I am pleased to find that it works as expected for tables and lists:

| diffing a list | diffing a table |
|--------|-------|
|  Image      |     Image  |

## Merged the previous PR

I realize that in merging #595, it may have messed up some of your code for this PR. I don't think I made any changes which were very serious — I mostly updated the `sort_key` function and some tests — so I _hope_ that it's not a big annoying problem to get this rebased on top of the main branch.

nofos/composer/templates/composer/writer/writer_history_compare.html
```diff
@@ -0,0 +1,31 @@
+{% extends 'base.html' %}
+{% load static tz martortags add_classes_to_tables add_footnote_ids add_captions_to_tables add_classes_to_lists add_classes_to_paragraphs convert_paragraphs_to_hrs conditional_questions_note nofo_section_name_separator %}
+
+
+{% block title %}
+  Update to "{{ subsection.name }}"
```
jlhogantech: Is this what you meant about H1 logic reuse?

Issue #566, opened 2025-12-08
Update Python libraries


pcraig3:
This is a chore but one that we have to check the various libraries for.

```
django       5.2.9 6.0   A high-level Python web framework that encourages rapid development and clean, pragmatic design.
django-ninja 1.4.5 1.5.1 Django Ninja - Fast Django REST framework
markdown     3.9   3.10  Python implementation of John Gruber's Markdown.
markdownify  1.2.0 1.2.2 Convert HTML to markdown.
```

### django

Supposed breaking change! We should check it out before we upgrade and make sure nothing is wrong.

### django-ninja

I did upgrade this originally and it _does_ break, so this is a breaking change for us.

### markdown & markdownify

These are not supposed to be breaking changes, but our app is so dependent on markdown conversions that we need to really make sure we haven't messed anything up here. 

We have some tests for markdown stuff, so that's the first thing to check. 

Other than that, I would suggest import a NOFO before and after and scroll through on 2 screens.

## Needs updating

- [ ] django
- [x] django-ninja
- [x] markdown
- [x] markdownify

pcraig3:
Markdown and markdownify have been updated #604

PR #604, opened 2026-01-02
Update poetry libs + remove deprecation warnings

pcraig3:
## Summary

This PR updates our markdown-related libraries, and makes changes based on the deprecation warnings that were coming up when testing with the `-Wa` flag on. 

It does _not_ update to app to Django 6, because one of our libraries ([django-easy-audit](https://github.com/soynatan/django-easy-audit)) doesn't support it (yet).

### Deprecation warnings

The documentation for [updating to a new major Django version](https://docs.djangoproject.com/en/6.0/howto/upgrade-version/) mentions that we should make sure to [resolve deprecation warnings](https://docs.djangoproject.com/en/6.0/howto/upgrade-version/#resolving-deprecation-warnings) before the upgrade.

Specifically, this is done by using the `-Wa` flag, which looks like this for us:

```
poetry run python -Wa manage.py test
```

Based on that, here are the changes I made:

- Replace CheckConstraint.check with condition to match Django 6 API
- Explicitly set URLField assume_scheme="https" to silence default-scheme warning
- Update BeautifulSoup find_all() usage to use `string=` instead of deprecated `text=`
- Stop calling format_html() without placeholders; use mark_safe for prebuilt HTML

After these changes, it seems like _our_ code is ready for Django 6, but unfortunately, we are waiting for [django-easy-audit](https://github.com/soynatan/django-easy-audit) to also be Django 6 compatible.

When upgrading to Django 6, the app blows up with this stack trace:

```
   File "/Users/paul/Library/Caches/pypoetry/virtualenvs/nofos-ztRUj9S3-py3.14/lib/python3.14/site-packages/easyaudit/models.py", line 7, in <module>
    class CRUDEvent(models.Model):
    ...<52 lines>...
            index_together = ['object_id', 'content_type', ]
  File "/Users/paul/Library/Caches/pypoetry/virtualenvs/nofos-ztRUj9S3-py3.14/lib/python3.14/site-packages/django/db/models/base.py", line 145, in __new__
    new_class.add_to_class("_meta", Options(meta, app_label))
    ~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/paul/Library/Caches/pypoetry/virtualenvs/nofos-ztRUj9S3-py3.14/lib/python3.14/site-packages/django/db/models/base.py", line 373, in add_to_class
    value.contribute_to_class(cls, name)
    ~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^
  File "/Users/paul/Library/Caches/pypoetry/virtualenvs/nofos-ztRUj9S3-py3.14/lib/python3.14/site-packages/django/db/models/options.py", line 227, in contribute_to_class
    raise TypeError(
        "'class Meta' got invalid attribute(s): %s" % ",".join(meta_attrs)
    )
TypeError: 'class Meta' got invalid attribute(s): index_together
```

This comes from the [django-easy-audit](https://github.com/soynatan/django-easy-audit) library, and there is an open issue which discusses this problem: https://github.com/soynatan/django-easy-audit/issues/334

There is also a PR that, if merged, will resolve this: https://github.com/soynatan/django-easy-audit/pull/333

For now, we are just waiting for that PR to make it into the codebase.

### Upgrade markdown-related libraries 

There are 2 markdown-related libraries we heavily rely on:

- `markdownify` ([changelog](https://github.com/matthewwithanm/python-markdownify/releases))
- `markdown` ([changelog](https://python-markdown.github.io/changelog/))

In the past, upgrading these libraries without checking them has resulted in breaking changes to the app, so that is why these versions are pinned.

Fortunately, this time they are very minor changes that don't appear to have any impact on our document conversions. 

I tested a before/after using a the HTML output of a converted NOFO and it is identical other than the document ID. 

This moves the ball forward on #566 — the last piece of the puzzle is updating Django.


pcraig3:
Merging this because it is a maintenance-y update and tests are green!

PR #610, opened 2026-01-08
Fix "opdiv" error during NOFO import

pcraig3:
## Summary

This PR makes 2 changes that improve the "opdiv" importing when creating a NOFO:

1. Make error message actually display to the user
2. Allow for leading whitespace before the "Opdiv:" keyword

### 1. Make error message actually display to the user

Previously, the exception triggered when trying to import a document with no Opdiv would pull up our generic 500-error page. Now, it gives you an actionable error message. 

| before | after |
|--------|-------|
|   generic error, does not describe the problem     |  specific error about not having an opdiv     |
|   Screenshot 2026-01-08 at 15 44 41     |   Screenshot 2026-01-08 at 15 44 59   |

### 2. Allow for leading whitespace before the "Opdiv:" keyword

So previously, this would work:

- `<p>Opdiv: Centers for Disease Control and Prevention</p>`

But this would _not_ work:

- `<p> Opdiv: Centers for Disease Control and Prevention</p>`

This second one has a space between the paragraph and the "Opdiv:" string, and that would cause imports to fail.

Since we don't really care if there is a space there, I have updated the logic to allow for optional whitespace.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,10 @@ Versioning since version 1.0.0.
 
 ### Fixed
 
+- Show validation message for "no opdiv" import error
+- Allow whitespace before "Opdiv:" keyword when looking for Opdiv
+  - This works for all "suggest\_\*" functions
+
 ## [3.24.0] - 2026-01-06
 
 ### Added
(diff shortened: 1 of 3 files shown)


pcraig3:
Gonna merge this because Yasmine ran into this recently.

PR #614, opened 2026-01-09
NOFO Composer UI fixes from Annie's review

### LINKED ISSUE #590: Open accordion for default state on Composer writer home page
## Description 

Users during Composer research have consistency suggested to have the resources on the home page visible, not hidden in the accordion, which is closed as its default state. 

**Where** https://nofos.dev.simpler.grants.gov/composer/writer

| &nbsp; |
| :--: |
|  |

## AC

- [ ] When user visits ../composer/writer page, the default state for the resources accordion is open
- [ ] When returning to ../composer/writer page, the accordion remains open, unless user closes it on their previous interaction

pcraig3:
## Summary

This PR makes 5 very small fixes to the NOFO Builder + Composer:

1. Make the "small" font size smaller for the right-column callout boxes on the Basic Information page
2. Open the "getting started" accordion by default on Composer index pages
3. Make sure that composer text labels don't break across lines
4. Call it a section, not a subsection
5. Get subsection name for unnamed subsections in "Not started" alert box

#### 1. Make the "small" font size smaller for the right-column callout boxes on the Basic Information page

We actually have a `--smaller` and a `--tiny` version of the right-column callout boxes. The `--tiny` version went in in this PR: https://github.com/HHS/simpler-grants-pdf-builder/pull/608

However, we found a NOFO where the `--smaller` text size is still going on to 2 pages, which means we need to be more conservative with our font sizing. 

Moved from `9.5pt` font to `9.25pt` fonts for `--smaller` right col callout boxes.

#### 2. Open the "getting started" accordion by default on Composer index pages

We noticed that people just fly past this during testing, so let's have it open by default.

This PR closes #590 

#### 3. Make sure that composer text labels don't break across lines

| before | after |
|--------|-------|
|  "not started" label breaks across lines      |  no longer breaks (ignore the devtools stuff)      |
|     Screenshot 2026-01-09 at 15 36 34   |      Screenshot 2026-01-09 at 15 36 22 |

#### 4. Call it a section, not a subsection

This is something that Annie picked up on. In the UI, we only ever refer to subsections as "sections". I didn't catch this when it went in.

| before | after |
|--------|-------|
|   Screenshot 2026-01-09 at 16 05 13     |  Screenshot 2026-01-09 at 16 05 44     |

#### 5. Get subsection name for unnamed subsections in "Not started" alert box

We previously introduced a filter for coming up with a subsection name for nameless subsections, but we weren't using it to generate the list of subsections in the alert box for writers. Now we are!

Here is the past PR for reference: https://github.com/HHS/simpler-grants-pdf-builder/pull/574

| before | after |
|--------|-------|
|  Screenshot 2026-01-09 at 15 59 23      |   Screenshot 2026-01-09 at 15 59 04    |

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,12 +14,16 @@ Versioning since version 1.0.0.
 ### Changed
 
 - NOFO numbers that include "RFA-" will default to the CDC theme
+- Default "getting started" accordion to open on Composer index pages
 
 ### Fixed
 
 - Show validation message for "no opdiv" import error
 - Allow whitespace before "Opdiv:" keyword when looking for Opdiv
   - This works for all "suggest\_\*" functions
+- Show names of untitled subsections in the "not started" alert for composer writers
+- No more line-breaks on the composer subsection accordion tags
+- Make "small" font size smaller for right-col callout boxes
 
 ## [3.24.0] - 2026-01-06
 
--- a/nofos/composer/templates/includes/_writer_not_started_alert.html
+++ b/nofos/composer/templates/includes/_writer_not_started_alert.html
@@ -1,3 +1,5 @@
+{% load subsection_name_or_placeholder %}
+
 <section
   class="usa-alert usa-alert--warning usa-alert--not-started usa-alert--heading-sm"
   aria-label="You have not started some sections."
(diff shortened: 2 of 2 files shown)


pcraig3:
Merging because I showed this to Annie + tests have gone green.

PR #617, opened 2026-01-13
615: Add history compare diff view for nofo builder history, only for subsection change events with change to 'body'

### LINKED ISSUE #615: Add a subsection diff page for the NOFO Builder audit trail
The NOFO Builder has the most usage and is already well set up for the same kind of diff page that we already see being used on the NOFO Composer since https://github.com/HHS/simpler-grants-pdf-builder/pull/602

We should add this same diffing interface to the NOFO Builder so that the changes to various subsections are also easy to find and understand.

This should be a fairly trivial implementation, as subsections in the NOFO Builder only have a ‘body’ and don’t have variables or anything else weird like that.

However we want to be careful and not include all the changes here.

I think what we want to do is check if `body` is in `changed_fields`, and, if so, include a link to the diff page. On the diff page, the only thing that we care about is the body changing, not the title or the “page break” status or anything like that.

We want this to be small and easy to implement, and “body” diffs gives us the biggest bang for our buck.

One final note is that whitespace-only updates _will_ show up here (like 2 trailing line breaks which are deleted), and we should just include those as normal, even if the diff doesn’t show anything. Let's keep the logic simple: if “body” exists at all, create a link and generate a diff page. At this stage, don’t try to figure out what kind of diff it is.

jlhogantech:
closes #615 

- Creates an all-purpose version of the history compare 'diff' view
- adds `changed_fields` to formatted event
- adds a step to NofoHistoryView that augments each event with a boolean flag `should_link_to_diff` if the event in the table should be linked to the new diff view
- creates a NofoHistoryCompareView for populating the old and new diff for the subsection body

### NOFO history view with link-able subsection change
Screenshot 2026-01-12 192731

### Diff page for NOFO subsection change
Screenshot 2026-01-12 192747

### DIFF
--- a/nofos/bloom_nofos/html_diff.py
+++ b/nofos/bloom_nofos/html_diff.py
@@ -118,9 +118,21 @@ def diff_leaf(tag1, tag2, dmp):
 
 def wrap_tag(tag, wrapper):
     if tag.name in BLOCK_WRAPPER_UNSAFE_TAGS:
-        return (
-            f"<{tag.name}><{wrapper}>{tag.decode_contents()}</{wrapper}></{tag.name}>"
-        )
+        # Special handling for <tr> - wrap each child cell individually
+        # because <tr><ins>...</ins></tr> is invalid HTML and may get reformatted by the browser
+        if tag.name == "tr":
+            wrapped_children = []
+            for child in tag.children:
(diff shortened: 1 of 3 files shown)


pcraig3:
Looks good, gonna merge.

pcraig3 (approved):
Thanks Julia, this looks great!

PR #618, opened 2026-01-13
Look for 'insert' in subsection body and instructions to determine if subsection is edit_mode = 'full'

### LINKED ISSUE #616: Automatically tag imported content guide subsections as “full”
After doing a scan through the CDC content guide, the magic word we are looking for is “insert”.

The word "insert" appears in 3 contexts:

1. In yellow or green text for an entire section that should have “full” editing (eg, Summary)
2. In yellow highlighted text for variables (eg, Key Facts)
3. In the instructions boxes (eg, Cost sharing (YES))

|  "Insert" in full editing content | "Insert" in variable editing content | "Insert" in instructions |
|--------|-------|-------|
|   Subsection: Summary<br>(This is the kind of subsection we are interested in for this ticket)     |   Subsection: Key facts    |   Subsection instructions: Cost sharing (YES)   |
|   Image     |  Image     |   Image   |

"Insert" never appears in the body of the text in any official/legal NOFO language; in our Content Guide it is _always_ meant as an instruction for writers.

So the algorithm should work like this:

- First, run our existing check for variables
- After, look for subsections that are not “variables” and check for the word “insert” (case-insensitive) inside of `subsection.body`. 
- If it is there, then make this a “full” editable subsection.

Note that the “optional” check is not affected by this new logic: it doesn’t affect the `edit_mode`, so we can do that check at any point. 

Also, I have added a few more curly braces so that we make sure variable sections are truly variable sections, so I will send you a new version of the content guide to build this out with.

jlhogantech:
closes #616 

Updates content guide ingestion to check for the presence of the word 'insert' in the subsections body and instructions, and uses its presence to flag a subsection as edit_mode = "full" if there are not any variables

### DIFF
--- a/nofos/nofos/nofo.py
+++ b/nofos/nofos/nofo.py
@@ -433,8 +433,10 @@ def _build_document(document, sections, SectionModel, SubsectionModel):
                         key: var.to_dict() for key, var in variables.items()
                     }
                     subsection_obj.variables = json.dumps(serializable_variables)
-                # If no variables, and subsection is optional or the word "insert" is present in the
-                # subsection body or instructions, then edit_mode = "full"
+                # Subsection edit_mode = "full" if no variables, and:
+                #   - subsection is optional OR
+                #   - the word "insert" is present in the subsection body OR
+                #   - the word "insert" is present in the instructions
                 elif (
                     subsection_obj.optional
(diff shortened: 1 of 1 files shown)


pcraig3:
Gonna merge!

PR #619, opened 2026-01-14
Add ADR for html to docx solution

jlhogantech:


pcraig3:
Looks good, approved!

PR #679, opened 2026-05-08
Remove curl from the final Dockerfile

pcraig3:
## Summary

This is a very small PR that removes the `curl` package from the Dockerfile 

We use curl to download poetry when we first create the Dockerfile, but then it is not used again or relevant to the application later.

It is also causing a vulnerability alert in our container scan so if we don't need it, we should pitch it out.


pcraig3:
Deployed and it works!

PR #693, opened 2026-05-20
Add Opdiv Admin users role

pcraig3:
## Summary

This PR adds in a new user role called "OpDiv Admin". An Opdiv Admin user can manage the users for their team. 

This means OpDiv Admins can:

  - See all users in their group
  - Add a new user to their group
  - Delete a user from their group
  - Edit users (including reset passwords) for users in their group

OpDiv Admins _can’t_:

- Edit the "group" for another user
- Edit "is_superuser" for another user
- See who is and isn't a superadmin
  
Note that OpDiv admins can add/delete other OpDiv admins, so you could have a nefarious user delete all the other users in their group. However, they would be scoped to their group, so a superuser can come in and remove them later on. This is not a thing we are trying to solve for because we don't think it will happen.

## Screenshots 

| screen | screenshot |
|--------|------------|
|   Profile page     |      Screenshot 2026-05-19 at 10 55 39   |
|   Team overview     |      Screenshot 2026-05-19 at 10 55 48   |
|   Delete a user page   |     Screenshot 2026-05-19 at 10 56 19    |


benbrasso:
@pcraig3 the description looks great. Does this work include or assume that an OpDiv Admin can view/edit all NOFOs in their Group? And, is the behavior for a standard user to view all NOFOs in their group? 

If so, I can see where we might want to have OpDiv group values that are actually program offices. E.g., CDC DGHT, CDC DGHP. This seems "in scope" based on what the functionality offers. 

And then, I wonder how feasible it would be to allow OpDiv Admin users to be assigned to 1 or more OpDiv groups, e.g., CDC DGHT and CDC DGHP. This I might consider a nice-to-have, but the rest seems feasible based on what you've described above.

pcraig3:
Hi @benbrasso-agile6, sorry for the delay — I broke a lot of tests by accident. 

>  Does this work include or assume that an OpDiv Admin can view/edit all NOFOs in their Group?

All users in a `group` are able to see all NOFOs for that group. So anyone in HRSA can see all HRSA NOFOs. It doesn't matter if they are an "Opdiv Admin" or not. The only exception to this are `bloom` users, who can see all NOFOs for all OpDivs (called `groups`). 

When I add a new NIH OpDiv, it will be a new `group`. 

> And, is the behavior for a standard user to view all NOFOs in their group?

Yes, same as OpDiv Admin. All users can see all NOFOs that belong to their group.

> If so, I can see where we might want to have OpDiv group values that are actually program offices. E.g., CDC DGHT, CDC DGHP. This seems "in scope" based on what the functionality offers.

We could do this by creating groups that are smaller than the entire OpDiv. Here are our `group` names: https://github.com/HHS/simpler-grants-pdf-builder/blob/main/nofos/bloom_nofos/settings.py#L598

There is one in there for CDC, but we can just add new ones whenever we want. However, there is no parent-child relationship for groups. If you are part of CDC DGHP, that's a different group than CDC, no overlap.

> And then, I wonder how feasible it would be to allow OpDiv Admin users to be assigned to 1 or more OpDiv groups, e.g., CDC DGHT and CDC DGHP. This I might consider a nice-to-have, but the rest seems feasible based on what you've described above.

This would be a large task to do. We would either need users to be allowed to be assigned more than 1 "group" (complicated), or add in a parent-child hierarchy to groups (complicated). So if it's not a hard requirement, I would skip it. Of course, the one group that can see all users and all NOFOs is Bloom, but it's an all-or-nothing permission, it's not granular at all.

benbrasso:
@pcraig3 those responses are great. Thank you!

pcraig3:
Okay! Going to merge

benbrasso:
Thanks; I created a [decisions doc](https://github.com/HHS/simpler-grants-pdf-builder/blob/main/documentation/DECISIONS.md) to keep track of some of our convos and decisions!

PR #743, opened 2026-07-20
Stack multi-block checkbox table content

### LINKED ISSUE #742: Checkbox table cells with multiple content blocks render side by side
## Summary

When a table cell begins with a checkbox and contains more than one block of content, NOFO Builder can lay the blocks out side by side instead of stacking them vertically.

The same multi-block content renders correctly when the cell does not contain a checkbox.

## Steps to reproduce

1. Create a Word table with a cell that starts with a checkbox character.
2. Add either:
   - a second paragraph below the checkbox label, or
   - a paragraph and a bulleted list below the checkbox label.
3. Import the document into NOFO Builder.
4. View the imported table.

## Current behavior

Builder recognizes the checkbox and wraps the cell content in a checkbox-specific container. Each paragraph or list becomes a separate child of that container. The container uses a flex layout, so the blocks can display horizontally.

Examples reproduced through the local import pipeline:

- Checkbox plus two paragraphs becomes two sibling `<p>` elements inside the checkbox wrapper.
- Checkbox plus supporting text and a list becomes sibling `<p>`, `<p>`, and `<ul>` elements inside the wrapper.
- The equivalent multi-block cell without a checkbox retains normal vertical flow.

## Expected behavior

The checkbox and its label should remain aligned, while any later paragraphs or lists should appear beneath the label in their original order.

Import should preserve the document's block structure without requiring authors to flatten valid Word table content into one paragraph.

## Verification

This was reproduced using a DOCX fixture derived from an existing OCS NOFO. All fixture tables survived parsing, cleanup, section creation, Markdown conversion, and the same template filters used by the Builder preview.

The issue is specific to checkbox-cell rendering. Column-width handling is a separate concern and is not part of this bug.

mikec-ai:
## Summary

Fixes #742.

NOFO Builder previously applied a flex layout to every top-level block inside checkbox table cells. That worked for a checkbox with a short label, but cells containing later paragraphs or lists could render those blocks side by side.

This change keeps the checkbox and its label aligned while allowing later content to flow vertically.

## What changed

- Detect checkbox cells containing multiple content blocks.
- Apply the existing flex alignment only to each checkbox line.
- Preserve current behavior for simple and linked checkbox cells.
- Handle mixed icon cells and repeated checkbox lines.
- Add focused regression coverage for simple, linked, nested, multi-paragraph, list, mixed-icon, repeated-checkbox, and non-checkbox cases.

## User impact

Authors can use valid multi-paragraph or list content in Word table cells containing checkboxes without having to rewrite the table as HTML or flatten the content into one paragraph.

## Validation

- 163 template-tag tests pass.
- Pre-commit formatting and repository checks pass.
- The DOCX table fixture passed through the real import and render filters.
- A computed browser layout check confirmed simple rows remain aligned and later blocks stack vertically.
- An independent skeptical review identified a mixed-icon edge case, which is covered by the final implementation and tests.

### DIFF
--- a/nofos/bloom_nofos/static/theme-base.css
+++ b/nofos/bloom_nofos/static/theme-base.css
@@ -337,6 +337,19 @@ body.cms-white table td img.usa-icon--check_box_outline_blank {
   display: flex;
 }
 
+.section--content table tr td.usa-icon__td--multi-block > div {
+  display: block;
+}
+
+.section--content
+  table
+  tr
+  td.usa-icon__td--multi-block
+  > div
+  > .usa-icon__line {
+  display: flex;
+}
+
 .section--content table tr td.usa-icon__td {
   vertical-align: top;
 }
--- a/nofos/nofos/templatetags/replace_unicode_with_icon.py
+++ b/nofos/nofos/templatetags/replace_unicode_with_icon.py
@@ -22,6 +22,17 @@ ICONS = [
     ("◻", uswds_check_box_outline_blank_icon),  # (◻) U+25FB WHITE MEDIUM SQUARE
 ]
 
+BLOCK_LEVEL_TAG_NAMES = {
+    "blockquote",
+    "div",
+    "dl",
+    "ol",
+    "p",
+    "pre",
+    "table",
+    "ul",
+}
+
 
 def has_link_in_above_rows(td):
     # Find the parent row of the cell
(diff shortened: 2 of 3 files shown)


benbrasso (approved):
To verify, I need to deploy this to dev. I'm going to merge and deploy to dev.

benbrasso:
@mikec-ai ready for me to deploy to `prod`? 

## Word document

Screenshot 2026-07-21 at 2 28 27 PM

## Prod

Screenshot 2026-07-21 at 2 27 01 PM

## Dev (with this branch) 

Screenshot 2026-07-21 at 2 28 13 PM

mikec-ai:
@benbrasso yep, good to go!

benbrasso:
deployed to `prod`

PR #748, opened 2026-07-21
Standardize blocking import errors

### LINKED ISSUE #746: Standardize full-page import errors and recovery guidance
## Current behavior

NOFO Builder uses several different full-page patterns when an import cannot continue:

- a generic 400 page for many validation failures
- a special-case blank OpDiv page
- strict-mode conversion failures shown through the generic 400 template
- unexpected import exceptions logged as 500 but returned as 400 responses
- separate 403, 404, and 500 pages with different navigation and support guidance

These pages do not consistently explain what failed, how to repair it, or how to return to the same import workflow. Some responses can expose implementation details such as Django validation dictionaries, Mammoth terminology, style maps, raw Style IDs, or exception text. Support destinations also differ between pages.

## Intended behavior

When an import must stop, Builder should present one consistent, application-styled blocking-error pattern that:

- names the task that failed
- explains the problem and its impact in plain language
- identifies the affected content when possible
- provides ordered recovery steps
- returns the user directly to the originating import or re-import workflow
- uses one confirmed support destination
- retains the correct HTTP status without exposing internal details

Known input and conversion failures should use appropriate 400 or 422 responses. Unexpected exceptions should return a sanitized 500 page and keep technical details in server logs.

## Initial scope

- Create a reusable full-page blocking-import error component or template interface.
- Migrate the existing blank OpDiv error without losing its useful guidance.
- Replace technical strict-mode/import conversion wording with user-facing language.
- Correct unexpected import exception handling so it returns a true, sanitized 500 response.
- Centralize retry and support links.
- Make the component available to issue #745 so ambiguous mixed headings can stop before a partial document is created.

## Out of scope

- Redesigning inline form validation.
- Changing the existing "Before publishing" warning area.
- Auditing every edit form, transient alert, or destructive action.
- Combining all HTTP statuses into the same generic message.
- Retaining uploaded Word files between attempts.

## Acceptance criteria

- Known blocking import failures use the shared pattern across new import and re-import flows where applicable.
- Each page says what failed, why it matters, what the user should do, and how to retry the same action.
- User-facing responses do not expose exception strings, Django dictionaries, Mammoth implementation names, style maps, or raw Style IDs by default.
- Expected input failures return a consistent 400 or 422 status.
- Unexpected exceptions return HTTP 500 and a sanitized generic page.
- Retry links preserve the originating workflow instead of sending users to a generic landing page.
- All migrated pages use one confirmed support destination.
- Automated tests cover response status, copy, recovery links, and accessibility hooks.
- Existing blank OpDiv and valid import behavior remains unchanged.

mikec-ai:
## Summary

Standardize blocking import errors across NOFO import, re-import, Composer, and Compare.

Previously, these failures used several different full-page responses. Some exposed converter or validation details, some returned HTTP 400 for unexpected server failures, and recovery and support guidance varied by workflow.

Now, blocking import failures use one plain-language pattern with:

- context-specific explanations and recovery steps
- retry links that return to the originating workflow
- stable support codes that are also recorded in logs
- the NOFO Builder Feedback Form and simplerNOFOs@agile6.com
- sanitized HTTP 500 responses for unexpected failures
- sanitized HTTP 422 responses for Word conversion and strict-formatting failures

The existing inline form validation and pre-publishing checks are unchanged.

### Recoverable import error

Recoverable blank OpDiv import error

### Unexpected import error

Sanitized unexpected import error

## Validation

- 25 focused import, Composer, and Compare tests passed
- full suite: 1,491 tests passed
- Black and isort checks passed
- skeptical subagent review completed with no remaining findings
- visually verified the recoverable and unexpected error pages locally

Closes #746

### DIFF
--- a/nofos/bloom_nofos/error_helpers.py
+++ b/nofos/bloom_nofos/error_helpers.py
@@ -1,0 +1,46 @@
+from django.shortcuts import render
+
+
+def render_blocking_import_error(
+    request,
+    *,
+    title,
+    summary,
+    error_code,
+    status=400,
+    recovery_steps=None,
+    retry_url=None,
+    retry_label="Try the import again",
+):
+    """Render a safe, actionable error page for a blocked document import."""
+    return render(
+        request,
+        "import_error.html",
+        status=status,
+        context={
+            "error_title": title,
+            "error_summary": summary,
+            "error_code": error_code,
+            "recovery_steps": recovery_steps or [],
+            "retry_url": retry_url,
+            "retry_label": retry_label,
+        },
+    )
+
+
+def render_import_server_error(request, *, retry_url=None):
+    """Return a sanitized 500 response for an unexpected import failure."""
+    return render_blocking_import_error(
+        request,
+        title="We couldn’t finish importing this document",
+        summary=(
+            "Something went wrong in NOFO Builder. The document was not imported."
+        ),
+        error_code="IMPORT-UNEXPECTED",
+        status=500,
+        recovery_steps=[
+            "Try the import again.",
+            "If the problem continues, use the help options below and include the error code.",
+        ],
+        retry_url=retry_url,
+    )
--- a/nofos/bloom_nofos/error_helpers.py
+++ b/nofos/bloom_nofos/error_helpers.py
@@ -1,5 +1,10 @@
 from django.shortcuts import render
 
+DOCUMENT_STRUCTURE_RECOVERY_STEPS = (
+    "Review the document’s required metadata and heading structure.",
+    "Save the document, then select it again.",
+)
+
 
 def render_blocking_import_error(
     request,
--- a/nofos/bloom_nofos/templates/400.html
+++ b/nofos/bloom_nofos/templates/400.html
@@ -7,35 +7,22 @@
 
 
 {% block content %}
-	{% if opdiv_blank_error %}
-		<h1 class="font-heading-xl margin-y-0">We couldn’t import this NOFO</h1>
-		<p>The <strong>‘Opdiv:’</strong> field on page 1 of the Word document is blank. NOFO Builder needs this field filled in before it can import the document.</p>
-		<p>Steps to fix:</p>
-		<ol class="usa-list">
-			<li>Open the Word document.</li>
-			<li>Add a value after ‘Opdiv:’ — this is the agency’s operating division (for example, ‘Administration for Children and Families’ or ‘CDC’).</li>
-			<li>Save the document.</li>
-			<li><a href="{% url 'index' %}">Import it again.</a></li>
-		</ol>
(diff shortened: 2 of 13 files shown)


benbrasso (commented):
Nice work on this one — the core fix (sanitizing internal errors, correct status codes, consistent recovery UX) looks solid and well-tested. A few things worth a look before/after merge:

## ⚠️ Blank-Opdiv error page conflicts with #732/#733/#734

Heads up — the blank-Opdiv NOFO import error page was just reworked over the last few days in #732 and its follow-ups (#733, #734). Since this PR moves that same path (`nofos/views.py`, the `opdiv_blank_error` branch of `400.html`) onto the new `render_blocking_import_error()` helper and shared `import_error.html` template, can you confirm the following survived the migration to `IMPORT-OPDIV-BLANK`?

- The `error_dict` + `"blank"` error-code detection logic (this replaced an exact-string match per Copilot review feedback, so it's robust to localization and multiple validators firing on the same field)
- The refined "Steps to fix" copy and the link back to the homepage
- The Feedback Form link opening in a new tab (`target="_blank" rel="noopener noreferrer"`)

Want to make sure we're not quietly losing that work in the rewrite.

## Composer

Composer isn't in active use right now, so we don't need to manually QA that error path as part of this round — but no need to strip the code out either, since it's just riding on the same shared helper as the other three workflows and is low-risk to leave in place for whenever Composer goes live.

## Suggestion: shared constant for the recovery-steps copy

The two-line recovery list ("Review the document's required metadata..." / "Save the document, then select it again.") is duplicated identically across five call sites. Since consistent recovery copy was part of the point of this PR, pulling it into a shared constant in `error_helpers.py` seems worth doing now rather than letting five copies drift out of sync later. Happy to do this as a follow-up if you'd rather keep this PR as-is.

## Question: is the Mammoth string-sniffing fallback still needed?

In `nofos/views.py`, this check still has a fallback:

```python
if "strict_formatting" in error_codes or "Mammoth" in error_message:
```

Since `nofo.py` now always sets `code="strict_formatting"` explicitly on that raise, and no other call site seems to raise a "Mammoth"-containing message, is the `"Mammoth" in error_message` fallback guarding some other caller, or is it leftover from before the code-based check was added? If it's not needed, removing it would avoid it silently masking a future regression in the code-based check.

None of these are blockers — just want to flag them before/while this lands.

mikec-ai:
Thanks, Ben. I made the two follow-up changes in commit 683e807d:

- moved the repeated document-structure recovery steps into a shared constant
- removed the Mammoth message-string fallback, since strict-formatting errors now carry the explicit `strict_formatting` code

I also checked the blank-OpDiv path against #732, #733, and #734:

- the `error_dict` plus `blank` code detection is preserved
- the recovery guidance is preserved
- the Feedback Form still opens in a new tab with `noopener noreferrer`

One intentional difference is the retry destination. The old page linked to the homepage. The consolidated error flow returns directly to the import form, which keeps the action tied to the workflow the user was in.

Validation: Black and isort pass, and all 196 focused NOFO, Composer, and Compare import tests pass.

PR #762, opened 2026-07-24
Block ambiguous mixed-heading imports

### LINKED ISSUE #745: Mixed H1/H2 hierarchy can silently omit most document content during import
## What happened

NOFO Builder can silently leave out most of a Word document when the document mixes Heading 1 and Heading 2 for its main sections.

In the reproduced case, the document had:

- seven main sections formatted as Heading 2
- one appendix near the end formatted as Heading 1

Because Builder found a Heading 1 anywhere in the document, it treated Heading 1 as the section level for the entire import. The imported NOFO contained only the appendix. All content before that Heading 1 was omitted without an error.

## Why this matters

A minor heading inconsistency in Word can produce a successful-looking but incomplete import. A user may not realize that many pages of content are missing.

## Reproduction

A reduced Word fixture with the following structure reproduces the behavior:

1. Add several populated main sections using Heading 2.
2. Add a final appendix using Heading 1.
3. Import the document into Builder.
4. Observe that Builder creates only the appendix section.

Changing only the appendix from Heading 1 to Heading 2 causes all intended sections to import.

## Expected behavior

Builder should never silently complete a partial import because of an ambiguous heading hierarchy.

At minimum, Builder should stop the import and show a clear message that identifies the conflicting heading levels and headings so the user can correct the Word document.

If Builder can determine the intended section level with high confidence, it may import all intended sections instead. Ambiguous cases should still fail clearly.

## Acceptance criteria

- Mixed heading levels do not result in silent section loss.
- Ambiguous documents are blocked with an actionable validation message.
- The message identifies the conflicting heading levels and example headings.
- Tests cover:
  - Heading 1-only documents
  - Heading 2-only documents
  - ambiguous mixed Heading 1/Heading 2 documents
- Existing valid imports continue to produce the same sections.

mikec-ai:
## What changed

- Detects when a document starts its main sections with Heading 2, then introduces Heading 1 later
- Stops the import before NOFO Builder can silently omit the earlier Heading 2 content
- Shows the shared import error page with example conflicting headings and clear recovery steps
- Uses one heading resolver for both validation and section parsing
- Ignores headings used inside tables when choosing the document's main section level

## Previous behavior

NOFO Builder chose Heading 1 whenever any Heading 1 appeared in the document. If populated Heading 2 sections came first, Builder could start at the later Heading 1 and silently leave the earlier content out of the imported NOFO.

## Intended behavior

When the document structure is ambiguous, NOFO Builder now blocks the import and asks the user to apply one consistent heading level to the main sections in Word. Documents that use Heading 1 only, Heading 2 only, or Heading 1 with Heading 2 subsections continue to import normally. Headings used as table formatting do not change the main section level.

## Validation

- 1,513 automated tests pass
- 198 focused import, compare, and composer tests pass
- Added an end-to-end regression proving an H1 inside a table does not hide surrounding H2 sections
- Reproduced the original failure with the real CMS content guide fixture
- Completed a skeptical review and follow-up review with no remaining findings

Closes #745

### DIFF
--- a/nofos/nofos/nofo.py
+++ b/nofos/nofos/nofo.py
@@ -116,6 +116,51 @@ def parse_uploaded_file_as_html_string(uploaded_file):
         raise ValidationError("Please import a .docx or HTML file.")
 
 
+def resolve_section_heading_level(soup):
+    """Choose the section heading level without silently discarding earlier content."""
+    section_heading_candidates = [
+        heading
+        for heading in soup.find_all(["h1", "h2"])
+        if clean_string(heading.get_text(" ", strip=True))
+        and heading.find_parent("table") is None
+    ]
+
+    first_h1_index = next(
+        (
+            index
+            for index, heading in enumerate(section_heading_candidates)
+            if heading.name == "h1"
+        ),
+        None,
+    )
+    if first_h1_index is None:
+        return "h2"
+
+    h2_before_first_h1 = next(
+        (
+            heading
+            for heading in section_heading_candidates[:first_h1_index]
+            if heading.name == "h2"
+        ),
+        None,
+    )
+    if h2_before_first_h1 is None:
+        return "h1"
+
+    h2_text = clean_string(h2_before_first_h1.get_text(" ", strip=True))
+    h1_text = clean_string(
+        section_heading_candidates[first_h1_index].get_text(" ", strip=True)
+    )
+    raise ValidationError(
+        "The document uses Heading 2 before its first Heading 1. "
+        f'NOFO Builder would skip content beginning with Heading 2 "{h2_text}" '
+        f'and start at Heading 1 "{h1_text}". '
+        "In Word, apply the same heading level to all main sections, save the "
+        "document, and import it again.",
+        code="ambiguous_heading_hierarchy",
+    )
+
+
 def process_nofo_html(soup, top_heading_level):
     """
     Takes a soup object, cleans it up and mutates it, and returns a modified soup object.
--- a/nofos/nofos/test_nofo.py
+++ b/nofos/nofos/test_nofo.py
@@ -64,6 +64,7 @@ from .nofo import (
     replace_chars,
     replace_src_for_inline_images,
     replace_value_in_subsections,
+    resolve_section_heading_level,
     sanitize_imported_text,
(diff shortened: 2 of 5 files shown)


benbrasso:
@mikec-ai 

> NOFO Builder chose Heading 1 whenever any Heading 1 appeared in the document. If populated Heading 2 sections came first, Builder could start at the later Heading 1 and silently leave the earlier content out of the imported NOFO.

Where did you see that this caused a problem? (This is by design.)

benbrasso:
@mikec-ai 

🔴 The fix isn't applied everywhere — reimport-confirmation path still has the old bug
[nofos/nofos/views.py:864](https://github.com/HHS/simpler-grants-pdf-builder/blob/main/nofos/nofos/views.py#L864), inside NofosConfirmReimportView.post(), still does:

soup = BeautifulSoup(reimport_data["soup"], "html.parser")
top_heading_level = "h1" if soup.find("h1") else "h2"
instead of calling the new resolve_section_heading_level(soup).

benbrasso:
@mikec-ai 

Missing observability for the new error path

docx_conversion and strict_formatting both call log_exception(...) before rendering their blocking error page, but the new ambiguous_heading_hierarchy branch (views.py:556-568) doesn't. Without it, there's no structured log/metric for how often this new block fires — which matters given it's a new wall users can hit that didn't exist before (previously the import just silently dropped content instead of blocking). Worth adding for parity and so we can measure real-world impact post-launch:

if "ambiguous_heading_hierarchy" in error_codes:
    log_exception(
        request,
        e,
        level="warning",  # expected validation rejection, not a system failure
        context="BaseNofoImportView:ValidationError:IMPORT-AMBIGUOUS-HEADINGS",
        status=422,
    )
    return render_blocking_import_error(
        ...
    )
(Suggesting level="warning" rather than the sibling branches' default error, since this is an expected rejection of a parseable document rather than a processing failure — don't want it tripping error-rate alerts.)

mikec-ai:
@benbrasso shared the word doc of the failure case via slack. Your other comments have been addressed

benbrasso:
@mikec-ai good to merge; go for it 🚢

benbrasso (approved):

PR #765, opened 2026-07-24
Recover OpDiv values split across Word paragraphs

### LINKED ISSUE #763: Recognize an OpDiv value placed in the next Word paragraph
## Problem

NOFO Builder reports that `Opdiv:` is blank when the value is present but begins in the next Word paragraph.

This occurred in a real GrantSolutions Announcement Module export. The user could not import the NOFO until an expert moved the value onto the same line as the label.

## Independently reproduced

I created four controlled Word documents with identical metadata and heading structure. Only the OpDiv layout changed.

| Word structure | Result |
| --- | --- |
| Label and value in the same paragraph | Import succeeds |
| Label and value separated by a soft line break, such as Shift+Enter | Import succeeds |
| `Opdiv:` paragraph followed by a separate value paragraph | Import fails with `IMPORT-OPDIV-BLANK` |
| `Opdiv:` with no value | Import fails with `IMPORT-OPDIV-BLANK` |

The full local import endpoint reproduced these results. The paragraph-break case returned HTTP 400 and the same error shown for a genuinely blank value.

## Why it happens

A soft line break converts to one HTML paragraph:

```html
<p>Opdiv:<br/>Administration for Children and Families</p>
```

A Word paragraph break converts to two paragraphs:

```html
<p>Opdiv:</p>
<p>Administration for Children and Families</p>
```

The current parser reads only the paragraph containing `Opdiv:`. It does not consider the following paragraph, so model validation receives an empty value.

## Intended behavior

Builder should recognize an unambiguous OpDiv value in the immediately following paragraph. It should continue blocking the import when the value is genuinely missing or the nearby content is ambiguous.

## Acceptance criteria

- [ ] Import recognizes a non-empty OpDiv value in the immediately following paragraph.
- [ ] Same-paragraph and soft-line-break imports continue to work.
- [ ] A genuinely blank OpDiv continues to block import.
- [ ] Another metadata label is not captured as the OpDiv value.
- [ ] A heading, table, or unrelated block is not captured as the OpDiv value.
- [ ] Ambiguous cases receive a specific recovery message rather than being silently guessed.
- [ ] Regression tests cover the accepted, blank, and ambiguous layouts.

## Test fixtures

Four small, non-sensitive Word fixtures were created locally and can be added to the repository when implementation begins.

mikec-ai:
## Summary

- Read an OpDiv value from the next Word paragraph when the surrounding metadata makes that layout clear.
- Avoid guessing when the following text could be a heading, another field, a placeholder, or unrelated content.
- Show clearer recovery instructions when Builder cannot confidently read the value.
- Add real Word fixtures for soft line breaks and paragraph breaks.

## Previous behavior

When `Opdiv:` and its value appeared on separate Word paragraphs, NOFO Builder could treat the field as blank and stop the import even though the value was visible in the document.

## Intended behavior

NOFO Builder now preserves and imports the value when the split layout is unambiguous. If the layout is uncertain, it stops safely and asks the user to put the OpDiv value on the same line as the label instead of silently importing unrelated text.

## Validation

- 34 focused OpDiv and import tests
- Full test suite: 1,539 tests
- Multiple adversarial review rounds covering metadata labels, placeholders, tables, headings, and unrelated text
- Word fixture rendered and visually inspected

Closes #763


benbrasso:
Works on local

PR #770, opened 2026-07-28
Treat imported metadata placeholders as missing

### LINKED ISSUE #756: Detect placeholder text in metadata fields (Author, Subject, Keywords) and warn user instead of treating them as complete
## Scenario

NOFO metadata fields — Author, Subject, and Keywords — are pulled in at import time and later written into the corresponding PDF metadata fields when the NOFO is rendered/generated as a PDF.

Currently, NOFO Builder shows a warning to the user when one of these fields is empty, prompting them to fill it in before the NOFO is considered complete.

However, some source Word documents contain placeholder text in these fields instead of a real value or an empty field — for example, curly-brace placeholder text like `{Leave blank. Coach will insert.}` shown in the Subject and Keywords fields. Because these fields aren't technically empty, the existing "field is empty" warning doesn't fire, and NOFO Builder treats the NOFO as complete even though the metadata still contains placeholder text rather than a real value. If left as-is, this placeholder text would get written directly into the PDF's metadata, which we don't want.

## What needs to be done

Investigate where NOFO Builder currently validates the Author, Subject, and Keywords metadata fields (the same logic that triggers the "field is empty" warning) and extend it to also detect placeholder text patterns, such as:
- Text wrapped in curly braces (e.g., `{Leave blank. Coach will insert.}` or similar variants)
- Any other known placeholder conventions used in NOFO source documents for these fields

**Potential solution to investigate:** rather than detecting and stripping placeholder text after it's already been imported into the NOFO, consider handling this at import time. As NOFO Builder parses the Word doc, specifically for these three fields (Author, Subject, Keywords), search for curly-brace-wrapped content and strip/remove that text during import, leaving the field empty. This would let the existing "field is empty" warning fire naturally, rather than needing new logic to detect and warn on placeholder text already living in the field. This may be a more reliable approach than trying to pattern-match placeholder text after the fact, since it addresses the issue at the source.

This should start as a read-only investigation into the current import parsing and validation logic for these three fields, followed by a scoped fix once the approach is confirmed.

## Acceptance criteria

- [ ] During import, if the Author, Subject, or Keywords field in the source Word doc contains curly-brace-wrapped placeholder text (e.g., `{Leave blank. Coach will insert.}`), that text is stripped out and the field is left empty rather than imported as-is.
- [ ] A field left empty after this stripping triggers the existing "field is empty" warning, so the user is still prompted to fill it in.
- [ ] A NOFO with placeholder text originally in any of these three fields is not treated as complete/ready to publish until a real value is entered.
- [ ] The existing empty-field warning behavior is unchanged for fields that were already empty in the source doc (no regression).
- [ ] Fields containing real, non-placeholder values continue to import and pass validation with no warning.
- [ ] Tests cover: (1) empty field in source doc (existing behavior), (2) field with curly-brace placeholder text in source doc (stripped at import, warning shown), (3) field with a real value (imports normally, no warning).

mikec-ai:
## Summary

- normalize whole-field curly-brace placeholders in PDF Author, Subject, and Keywords to empty values during import
- use the same metadata policy when determining whether existing records need the incomplete-metadata warning
- preserve real metadata values, including values that contain embedded braces

## Why

Source Word documents can contain instructions such as
`{Leave blank. Coach will insert.}` in PDF metadata fields. Because these
strings were non-empty, NOFO Builder treated the fields as complete and could
carry the instructions into generated PDF metadata.

The investigation found that metadata extraction and completeness checking
used separate notions of “missing.” This change introduces one small shared
policy and applies it at both boundaries, without broadening the refactor of
`nofo.py`.

## User impact

New imports and reimports leave placeholder-only metadata empty, allowing the
existing warning to prompt users for real values. Existing NOFOs that already
contain placeholders now receive the same warning.

## Validation

- 26 focused metadata policy, import extraction, and warning tests
- full `nofos` Django test suite
- formatting and repository pre-commit checks
- independent subagent review found no actionable issues

## Screenshots

**Before**
before-main-6927a6fd

**After**
after-d6e54280

Closes #756

### DIFF
--- a/nofos/nofos/nofo.py
+++ b/nofos/nofos/nofo.py
@@ -32,6 +32,7 @@ from .import_transforms import (
 )
 from .models import Nofo, Section, Subsection
 from .nofo_markdown import PRESERVE_BOOKMARK_TARGET_ATTR, md
+from .pdf_metadata import normalize_pdf_metadata_value
 from .utils import (
     add_html_id_to_subsection,
     clean_string,
@@ -1779,17 +1780,17 @@ def suggest_nofo_tagline(soup):
 
 def suggest_nofo_author(soup):
     suggestion = _suggest_by_startswith_string(soup, "Metadata Author:")
-    return suggestion or ""
+    return normalize_pdf_metadata_value(suggestion)
 
 
 def suggest_nofo_subject(soup):
     suggestion = _suggest_by_startswith_string(soup, "Metadata Subject:")
-    return suggestion or ""
+    return normalize_pdf_metadata_value(suggestion)
 
 
 def suggest_nofo_keywords(soup):
     suggestion = _suggest_by_startswith_string(soup, "Metadata Keywords:")
-    return suggestion or ""
+    return normalize_pdf_metadata_value(suggestion)
 
 
 def suggest_nofo_cover_image(nofo):
--- a/nofos/nofos/pdf_metadata.py
+++ b/nofos/nofos/pdf_metadata.py
@@ -1,0 +1,26 @@
+import re
+
+PDF_METADATA_FIELDS = (
+    ("author", "Author"),
+    ("subject", "Subject"),
+    ("keywords", "Keywords"),
+)
+
+_WHOLE_FIELD_PLACEHOLDER = re.compile(r"^\{[^{}]+\}$")
+
+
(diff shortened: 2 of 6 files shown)


benbrasso (approved):
Approved to merge

PR #775, opened 2026-07-28
Allow editors to add or remove subsection callout styling

### LINKED ISSUE #772: Allow editors to add or remove callout styling from a subsection
## Problem

Editors can change a subsection's content, heading, and page-break behavior in NOFO Builder, but they cannot add or remove callout-box styling. The subsection already stores this as an editable state; the current edit screen silently resubmits the existing value instead of exposing a choice.

This means an imported subsection that has the wrong presentation cannot be corrected in Builder without engineering support or changing the source document and reimporting it.

## User evidence

Expert user feedback identified the ability to add or remove callout styling as a useful editing capability. Ben recommended prioritizing it because the need is concrete and the existing implementation makes it feasible.

## Current behavior confirmed in code

- Subsections already store whether they are callout boxes.
- The subsection form already accepts that value.
- Existing HTML and Word rendering support named and unnamed callouts.
- The edit template preserves the current value with a hidden field, so editors cannot change it.
- Some recognized Step 1 callouts use special floating placement in portrait layouts.
- H3 callouts remain part of the table of contents.
- Reimport does not currently preserve callout state in the same way it preserves page-break classes.

## Desired outcome

An editor can intentionally add or remove callout styling from a subsection without losing or unexpectedly changing its content, heading, anchor, page-break setting, or navigation behavior.

## Acceptance criteria

- The subsection editing experience allows both transitions:
  - regular subsection to callout;
  - callout to regular subsection.
- Editing unrelated subsection fields continues to preserve the existing callout state.
- Both named and unnamed subsections are supported.
- The subsection renders exactly once in portrait and landscape layouts, including recognized Step 1 content.
- Heading, body content, anchor, page-break setting, and table-of-contents behavior remain intact.
- HTML/PDF rendering and Word export reflect the selected state consistently.
- Reimport behavior for an explicitly selected callout state is defined and covered by tests, so the result is not an accidental side effect.

## Scope boundary

This issue does not redesign Step 1 callout recognition, callout visual design, or general page-overflow handling.

mikec-ai:
## Summary

- allow editors to add or remove callout-box styling from a subsection
- explain the existing Step 1 portrait placement behavior for recognized callouts
- clarify that reimport uses styling from the source document
- add regression coverage for editing, rendering, navigation, Word export, and reimport

## Why

Subsections already stored callout state and Builder already rendered both callout and regular subsections, but the edit screen submitted the existing value through a hidden field. Editors therefore could not correct callout styling without changing and reimporting the source document or requesting engineering support.

This exposes the existing state as an intentional editing choice without redesigning callout recognition or rendering.

Closes #772.

## User impact

Editors can now apply or remove callout styling while preserving the subsection's name, heading level, content, order, anchor, page-break setting, and table-of-contents behavior.

Recognized Step 1 callouts retain their existing portrait right-margin placement. Reimport remains source-authoritative, and the previous manual state remains available in the archived revision.

## Validation

- 1,577 tests passed before final rebase
- 27 focused subsection-edit and reimport tests passed after rebasing onto the latest `main`
- Django system checks passed
- Black and isort checks passed
- `git diff --check` passed
- independent code review found no actionable issues

## Visual evidence

**Before**, the subsection edit screen did not expose the stored callout state. 

subsection-edit-before

**After** this change, it presents a USWDS tile checkbox with placement and reimport guidance.

image

### DIFF
--- a/nofos/nofos/templates/nofos/subsection_edit.html
+++ b/nofos/nofos/templates/nofos/subsection_edit.html
@@ -65,19 +65,6 @@
         </legend>
       {% endif %}
 
-      <!-- callout_box -->
-      <div class="form-group margin-top-4 width-mobile">
-        <div class="usa-checkbox">
-          <input class="usa-checkbox__input usa-checkbox__input--tile" id="callout_box" type="checkbox" name="callout_box" {% if form.callout_box.value %}checked{% endif %}>
-          <label class="usa-checkbox__label" for="callout_box">Use callout box styling
-            <span class="usa-checkbox__label-description">
-              Callout boxes use an accent color to call attention to important content. In Step 1 of portrait documents, recognized Key facts, Key dates, and question callouts appear in the right margin.
-            </span>
-          </label>
-        </div>
-        <p class="usa-hint margin-top-1">Re-importing this NOFO replaces this setting with the styling from the source document.</p>
-      </div>
-
       <!-- name -->
       <div class="form-group">
         {% with id="name" label=form.name.label value=form.name.value hint=form.name.help_text error=form.name.errors.0 input_type=form.name|input_type %}
@@ -137,6 +124,20 @@
           </select>
         </div>
       {% endif %}
+
+      <!-- callout_box -->
+      <div class="form-group margin-top-4 width-mobile">
+        <div class="usa-checkbox">
+          <input class="usa-checkbox__input usa-checkbox__input--tile" id="callout_box" type="checkbox" name="callout_box" {% if form.callout_box.value %}checked{% endif %}>
+          <label class="usa-checkbox__label" for="callout_box">Use callout box styling
+            <span class="usa-checkbox__label-description">
+              Best for short, high-priority content that applicants need to notice quickly. Avoid using callout boxes for long narrative content or detailed policy explanations.
+            </span>
+          </label>
+        </div>
+        <p class="usa-hint margin-top-1">In Step 1 of portrait documents, recognized Key facts, Key dates, and question callouts appear in the right margin.</p>
+        <p class="usa-hint margin-top-1">Re-importing this NOFO replaces this setting with the styling from the source document.</p>
+      </div>
 
       <!-- html_class -->
       <div class="form-group margin-top-4">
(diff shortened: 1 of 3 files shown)


mikec-ai:
@benbrasso maybe we can use this one as the test case for Jana and Yasmine to review in dev?

benbrasso:
@mikec-ai, yes, agreed. 

Separately, review the following guidance in the Builder Wiki on when to use callout boxes and determine if wording adjustments are needed in the new component in NOFO Builder: 

---

**When should I use a callout box (or callout component)?**

Callout boxes are best for short, high-priority content that applicants need to notice quickly. Common uses include:

- Contact and support information
- Reminders and post-award requirements
- Action-oriented information that stands on its own

Avoid using callout boxes for long narrative content or detailed policy explanations — those belong in the body of the NOFO. The goal is to surface information that applicants might otherwise miss, not to highlight everything.

**Adding a callout box in NOFO Builder**

Select Configure section on the relevant section, then select Add subsection where you want the callout box to appear. Check the Is callout box checkbox and enter your content.

**Importing a callout box from Word**

NOFO Builder automatically recognizes 1x1 tables in your Word document as callout boxes on import. Make sure any content you want displayed as a callout box is inside a 1x1 table in Word before importing.

benbrasso:
@mikec-ai I'd also consider positioning it immediately before the page break component. I.e., semantically, it probably makes sense to identify the subsection, then decide on decorative options. 

Also, take a look in prod now, where is the current callout box positioned on the edit subsection page. Ideally, we'd be consistent with its position on the edit pages.

benbrasso:
Ooooh, now I see why you selected that position for the checkbox

---

Screenshot 2026-07-28 at 12 19 17 PM

benbrasso:
Well, here's the edit page when "adding a new subsection"

---

Screenshot 2026-07-28 at 12 20 35 PM

mikec-ai:
Addressed the callout guidance and placement feedback in `428bd932`:

- moved the callout control after the subsection name and heading level, immediately before the page-break option
- added the Wiki guidance that callouts are for short, high-priority content and should not be used for long narrative or detailed policy content
- retained the Step 1 placement and reimport behavior notes

Validation: 1,584 full-suite tests and 27 focused edit/reimport tests passed.

Updated screenshot:

subsection-edit-after

mikec-ai:
@benbrasso see the last comment about the update made

benbrasso:
@mikec-ai any reason we don't want to use the same component and content that we already have built? I think the help text isn't needed. A lot of information creates unnecessary noise and cognitive load. 


---

Screenshot 2026-07-28 at 12 20 35 PM

mikec-ai:
@benbrasso reused the same component and content now

benbrasso (approved):
approved to merge

PR #777, opened 2026-07-28
Restore guidance for mistagged Word headings

### LINKED ISSUE #776: Import error no longer identifies accidentally mistagged heading content
## Problem

When a Word document contains paragraph-length text that was accidentally given a heading style, NOFO Builder blocks the import but does not identify the actual problem or the affected content.

Ben encountered this with a document whose only issue was one ordinary paragraph styled as a heading. Builder showed:

> We couldn’t create this NOFO
>
> NOFO Builder could not create a valid NOFO from the uploaded document.
>
> Review the document’s required metadata and heading structure.
>
> Save the document, then select it again.
>
> Error code: `IMPORT-CREATE-INVALID`

The guidance leaves an editor to inspect all metadata and headings without knowing that one paragraph has the wrong style.

## Why this matters

- The document’s cause is specific and repairable, but the message makes it appear broad and uncertain.
- Editors may need trial and error or support from someone who can inspect application logs.
- The error does not name or locate the content that needs to be corrected.

## Code investigation

This is a known failure mode previously tracked in #25 and addressed by #97.

The current import path still detects useful details:

- Section names are limited to 250 characters; subsection names are limited to 400 characters.
- During document creation, a name-length validation failure is recognized as a likely paragraph incorrectly styled as a heading.
- The generated validation message includes the affected object type, its order, the offending text, and a warning that other mistagged headings may exist.
- That detailed validation error is retained in server logs.

However, the new-import handler groups every non-OpDiv model validation failure under `IMPORT-CREATE-INVALID` and replaces the detailed diagnosis with generic metadata/heading guidance. This behavior was introduced when the standardized blocking import pages were added in #748.

The re-import path also replaces document validation details with a generic `REIMPORT-DOCUMENT-INVALID` response, so the same failure may be difficult to diagnose there as well.

## Expected outcome

When import is blocked because paragraph-length content was styled as a heading, an editor can understand:

- that heading styling is the reason the document cannot be imported;
- which content is affected, or where in the document to look;
- what needs to be corrected before trying again.

The message should remain safe for users and should not expose unrelated internal validation or implementation details.

## Acceptance criteria

- [ ] A new import with a paragraph-length section or subsection heading does not fall back to the generic `IMPORT-CREATE-INVALID` guidance.
- [ ] The user can identify the mistagged content or its document location from the error.
- [ ] The error distinguishes this failure from unrelated metadata and document-validation failures.
- [ ] Re-import behavior for the same input is explicitly evaluated.
- [ ] Automated coverage verifies both section and subsection name-length failures.
- [ ] Valid documents and other import error paths remain unchanged.

## References

- Original issue: #25
- Original fix: #97
- Standardized import-error work after which the detail was lost: #746 / #748

mikec-ai:
Closes #776.

## What changed

- Restores specific, actionable guidance when paragraph-length content is accidentally styled as a Word heading.
- Identifies the affected heading content without exposing unrelated validation details.
- Applies the same handling across Builder, Compare, and Composer import paths.
- Makes Builder re-import atomic so a failed import cannot remove or partially replace the existing NOFO.
- Keeps unexpected validation failures on the existing generic, safe error pages.

## Why

The standardized import error work in #748 caused a known mistagged-heading error to fall through to generic `IMPORT-CREATE-INVALID` guidance. Editors could no longer tell which paragraph had the incorrect Word style or how to repair the document.

## Validation

- Added a real `.docx` fixture containing an ordinary paragraph accidentally styled as a heading.
- Exercised new Builder import, Builder re-import, new Compare import, Compare import into an existing document, and Composer import.
- Verified each route returns `IMPORT-HEADING-TOO-LONG` with the affected content rather than its generic fallback.
- Verified a failed Builder re-import leaves the original sections and subsections unchanged.
- Rendered and visually inspected the Word fixture.
- Full test suite: 1,581 tests passing.


benbrasso (approved):
Approved to merge

PR #791, opened 2026-08-05
Add vulnerability scans to CI via new workflow

dfitchett:
## What changed

Adds the three vulnerability scans from the deployment pipeline to CI, so image vulnerabilities are caught on the PR instead of at deploy time.

**`.github/workflows/code_scans.yml`** (new) — reusable workflow with four jobs:

| Job | What it does |
|---|---|
| `build` | Builds the prod image (`make build IS_PROD_ARG=1`), `docker save`s it, caches it |
| `trivy-scan` | `aquasecurity/trivy-action@master` — OS vulns + secrets, `exit-code: 1`, `ignore-unfixed: true` |
| `anchore-scan` | `anchore/scan-action@v6` — JSON + table, `severity-cutoff: medium`, `fail-build: true` |
| `dockle-scan` | `erzz/dockle-action@v1.4.1` — container lint, `failure-threshold: WARN` |

**`.github/workflows/main.yml`** — new `code-scans` job calling the above, running in parallel with the existing `ci` job on every PR and on push to `main`.

**`.github/workflows/django_ci.yml`** — removed its standalone `pull_request` trigger. It had one *and* was called by `main.yml`, so Django CI ran twice on every PR (visible as both `ci` and `ci / ci`). `main.yml` is now the sole entry point for both workflows.

**Job naming** — the old `ci / ci` check name said nothing about what ran. Jobs are now named for what they actually do ("unit tests" specifically, since integration tests aren't included):

| Old check name | New check name |
|---|---|
| `ci`, `ci / ci` (ran twice) | `Run Tests / Run Unit Tests` |
| `code-scans / Trivy Scan` etc. | `Code Scans / Trivy Scan` etc. |

**Three new config files** — the scans read these from the workspace, and none existed in this repo. Without them the scans would use defaults and fail on findings the deploy pipeline ignores.

- `.grype.yml` — `fail-on-severity: medium`, the three `fix-state` ignores (`not-fixed`, `wont-fix`, `unknown`), and the seven CVE safelist entries with their justification comments.
- `.trivyignore` — upstream's has no active entries (all comments); it's included because the `bypass_ignore` path does `rm .trivyignore`, which needs the file to exist.
- `.dockleconfig` — supplies `DOCKLE_ACCEPT_FILES`.

## Why

The deployment pipeline in HHS/simpler-grants-gov already gates on these scans via [`vulnerability-scans-nofos.yml`](https://github.com/HHS/simpler-grants-gov/blob/main/.github/workflows/vulnerability-scans-nofos.yml), but only at deploy time — after the code has merged here. This moves the same rules onto the PR.

All scan parameters, the Trivy DB-caching sequence, the `TRIVY_SKIP_FILES` PyJWT false-positive skip, the runtime `.dockleignore` carrying `DKL-DI-0006`, the `bypass_ignore` input, and the config file contents are taken from the reference workflow and its config files.

## Notes for reviewers

**The build is its own job.** Upstream builds and caches the image in a separate workflow (`ci-nofos.yml`) that the scan workflow restores from; there's no equivalent here. Rather than build the image three times — identical image, triple the runner minutes — the `build` job fills that role and the three scan jobs restore its cached image using the same cache key. Trade-off: the image tar is restored three times, so if cache transfer turns out slower than rebuilding, this is the knob to turn.

**`.dockleconfig` is kept verbatim, and most of it is inert here.** Upstream's copy is shared across every image in the monorepo, so only the `app/`-prefixed entries can ever match this image (`app/nofos/bloom_nofos/settings.py` → `/app/nofos/...` in the container). The `api/` and `analytics/` entries belong to other images and never match. I kept them so syncing with upstream stays mechanical instead of requiring judgment — happy to trim to just the six `app/` entries if you'd prefer.

**Three files now need to stay in sync with upstream** (`.grype.yml`, `.trivyignore`, `.dockleconfig`) when a CVE or accepted file is added or dropped. Each carries a comment saying so. Open to a better approach if one exists.

**One deviation from upstream's step conditions:** the Anchore scan steps are gated on `steps.load.outcome == 'success'` rather than a bare `if: always()`, so a cache-restore failure surfaces as the load error rather than a confusing "image not found" from the scanner.

**No secrets are used.** The scans need no registry auth (the image is local) and grype/trivy pull their vulnerability DBs unauthenticated, so the `code-scans` job passes no secrets. The existing `ci` job's `secrets: inherit` is left as-is.

## ⚠️ Requires a ruleset change before merging

Deduplicating and renaming the Django CI run changes its check context, and the old name is a required check:

- The `no force pushes to main` ruleset requires the status check **`ci`**.
- That bare `ci` context was produced only by the standalone `pull_request` trigger this PR removes. Reusable-workflow checks are always `caller-job / called-job`, so bare `ci` cannot be reproduced — the run now reports as **`Run Tests / Run Unit Tests`**.

**The ruleset's required check must be changed from `ci` to `Run Tests / Run Unit Tests` before this merges**, otherwise every subsequent PR will block on a check that never reports.

While that ruleset is open, these are the other check names now available if you want them required too: `Code Scans / Build Image`, `Code Scans / Trivy Scan`, `Code Scans / Anchore Scan`, `Code Scans / Dockle Scan`.

## Testing

The workflow parses as valid YAML, the job graph and `uses:` references resolve, and every scan parameter was diffed against the reference workflow.

**Whether the image actually passes these scans is unverified until CI runs on this PR.** The reference workflow scans the image built from this repo's Dockerfile, so behavior should match what the deploy pipeline already sees — and the `FROM scratch` final stage exists specifically to satisfy Dockle — but the first run here is the real test. If findings surface, the fix is either a dependency bump or a new safelist entry in both copies of the relevant config file.

### DIFF
--- a/.dockleconfig
+++ b/.dockleconfig
@@ -1,0 +1,17 @@
+# This file is allows you to specify a list of files that is acceptable to Dockle
+# To allow multiple files, use a list of names, example below. Make sure to remove the leading #
+# DOCKLE_ACCEPT_FILES="file1,path/to/file2,file3/path,etc"
+# https://github.com/goodwithtech/dockle#accept-suspicious-environment-variables--files--file-extensions
+#
+# This file mirrors the .dockleconfig used by the deployment pipeline in
+# HHS/simpler-grants-gov so that the scan run on PRs here enforces the same rules.
+# Keep the two files in sync when adding or removing an entry.
+#
+# Note: that file is shared across every image built in the monorepo, so it is kept
+# here verbatim to make syncing mechanical. Only the "app/" entries can match this
+# image (they resolve to /app in the container, i.e. app/nofos/... is /app/nofos/...);
+# the "api/" and "analytics/" entries belong to other images and simply never match.
+#
+# The apiflask/settings file is a stub file that apiflask creates, and has no sensitive data in. We are ignoring it since it is unused
+# The settings.py files under the docx have been reviewed and do not hold sensitive info in them.
+DOCKLE_ACCEPT_FILES=api/.venv/lib/python3.14/site-packages/newrelic/api/settings.py,api/.venv/lib/python3.14/site-packages/apiflask/settings.py,api/.venv/lib/python3.14/site-packages/docx/oxml/settings.py,api/.venv/lib/python3.14/site-packages/docx/parts/settings.py,api/.venv/lib/python3.14/site-packages/docx/settings.py,analytics/.venv/lib/python3.14/site-packages/newrelic/api/settings.py,analytics/.venv/lib/python3.14/site-packages/jedi/settings.py,app/.venv/lib/python3.14/site-packages/isort/settings.py,app/nofos/bloom_nofos/settings.py,app/.venv/lib/python3.14/site-packages/easyaudit/settings.py,app/.venv/lib/python3.14/site-packages/constance/settings.py,app/.venv/lib/python3.14/site-packages/cssutils/settings.py,app/.venv/lib/python3.14/site-packages/martor/settings.py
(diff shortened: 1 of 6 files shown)


amponce (approved):
Awesome! Super useful

PR #796, opened 2026-08-11
Silently strip redundant "Before You Begin" heading on Word import

### LINKED ISSUE #794: Ignore/strip "Before You Begin" heading and section on Word import
## Scenario
GS AM's Word export for GS AM-authored NOFOs includes a "Before You Begin" heading and section (SAM.gov/Grants.gov registration reminders, "how to use this NOFO" guidance) ahead of Step 1. This content is redundant with information NOFO Builder already surfaces elsewhere, and its continued presence adds unnecessary manual cleanup for writers during NOFO Builder authoring.

We asked GS AM to remove this heading/section from their export template. Per their response, removing it would have broad downstream impact on other NOFOs already relying on this section in their designed format, so they're not able to remove it on the template side. GS AM confirmed that removing it did not cause any negative effects when they tested importing into NOFO Builder — so this is safe to handle defensively on our side instead.

## What Needs to Be Done
- On Word document import, detect the "Before You Begin" heading (and its associated section content, up to the next heading) and exclude it from the resulting NOFO, so it does not require manual removal by the writer post-import.
- Investigate whether this should be:
  - Silently stripped on import (no writer action needed), or
  - Flagged for the writer to confirm removal (in case a future GS AM export legitimately needs the section retained for some agencies).
- Confirm detection logic is resilient to minor variations in heading text/casing, since this is sourced from an external, non-NFB-controlled template.
- Flag as a design decision if there's any ambiguity about which approach (silent strip vs. flagged) is preferred — don't guess.

## Acceptance Criteria
- [ ] A Word document containing a "Before You Begin" heading and section imports into NOFO Builder without that heading/section appearing in the resulting NOFO.
- [ ] The fix does not affect NOFOs that don't contain this heading (no false positives on similarly named or structured content).
- [ ] Behavior is documented (silent strip vs. flagged) and confirmed with the team before merge.
- [ ] Existing NOFOs already in NOFO Builder are unaffected — this only applies to net-new Word imports.

benbrasso:
Closes #794

### Summary

GS AM (GrantSolutions Announcement Module) Word exports sometimes include a "Before You Begin" heading and section. This duplicates the "Before you begin" page NOFO Builder already generates for every NOFO from its own `before_you_begin` field, so it's always safe to remove — no writer decision is needed.

This PR adds `decompose_before_you_begin_section`, wired into the existing `process_nofo_html` cleanup pipeline, so the heading (and any real content under it) is stripped automatically and silently during Word/HTML import.

### Approach

- Matches any heading (`h1`–`h6`) whose text, after normalizing case/whitespace/punctuation, reads exactly `"before you begin"`. Exact matching (not a substring/prefix match) avoids false positives on unrelated headings like "Before You Begin Your Application."
- Removes the heading plus everything that follows it, up to whichever comes first:
  - the next heading of the same or a higher level (the section boundary), or
  - a `"Label: value"`-style metadata line.
- The metadata-line boundary matters: real GS AM exports place the NOFO's front matter (OpDiv, Agency, Opportunity Number, Tagline, etc.) directly after this heading with no heading of its own in between. That metadata is read later in the pipeline by `suggest_nofo_opdiv`, `suggest_nofo_agency`, and friends, so it must never be swept up as if it were part of the "Before You Begin" section. This was caught by testing against a real reference export and fixed before this PR's final version.
- Documents that don't include the heading are completely unaffected — verified against a reference export with no "Before You Begin" content at all.

### Testing

- Added `DecomposeBeforeYouBeginSectionTests` (9 cases) covering: basic removal, heading-level boundaries, case/whitespace/punctuation tolerance, no-op on documents without the heading, avoiding false positives on similar-but-different headings, end-of-document removal, and — critically — preserving a metadata block that immediately follows the heading (with and without real duplicate prose in between).
- Ran the full `nofos` test suite (974 tests) — all passing, no regressions.
- Manually verified end-to-end against two real GS AM reference exports by running them through `mammoth` → `process_nofo_html` → `suggest_nofo_*`: the heading and duplicate content are removed, front-matter metadata is preserved and correctly auto-populated, and no phantom "Before You Begin" section appears in the resulting NOFO structure.


benbrasso:
@mikec-ai can I get a review on this?

mikec-ai:
I reviewed this against current main and found two interactions worth fixing before merge:

- A split metadata field, such as `Opdiv:` in one paragraph and its value in the next, could be removed with the duplicate Before You Begin section.
- A Heading 2 Before You Begin section placed before the first Heading 1 could trigger the mixed-heading import error before the redundant section was removed.

I merged current main into this branch, preserved standalone metadata labels and their following values, and moved the initial cleanup ahead of heading-level detection. I also added regression coverage for both cases.

Validation: 588 import and NOFO unit tests pass, along with Black, isort, and the repository pre-commit checks.

benbrasso:
Validated on dev

https://nofos.dev.simpler.grants.gov/nofos/abe15a7e-7ab7-48c3-8fe6-ee89e1d479fa/edit

---

Screenshot 2026-08-12 at 10 42 16 AM

PR #797, opened 2026-08-12
[Experimental] Add on-demand readability metrics to the NOFO editor

mikec-ai:
## Experimental review

This draft PR provides a working integration of the standalone `hhs-nofo-metrics` package in NOFO Builder. It is intended for editor and developer testing before a production rollout decision.

The feature is disabled by default, labeled **Beta**, calculates lazily the first time an editor opens the panel, and does not persist results or add database schema.

## What changed

- Pin the Agile Six `hhs-nofo-metrics` v0.5.2 wheel and lock its artifact hash.
- Render the same source-native HTML fragment used for Builder's document export and send it directly to the package.
- Add an authenticated, permission-aware readability endpoint for the current server-side NOFO revision.
- Add a compact, collapsed **Readability metrics — Beta** panel to the NOFO edit screen.
- Calculate once when an editor first opens the panel and retain an explicit **Recalculate** action.
- Display six guidance-backed metrics in a 3×2 grid: word count, words per sentence, sentences per paragraph, Flesch Reading Ease, Flesch-Kincaid Grade Level, and passive-sentence percentage.
- Compare applicable metrics with default targets: 13,500 words, 15 words per sentence, 3 sentences per paragraph, Reading Ease of 39 or higher, and 8% passive sentences.
- Display a grade-level threshold range of 11.5–12.5 without inferring the NOFO category: values at or below 11.5 are **Within target**, values between the thresholds say **Check NOFO type**, and values above 12.5 say **Needs improvement**.
- Use assessment language rather than pass/fail language.
- Explain that word count covers all NOFO text while sentence-based formulas use complete sentences, including the excluded word count, percentage, and examples of why text may fall outside that denominator.
- Explain that editing clears transient results until the panel is opened again.
- Surface calculation notes, loading, timeout, and safe error states.
- Keep the feature behind `HHS_NOFO_METRICS_ENABLED`, which defaults to `false`. Environments can override or hide target comparisons with `HHS_NOFO_METRIC_GOALS`.

## Why

This tests the smallest useful Builder integration before adding persistence, background work, or automatic save-time calculation. Using Builder's structured export HTML avoids PDF reconstruction errors and keeps the measurement source aligned with current Builder content.

The package calculation is inexpensive enough to run when the panel first opens. This avoids database, cache, and background-job complexity during the experimental release.

## Local test instructions

1. Check out this branch and install dependencies with Poetry.
2. Start Builder with the feature enabled:

   ```bash
   HHS_NOFO_METRICS_ENABLED=true poetry run python nofos/manage.py runserver
   ```

3. Open the edit page for a populated local NOFO.
4. Expand **Readability metrics — Beta** and confirm the first calculation starts automatically.
5. Confirm all six metrics, target comparisons, the calculation-scope explanation, and calculation notes appear.
6. Close and reopen the panel and confirm it does not calculate again.
7. Select **Recalculate** and confirm a fresh calculation completes.
8. Edit the NOFO, return to its main edit page, and reopen the panel to confirm the metrics use the current revision.

To override or hide comparisons locally, set `HHS_NOFO_METRIC_GOALS` to a JSON object. An empty object hides all target and assessment language. Goals support `at_most`, `at_least`, and `at_most_by_category` comparisons.

The implementation was exercised locally with populated THCGME, CDC K01, and CDC DD26 NOFOs. Their words outside complete sentences represented 15.0%–18.6% of total word count, supporting the explicit scope explanation.

## Feedback requested

- Is this the right placement and level of prominence for editors?
- Are the metric names, targets, calculation-scope explanation, and notes understandable?
- Are the three grade-level assessment states clear without Builder inferring the NOFO type?
- Is first-open lazy calculation the right experimental behavior?
- Is on-demand, non-persisted calculation sufficient for this initial release?
- Does the source-native HTML integration boundary fit the intended Builder architecture?

Persistence, saved review snapshots, background calculation, and status-triggered workflows are deliberately deferred.

## Validation

- Focused readability integration suite — 13 tests pass.
- Full Builder test suite — 1,674 tests pass after merging current `main`.
- `poetry run pre-commit run --all-files` — passes.
- Manual local UI verification against a populated THCGME NOFO using the published Agile Six v0.5.2 wheel, including automatic first-open calculation, explicit recalculation, the six-card grid, calculation-scope disclosure, and all three grade-level assessment states.
- Browser console — no warnings or errors during the tested flows.


benbrasso:
Would it be valuable (to some or all potential user types) to understand where their NOFO's metrics are in comparison to HHS ASFR's NOFO metrics goals? 

It might create more noise by adding more information, so, it's more of a question for the initial release. Example: <br><br>


image

benbrasso:
On the scope explanation text — "Word count uses the broader document scope (12,205 words)" — I found this confusing on first read against a real NOFO (10,018 words / 588 sentences vs. 12,205 total). Once I dug into the PR description I think I understand the mechanism: word count covers the full document (headings, table cells, list items, etc.), while the sentence-based metrics only run over text that parses as complete sentences, since Flesch/FK-grade formulas need actual sentences to be meaningful.

If that's right, I'd suggest making that explicit in the copy rather than relying on "broader document scope" alone — something like: *"Word count includes all document text (12,205 words). Readability scores are based on the 10,018 words in complete sentences (588 sentences), since headings, labels, and list items aren't sentences."*

Two follow-up questions:
1. Is an ~18% gap between the two counts (like in my test NOFO) typical, or does that vary a lot by NOFO structure? If it's usually this large, it might be worth surfacing the excluded-word count too, not just the two totals.
2. Should the "broader document scope" framing survive into the eventual non-beta release, or is this phrasing itself still up for revision?

---

Screenshot 2026-08-14 at 10 48 30 AM

benbrasso:
On the intro copy for the panel — "Check readability for the current NOFO. Results aren't saved." — this led me to expect that if I calculated metrics, edited the NOFO, and came back, my previous results would still be there and I'd just need to hit recalculate to refresh them. Instead, the results disappeared entirely after saving my edit and returning to the page.

I think clearing the metrics on every revision is the right behavior (stale metrics next to updated content would be worse), but the current copy doesn't tell the user that's what's going to happen — "not saved" reads as "not persisted to the database," not "will disappear from your screen the moment you save an edit."

Suggest something more explicit, e.g.:

> "Calculate readability for the current revision. Metrics aren't stored, and clear automatically whenever the NOFO is edited."

or shorter:

> "Metrics reflect the current revision only. Editing the NOFO clears them until you recalculate."

Either way, I'd rather the user know upfront that saving = losing the displayed numbers, rather than discovering it by surprise like I did.

---

Screenshot 2026-08-14 at 10 55 59 AM

benbrasso:
**Separate thread — accessibility pass on the panel:**

## Accessibility review — Readability metrics panel (WCAG 2.1 AA, PR #797)

Ran a contrast/focus/ARIA pass against the compiled `uswds.css` values, then checked whether the flagged elements are net-new to this panel or shared patterns used elsewhere in Builder — wanted to make sure any suggested fix stays scoped to this PR rather than rippling into other pages.

### Confirmed contrast failure — isolated to this panel, safe to fix here

**`text-base` (#71767a) on `bg-base-lightest` (#f0f0f0) → 4.03:1, needs 4.5:1 — FAIL.** Four instances, all in `readability_metrics_panel.html` / `nofo_readability_metrics.js`:

- Header status text (`data-metrics-summary-status`)
- Per-metric status text (`data-metric-status`, ×6)
- Goal label text (`data-metric-goal`, built in `showGoal()`)
- Scope explanation paragraph (`data-metrics-scope-summary`)

Confirmed this class combination (`text-base` on `bg-base-lightest`) doesn't exist anywhere else in the codebase — checked all other `bg-base-lightest` usages (5 files) and all other `text-base` usages (11 files), none pair the two. The metric-card container, goal-label logic, and scope-explanation callout are all bespoke to this file too. So fixing this — swap to `text-base-dark`, 5.92:1, passes — touches nothing shared and has zero blast radius on other pages.

### Passing, but worth considering

`text-base` on white (`#readability-metrics-status`) → 4.59:1, technically passes but only by 0.09. Also isolated to this panel — worth bumping to `text-base-dark` for headroom/consistency while you're already touching the other four.

### Goal badges — pass, no action needed

"Within goal" / "Review goal" badges (`text-base-dark` on `bg-success-lighter`/`bg-warning-lighter`) both clear AA comfortably (5.97:1, 6.04:1), and differ by text content so 1.4.1 is satisfied independent of contrast.

### Focus ring — shared/stock behavior, out of scope for this PR

The accordion header's focus ring (`#2491ff` on `#f0f0f0` → 2.81:1, fails 3:1) comes from a single global rule (`.usa-accordion__button:focus`) already applied identically across 11 existing templates in `nofos`, `composer`, `uploads`, and `compare` — this predates the PR and isn't specific to the readability panel. Changing it would visibly alter the focus indicator on every accordion in Builder, which is a global USWDS theming decision, not something to fix here. Flagging for awareness only.

### Net: what I'd actually ask you to change in this PR

Just the 4 (or 5, including the marginal one) `text-base` → `text-base-dark` swaps inside the readability panel's own template/JS. Nothing else needs to move — the accordion shell, focus ring, and other shared USWDS pieces are all pre-existing and out of scope.

benbrasso:
<html><head></head><body><p>Suggested goal targets, based on FY27 guidance:</p>

Metric | Current example | Suggested target
-- | -- | --
Word count | 9,000 or lower | 13,500 or fewer
Words per sentence | 14 or lower | 15 or fewer
Passive sentences | 20% or lower | 8% or lower
Flesch-Kincaid grade level | (none shown) | 11.5 or lower
Characters per word | (none shown) | 5 to 6
Flesch reading ease | 50 or higher | 39 or higher


<p>Notes:</p>
<ul>
<li>Word count, words per sentence, and passive sentences come directly from the guidance's stated numbers.</li>
<li>Grade level uses the general-NOFO threshold (11.5); scientific/research NOFOs get 12.5, but Builder can't yet tell them apart, so defaulting to the stricter number for now.</li>
<li>Characters per word isn't in the guidance — this is our own target.</li>
<li>Reading ease has no guidance target, but can be derived from the guidance's own grade-level and sentence-length numbers — landing at ~39.</li>
</ul>
<p>Also suggest changing the badge labels (keep the green/yellow coloring, just update the text):</p>
<ul>
<li>"Within goal" → <strong>"Within range"</strong></li>
<li>"Review goal" → <strong>"Needs improvement"</strong></li>
</ul>
<p>And updating the goal-line prefix from "Example only:" to <strong>"Target:"</strong> now that these are real values rather than placeholders — e.g. "Target: 13,500 or fewer."</p></body></html>

benbrasso:
On your question about what should be persisted or recalculated automatically in a later iteration — wanted to lay out the actual workflow this would need to support, since I think that shapes the answer more than a general "should we add a database table" framing would. Had Claude Code dig into the codebase to sanity-check feasibility once I had the workflow mapped out; findings below.

**The review cycle, as I understand it:**

An editor makes a round of edits in Builder, then decides the NOFO is ready to hand off for review — first to their Grants Policy Office or division/program lead, and later to HHS and OMB. That handoff decision is the meaningful checkpoint, not every save. The editor gets feedback, comes back into Builder, makes changes, and again decides when it's ready to resubmit. This can repeat across multiple review cycles (division-level, then HHS/OMB, possibly multiple rounds of each) before the NOFO is published.

At each of these "ready for review" moments, I think the editor is going to want a timestamped snapshot of the metrics — both because they likely need to hand those numbers along as part of the review package each cycle, and so they can compare snapshots across cycles to see whether their revisions actually improved or regressed readability.

This isn't just a nice-to-have workflow feature, either — the draft FY27 Content and Clearance Guidance requires a readability metrics table (Flesch Reading Ease, Flesch-Kincaid Grade Level) in the transmittal email accompanying *every* NOFO review package. That reads as a per-submission requirement, not a one-time report, which lines up with the checkpoint model above.

**Recommendation for v1: explicit action only, no status-field dependency.**

- Add a **"Save these results"** button, separate from "Calculate metrics," that only appears after a successful calculation. Calculating is exploratory (click it many times mid-edit); saving is the deliberate checkpoint, and should require its own intentional click.
- On save, recompute server-side rather than trusting the client's last-fetched result, so the saved snapshot reflects current content even if the editor calculated once, kept editing, then saved later.
- Deliberately **not** tying this to the NOFO's `status` field for v1. Builder already has a status field (draft, active, ready-for-qa, review, doge, published, paused, cancelled), and three of those values map directly onto the actual handoff moments in this workflow — `ready-for-qa` for the pre-QA handoff, `review` for division/OpDiv-level review, and `doge` for Dep Sec/HHS-level review. There's also only one editor-facing code path that changes status, so hooking into it would be architecturally cheap.

  That said, the status dropdown submits immediately on selection with zero confirmation, so it's easy to change accidentally or while testing — a poor trigger for something that creates a permanent review-package record. So for v1, I'd rather keep saving as an explicit, deliberate action.

  Longer-term, this does feel like a real opportunity worth revisiting: once the explicit save action exists, we could layer in an optional prompt when status changes to one of those three values ("Status changed to In review — save a snapshot for your review package?") as a convenience for editors who keep status current, without ever depending on it as the only way to capture a snapshot.

Mockup of what this could look like:

Screenshot 2026-08-14 at 1 09 39 PM

---

**Schema:** genuinely net-new — nothing like this exists today (the audit log explicitly excludes status/update changes from user-facing history, and the one existing "modified since publish" field is a single flag, not a timestamped collection). Follows the existing Section/Subsection FK-to-Nofo pattern: a new `NofoReadabilityMetricsSnapshot` model (Nofo FK, timestamp, user, NOFO status *at capture time*, and the metrics payload stored as JSON so it doesn't need a migration every time a metric changes).

**Display:** a compact "Saved snapshots" list directly in the panel (last 3-5, with timestamp and an up/down delta vs. the immediately prior snapshot), plus a "See all snapshots" link out — mirroring the existing "Last updated by... See all updates" pattern already sitting right above the status dropdown on the edit screen. Keeps the "did my revision help" comparison exactly where editors are already looking.

**Deferred for a later iteration:** status-triggered prompts, comparing arbitrary snapshot pairs (not just previous-vs-current), exporting snapshots directly into a transmittal-email-ready format, and any changes to the separate NOFO History page.

**Flag:** this is a real net-new model + migration + new view, not a small addition to the existing on-demand endpoint — and persistence should stay behind the same `HHS_NOFO_METRICS_ENABLED` flag so it can ship and be pulled back together with the rest of this feature.

benbrasso (commented):
left comments; great work

dfitchett:
I cannot speak much to the usability side of things because I don't know a whole lot about the current process that users have for using the tool. However, here's my thoughts on the following:

> Is on-demand calculation the right first release behavior?
> What, if anything, must be persisted or recalculated automatically in a later iteration?

I believe that we could automatically show the metrics on page load. For example, if a user edits a section, upon saving the edit, it could asynchronously run the metrics analysis and store the results in a cache for that nofo id. Then when viewing the main nofo page, it would just get the cached metrics and display them. 

I don't believe we need any DB persistence of the data until there's a need to keep readability metrics history. But when that comes along, it would be pretty easy just creating a metrics table with a foreign key for the nofo id and potentially the id correlating to the change that caused the metrics to be re-run. Additionally at that point we could include some metric data on the `/history` page in the `Latest changes` table. 

> Does the source-native HTML integration boundary fit the intended Builder architecture?

I'd be curious to see if the metrics generated using the metrics lib [from_pdf](https://github.com/mikec-ai/hhs-nofo-metrics/blob/c987d892349ad62a0a34a2f73aa019d2c4488664/src/hhs_nofo_metrics/sources.py#L74) would generate the same results as the HTML generated. I assume it should, but it's probably dependent somewhat on the pdf reading library. I think if the scores accurately resemble the desired analysis, reading only the html is sufficient.

mikec-ai:
@benbrasso I incorporated the latest readability feedback and would appreciate another pass when you have time.

The updated experiment now:

- pins the released `hhs-nofo-metrics` v0.5.1 package;
- adds sentences per paragraph from the package's source-native sentence scope;
- supports multiple labeled targets for one metric, so both general and scientific/research grade-level targets can be shown without Builder guessing which kind of NOFO it is;
- uses a more compact card layout with the grade-level comparisons grouped together;
- makes the full-document word count versus complete-sentence readability scope explicit, including the excluded word count and percentage;
- explains that editing clears the transient results;
- uses the higher-contrast text treatment you recommended; and
- includes loading timeout and safer error states.

I re-tested the complete panel locally against the published v0.5.1 wheel and a populated THCGME NOFO. The seven metrics rendered correctly, both grade-level comparisons appeared, the scope explanation matched the returned counts, and there were no browser-console errors.

Could you take another look and let me know specifically whether:

1. the information hierarchy and compact layout feel right;
2. showing both grade-level targets is clearer than selecting one implicitly;
3. the target and assessment language is understandable; and
4. this on-demand, non-persisted behavior is sufficient for the experimental version?

The feature remains disabled by default and this PR is still a draft for testing.

benbrasso:
Following up on my earlier comment about goal targets and badge language — two things on the latest update:

**On showing both grade-level targets:** I don't think we need to display both. A NOFO is either a scientific/research NOFO or it's not — showing a grade-level target for the category it doesn't belong to doesn't add useful information, and I think just adds noise to the panel. My earlier recommendation was to show a single range instead (11.5–12.5) rather than picking one threshold or showing both — users will already know which end of that range applies to them based on the FY27 clearance guidance and what kind of NOFO they're writing, without Builder needing to guess or display an irrelevant number alongside the relevant one.

**On the badge label:** "Review target" isn't clear to me — what am I reviewing, and against what? I'd recommended "Needs improvement" in my earlier comment, which I still think reads more plainly to an editor.

**On missing targets:** Right now Flesch Reading Ease and Characters per word don't have targets configured. Flesch Reading Ease is actually required in the NOFO Summary Template's readability table per the FY27 guidance, so it should get a target too, not be left out — I'd suggested "39 or higher" earlier, derived from the guidance's own words-per-sentence and grade-level numbers. And since Characters per word would otherwise be the only metric on the panel without a target at all, I think we should give it one as well, even though it isn't addressed directly in the guidance — I'd suggested "5 to 6" as our own reasonable target.

benbrasso:
Following up on my earlier comment about the scope explanation — this update covers the core ask well: it makes the word count and complete-sentence basis explicit, and even surfaces the excluded word count and percentage, which directly answers my question about whether that gap is typical.

One thing I think is still missing: the reasoning behind the exclusion. My understanding from the PR description is that headings, table cells, and list items are excluded because they aren't complete sentences, and the readability formulas require sentences to calculate meaningfully — but I want to flag that this exclusion logic is a Builder/package design choice, not something defined in the FY27 guidance itself. The guidance just says to generate metrics using Word's built-in readability statistics or an approved online calculator; it doesn't specify a sentence-detection methodology. So before we lock in language that states this as fact, can you confirm I've got the mechanism right?

Assuming so, something like this would make the "why" explicit without overstating it as an HHS-defined rule:

> "Word count includes all NOFO text: 11,856 words. Readability scores use 9,859 words in 582 complete sentences — headings, table cells, and list items aren't sentences, so their 1,997 words (16.8%) aren't counted."

Also still curious on my second follow-up from before: should the "broader document scope" framing survive into the eventual non-beta release, or is this phrasing itself still up for revision?

benbrasso:
> Is on-demand, non-persisted behavior sufficient for the experimental version?

Yes

benbrasso:
Everything else looks good, Mike. Thanks

mikec-ai:
@benbrasso I incorporated your latest feedback: the panel now shows one display-only grade-level range, adds targets for Reading Ease and characters per word, uses “Needs improvement” when a target is missed, and explains why some text is excluded from sentence-based formulas without presenting that methodology as HHS policy.

I rechecked the populated THCGME panel locally, and the full 1,667-test suite also passes locally and in GitHub Actions. When you have a chance, could you take another look?

benbrasso:
Suggest dropping **Characters per word** from the panel. Looking back at the FY27 guidance, every other metric here traces to it directly — word count, words per sentence, sentences per paragraph, and passive sentences are all from the writing standards' "strive for"/limit language, and Flesch-Kincaid Grade Level and Flesch Reading Ease are both required in the NOFO Summary Template's readability table. Characters per word doesn't appear anywhere in the guidance or the required reporting — the 5–6 target we gave it earlier was just something we invented so it wouldn't be the only metric without one, not something tied to any actual standard. Since it's not part of the writing standards or the review/clearance requirements, I don't think it needs to be in this panel.

Dropping it also gets us to an even 3x2 grid of guidance-backed metrics:

| Word count | Words per sentence | Sentences per paragraph |
|---|---|---|
| **Flesch reading ease** | **Flesch-Kincaid grade level** | **Passive sentences** |

Separately: I'm not seeing a "Needs improvement" tag show up on Flesch-Kincaid Grade Level even when the value is outside the target range — in my test NOFO it's at 13.12, above the 11.5–12.5 range, but no badge appears at all. Every other metric on the panel evaluates against its target and shows a badge either way (within target or needs improvement), so this one is inconsistent with the rest. Was this intentional because the range is "display-only," or is the evaluation logic just not wired up yet for this metric? If it's meant to be display-only, I'd still expect some kind of visual signal when a NOFO is outside the range — otherwise this is the one metric on the panel where an editor could be over the limit and never know it.

benbrasso:
Everything else looks good 👍

mikec-ai:
@benbrasso I addressed the last two items: Characters per word has been removed, leaving the six guidance-backed metrics in a 3×2 grid, and grade level now shows Within target at or below 11.5, Check NOFO type between 11.5 and 12.5, and Needs improvement above 12.5.

The branch is current with main and conflict-free. All 1,674 tests pass locally and in GitHub Actions, and I verified all three grade states in the populated local panel with no browser errors. Could you take one final look when you have a chance?

mikec-ai:
Final review and dev validation are complete. I did not find any blocking code, security, accessibility, or deployment concerns.

Validation completed:

- updated the branch from current `main`, including the `sqlparse` 0.6.0 remediation;
- ran the focused readability suite: 13 tests passed;
- ran the complete application suite locally: 1,674 tests passed;
- confirmed all GitHub checks pass: unit tests, image build, Trivy, Dockle, and Anchore;
- deployed this exact branch to dev with `HHS_NOFO_METRICS_ENABLED=true`; and
- smoke-tested the authenticated editor using the dev-only test NOFO.

The live smoke test confirmed that opening the panel calculates automatically, all six metrics and their configured target states render, recalculation succeeds, and saving an edit resets the transient results to Not calculated. I restored the test NOFO content after checking the edit flow. The deployment completed successfully: https://github.com/HHS/simpler-grants-gov/actions/runs/32478867561

For the integration-boundary question: Builder structured HTML is the appropriate source for editor feedback because it measures the source document fragment shared with Word export and avoids PDF extraction/reflow variability. Exact numerical parity with a generated PDF is not an acceptance criterion for this experiment; representative PDF calibration can remain a separate follow-up if needed.

Given the successful final review and dev validation, this is ready to merge behind the feature flag.

PR #807, opened 2026-08-17
Default NIH-group NOFO imports to the eRA "Before you begin" page

benbrasso:
## Background

When a NOFO is imported and NOFO Builder determines it belongs to the NIH group, the design theme is already suggested correctly (NIH's "portrait-nih-white" theme). Separately, every NOFO has a "Before you begin" (BYB) page setting (`Nofo.before_you_begin`) with four choices: `full` (current default for every group), `era` ("BYB page with eRA", referencing eRA Commons, NIH's grants system), `sole_source`, and `none`.

Today, `before_you_begin` defaults to `full` for every group, including NIH — there's no logic giving NIH NOFOs the eRA-specific variant automatically. This PR adds that: new NOFOs imported by an NIH-group user get `before_you_begin = "era"` automatically, the same way `theme`/`cover` already get NIH-aware defaults on import.

## What changed

- `nofos/nofos/nofo.py`: added `suggest_nofo_before_you_begin(nofo_group)`, wired into `suggest_all_nofo_fields()` alongside the existing theme/cover first-time-import defaults.
- `nofos/nofos/views.py`: reordered `NofosImportNewView.handle_nofo_create()` so `nofo.group` is set before `suggest_all_nofo_fields()` runs (previously set after — the NIH check needs it available).
- Tests: unit tests for the new helper, an NIH-group case added to `suggest_all_nofo_fields` tests, and an integration test exercising the real import call path end-to-end.

## Scope / non-goals

- Only affects new imports going forward — no existing NOFOs are touched, no migration involved.
- The global model default (`before_you_begin="full"`) is unchanged, since it's shared by every group.
- Reimporting an already-populated NIH NOFO won't reset a coach's manually chosen BYB page (gated by the existing `first_time_import` check).

## Acceptance criteria

- [x] New NIH-group import sets `before_you_begin` to `era`
- [x] New non-NIH import leaves `before_you_begin` at `full`
- [x] Reimport of an existing NOFO doesn't change `before_you_begin`
- [x] No existing NOFO rows modified by this change
- [x] Full test suite passes

## Test plan

- `poetry run python manage.py test` — full suite passes
- New/updated tests: `test_nofo.py::HTMLSuggestBeforeYouBeginTests`, `test_nofo.py::SuggestNofoFieldsTests::test_suggest_all_nofo_fields_nih_group_sets_before_you_begin_era`, `tests_nofos/test_before_you_begin_import.py`

### DIFF
--- a/nofos/nofos/nofo.py
+++ b/nofos/nofos/nofo.py
@@ -1694,6 +1694,13 @@ def suggest_nofo_cover(nofo_theme):
     return "nofo--cover-page--medium"
 
 
+def suggest_nofo_before_you_begin(nofo_group):
+    if nofo_group == "nih":
+        return "era"
+
+    return "full"
+
+
 def suggest_nofo_theme(nofo_number, opdiv=""):
     opdiv_lower = (opdiv or "").lower()
     if "national institutes of health" in opdiv_lower or re.search(
@@ -1826,6 +1833,10 @@ def suggest_all_nofo_fields(nofo, soup):
         )  # guess the NOFO theme
     if first_time_import:
         nofo.cover = suggest_nofo_cover(nofo.theme)  # guess the NOFO cover
+    if first_time_import:
+        nofo.before_you_begin = suggest_nofo_before_you_begin(
+            nofo.group
+        )  # guess the NOFO "Before you begin" page
 
 
 ###########################################################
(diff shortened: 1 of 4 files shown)


benbrasso:
@mikec-ai when you get a sec; this could use a quick review

Issue #814, opened 2026-08-19
Inline images in NOFO subsections have no supported way to add alt text


benbrasso:
### Problem

Subsection body content is edited via Martor (a markdown/source editor with an Ace textarea + preview pane), not a WYSIWYG editor. There is no image-insert or image-properties UI at all for this content:

- `image-link` and `image-upload` are explicitly commented out of the Martor toolbar config (`MARTOR_TOOLBAR_BUTTONS` in `nofos/bloom_nofos/settings.py`).
- `MARTOR_UPLOAD_URL = ""` — the image upload endpoint is completely disabled.
- `MARTOR_ENABLE_CONFIGS["imgur"] = "false"` — this actually removes Martor's built-in upload button from the DOM entirely.
- Martor has no drag-and-drop support at all (confirmed: no `dragover`/`dragenter`/`drop`/`dataTransfer` handling anywhere in the vendored `martor.bootstrap.js`/`martor.semantic.js`). Its only native image mechanism is a click-to-upload toolbar button, which is fully disabled here three different ways.
- `subsection_edit.js` has no paste/drop handler either (its only behavior is a "copy html_id" button).

As a result, inline images only end up in a subsection's body content in one of two distinct ways, and **neither has any alt-text UI**:

1. **Docx import** — `mammoth.convert_to_html()` (`nofos/nofos/nofo.py`) converts inline Word images into base64 `data:` URIs embedded directly in the HTML, which is then converted to markdown for storage. Word's alt-text/description field is not carried over.
2. **Manual paste/typing** — since there's no image-insert mechanism at all (see above), a user's only option to get an image into a subsection after the fact is to paste or type raw markdown (`![](data:...)`) or HTML (`<img src="data:...">`) directly into the Martor source textarea.

The *only* way for a user to add or fix an image's `alt` text today is to hand-edit the markdown/HTML source directly — e.g. changing `![](data:image/png;base64,...)` to `![Descriptive text](data:image/png;base64,...)`. This is not a reasonable workflow for non-technical end users building NOFOs.

Additionally, for information-dense images (e.g., a concept/logic model diagram with multiple labeled boxes and relationships), a short `alt` attribute can only ever provide a high-level label, not the full content a sighted reader gets. There's currently no supported way to attach a longer text equivalent (a caption, adjacent summary list, or linked long description) for these cases.

### Notable inconsistency

[...]

PR #816, opened 2026-08-21
Add empty alt="" to imported images missing alt text (#814 short-term…

benbrasso:
## Summary

Short-term fix for #814: inline images added through Word import have no supported way to carry alt text, and Word's conversion silently drops any alt text that did exist. That leaves accessibility gaps invisible in the stored source.

This PR backfills `alt=""` onto any imported `` tag that has no `alt` attribute at all, so the gap is visible and greppable in the stored HTML/markdown instead of silently absent — a normalization step, not a UI change.

## What changed

- Added `add_missing_alt_text_to_imgs(soup)` in `nofos/nofos/nofo.py`.
- Wired it into `process_nofo_html()`, immediately after the existing `normalize_whitespace_img_alt_text()` step, so it runs during Word import.
- Images that already carry an `alt` attribute — including an intentionally empty `alt=""` on a decorative image — are left untouched. Only images missing the attribute entirely are backfilled.

## What this does *not* change

- **Cover images are unaffected.** `Nofo.cover_image` / `Nofo.cover_image_alt_text` are separate model fields, set through their own upload flow, and never pass through the imported-document soup this function operates on. This is covered by an explicit test (`test_does_not_affect_cover_image_markup`).
- **Manually-edited subsections are out of scope.** This only runs on the Word-import path. Applying the same normalization to hand-edited markdown, plus a proper alt-text UI and in-editor warnings for missing alt text, are called out in #814 as longer-term follow-ups and are intentionally left for later.

## Test plan

- Added `AddMissingAltTextToImgsTests` (7 cases): missing alt, existing non-empty alt preserved, existing empty alt preserved, a batch with mixed alt states, a base64-embedded image, no images present, and cover-image markup left untouched.
- `python manage.py test nofos.test_nofo.AddMissingAltTextToImgsTests nofos.test_nofo.NormalizeWhitespaceImgAltTextTests` — 12/12 pass.
- Full suite (`python manage.py test`) — 1448/1448 pass, no regressions, after merging in the latest `main`.

Addresses #814 (short-term scope only — not closing it, since the long-term follow-ups noted above are still open).


benbrasso:
## Finding: this fix currently has no observable effect

While testing this end-to-end (real import → subsection edit page, screenshot attached), I found that `add_missing_alt_text_to_imgs()` doesn't actually change anything visible in the app. Filing this before merging so it gets a look.

**What I did:** imported a small HTML NOFO containing `<p></p>` (no `alt` at all) through the real import flow, then opened the resulting subsection's edit page.

**What showed up in the markdown editor:**
```
![](diagram.png)
```
Not ``. Markdown's `` syntax has no way to represent "alt attribute present but empty" vs. "alt attribute missing" — both collapse to empty brackets.

**Why:** the HTML→Markdown conversion (`markdownify`, via `nofo_markdown.py`) already defaults a missing `alt` to `""` when writing ``, independent of our fix:

```python
>>> markdownify('<img src="diagram.png">')        # no alt at all
'![](diagram.png)'
>>> markdownify('<img alt="" src="diagram.png">')  # our backfilled alt=""
'![](diagram.png)'
```
Identical output either way. `add_missing_alt_text_to_imgs()` mutates the soup *before* this conversion runs, so the distinction it creates is erased by the very next step in the pipeline.

**And it's redundant going the other direction too:** rendering `` back to HTML (for the PDF/view) already produces `alt=""`, with or without this fix:
```python
>>> markdown.markdown('![](diagram.png)')
'<p><img alt="" src="diagram.png" /></p>'
```

So the "make missing alt text visible and greppable" goal from #814 isn't achievable through the current Markdown storage layer — `` already is the greppable signature for missing alt, before and after this PR. As merged, this function is a no-op in production.

**What it would actually take:** `NofoMarkdownConverter` would need a `convert_img` override that emits raw `` HTML (bypassing markdown image syntax) for backfilled images — similar to how it already special-cases `convert_ul`/`convert_ol`/`convert_table` to preserve raw HTML in specific situations. That's a larger change than what's currently in this PR.

Screenshot of the subsection editor showing the actual `` output attached below.

image

benbrasso:
## Follow-up: prompt to fix the finding above

Once reviewed, here's a self-contained prompt for a Claude Code session to implement the actual fix (make missing alt text show up as raw `` HTML in the stored markdown, instead of collapsing into indistinguishable `` syntax).

```
Context: PR #816 (short-term fix for #814) added `add_missing_alt_text_to_imgs()` in
`nofos/nofos/nofo.py`, which backfills `alt=""` onto imported `<img>` tags with no `alt`
attribute. End-to-end testing showed this has NO observable effect: the HTML→Markdown
conversion (`markdownify`, via `nofos/nofos/nofo_markdown.py`) already collapses both
"no alt attribute" and "alt=\"\"" into the same `![](src)` output, and rendering `![]()`
back to HTML always produces `<img alt="" src="...">` regardless of this fix. So the
"make missing alt text visible/greppable" goal from #814 isn't achieved — grepping `![]`
today already finds these images whether or not the backfill function ran.

Task: make missing alt text actually visible and greppable in the STORED markdown source,
by keeping backfilled images as raw HTML instead of letting them collapse into markdown
image syntax — mirroring the existing pattern in `nofos/nofos/nofo_markdown.py` where
`PRESERVE_BOOKMARK_TARGET_ATTR` (`data-nofo-preserve-bookmark-target`) marks elements that
`NofoMarkdownConverter.convert_a` must keep as raw HTML through the HTML→Markdown round trip.

1. In `nofos/nofos/nofo.py`, in `add_missing_alt_text_to_imgs(soup)` (next to
   `normalize_whitespace_img_alt_text`, wired into `process_nofo_html`): when backfilling
   `alt=""` onto an `<img>` missing the attribute, ALSO set a marker attribute — e.g.
   `data-nofo-missing-alt-text="true"` — on that same `<img>` tag. The marker distinguishes
   "we backfilled this because alt was missing" from an img that already had an intentional
   `alt=""` for a decorative image (which should keep converting to ordinary markdown
   `![]()`, since it was never missing).

2. In `nofos/nofos/nofo_markdown.py`, add a `convert_img` override to `NofoMarkdownConverter`
   (alongside the existing `convert_a`, `convert_div`, `convert_ol`, `convert_ul`,
   `convert_table`, `convert_p`, `convert_th` overrides). When the img tag carries the
   marker attribute, strip the marker and return the raw HTML string for the tag (`str(el)`)
   instead of calling `super().convert_img(...)`, so the stored markdown source shows literal
   `<img alt="" src="...">` HTML instead of `![]()`. For any img without the marker, keep the
   default markdown `![alt](src)` behavior via `super().convert_img(el, text, parent_tags)`.

3. Confirm the round trip: raw inline HTML inside markdown source is passed through
   unchanged by the `markdown` library when rendering back to HTML for the PDF/view (the
   same assumption the bookmark-target and heading-div overrides already rely on) — add a
   test confirming `<img alt="" src="...">` in a stored subsection body renders correctly
   through the existing markdown-to-HTML rendering path.

4. Update/add tests:
   - `nofos/nofos/test_nofo.py`: `AddMissingAltTextToImgsTests` should also assert the
     marker attribute is set on backfilled images, and NOT set on images that already had
     a non-missing alt (including an intentional `alt=""`).
   - `nofos/nofos/tests_nofos/test_nofo_markdown.py`: new tests for
     `NofoMarkdownConverter.convert_img` — a marked image renders as raw HTML with
     `alt=""`; an unmarked image (with or without alt) renders as ordinary `![alt](src)`
     markdown; the marker attribute itself never leaks into the final converted output.
   - An end-to-end test (extending the existing HTML-import tests in
     `nofos/nofos/tests_nofos/test_import.py`) confirming a real import of an `<img>` with
     no alt produces a subsection body containing literal `<img alt="" src="...">`, not
     `![](...)`.

5. Run the full test suite (`poetry run python manage.py test` from `nofos/`) and confirm
   no regressions — especially `test_nofo_markdown.py`, `test_import.py`, and anything
   touching table/list HTML-preservation (`convert_table`, `convert_ul`, `convert_ol`),
   since an img inside a table cell needs the same raw-HTML behavior when marked.

Branch: continue on `claude/issue-814-short-term-fix-gn6jla` (already has the initial
backfill commit, merged with upstream `main`). Push to the same fork/branch feeding PR
#816 — do not open a new PR.
```

benbrasso:
## Fix implemented and verified end-to-end

Following up on the earlier finding: backfilled alt text now stays visible as raw HTML instead of collapsing into markdown's indistinguishable `` syntax.

**What changed:**
- `add_missing_alt_text_to_imgs()` in `nofos/nofos/nofo.py` now also tags each backfilled image with a `data-nofo-missing-alt-text` marker (same pattern the codebase already uses for bookmark targets via `PRESERVE_BOOKMARK_TARGET_ATTR`).
- `NofoMarkdownConverter` in `nofos/nofos/nofo_markdown.py` gets a new `convert_img` override: marked images are emitted as raw `` HTML (and the marker is stripped before output); everything else still converts to normal `alt` markdown, including images with an intentional `alt=""` (decorative images), which are correctly left alone.

**Verified with a real import**, through a throwaway local dev instance (not staging/prod — a disposable container clone, its own SQLite DB, torn down after this session): imported an HTML doc with `<p></p>` (no alt at all), then opened the resulting subsection's edit page.

The markdown editor now shows:
```
Some eligibility content, followed by an inline diagram with no alt text at all:

<img alt="" src="diagram.png"/>

End of subsection.
```
instead of the previous (invisible) ``. Screenshot attached below.

---

06subsectioneditfixed

---

**Test coverage added:**
- `AddMissingAltTextToImgsTests` (`test_nofo.py`) — updated to assert the marker attribute is set only on backfilled images, never on images that already had any alt (including intentional `alt=""`).
- `NofoMarkdownConverterIMGTest` (`tests_nofos/test_nofo_markdown.py`) — new: marked image → raw HTML; unmarked image (with or without alt) → normal markdown; marker never leaks into output; marker honored inside table cells too.
- `TestNofoImportMissingAltText` (`tests_nofos/test_import.py`) — new end-to-end test: posts a real HTML import and asserts the saved `Subsection.body` contains literal ``, not ``.
- Full suite: `python manage.py test` — 1674/1674 pass, no regressions.

This is ready to come out of draft once you've had a chance to look.

PR #818, opened 2026-08-21
Add accessible scopes to grouped table headers

### LINKED ISSUE #787: Add automatic `scope` attribute correction for multi-row table headers
### Scenario
NOFO Builder tables can be built with a two-row `<thead>` — a top row of spanning `<th colspan>` group labels (e.g., "Application Due Dates," "Review and Award Cycles") followed by a row of specific column labels (e.g., "New," "Renewal / Resubmission...," "Scientific Merit Review"). Because both rows use `<th>` inside `<thead>`, they render visually identically as headers, which can look like an error but often isn't.

From a WCAG 2.2 AA perspective (Success Criterion 1.3.1, Info and Relationships), this pattern is valid — but only if the header relationships are programmatically determinable via `scope` (or `headers`/`id`) attributes. Currently, tables built in NOFO Builder's editor don't include `scope` attributes on header cells, so a two-row header structure relies on assistive tech correctly inferring relationships from `<thead>`/`<th>` placement alone, which isn't guaranteed across all screen reader/browser combinations.

An example of this table structure in NOFO Builder can be found here: https://nofos.simpler.grants.gov/nofos/fb445e45-ab63-4320-910e-b9839d4bf4e2/section/2cbd7267-b8ed-43ef-a264-7abf5c1236fa/subsection/8dbc4d4b-7d07-451b-b5f4-348e070eaa95/edit

### Potential Solutions

**1. Auto-inject `scope` attributes on save/import (recommended, low risk)**
When NOFO Builder detects a `<thead>` containing a spanning `<th colspan>` row followed by a row of single-column `<th>` cells, automatically add `scope="colgroup"` to the spanning cells and `scope="col"` to the individual cells beneath them. This is purely mechanical — it adds accessibility metadata without altering visible content or table structure, so it can run automatically with no editorial review needed.

**2. Flag ambiguous/irregular header structures for manual review**
For structures that can't be safely auto-corrected — mismatched colspan totals, irregular cell counts across header rows, or headers nested more than two levels deep — surface a warning in the editor (e.g., "This table has a header structure we can't automatically validate — please check manually") rather than guessing or silently doing nothing.

**3. Auto-flattening headers (not recommended as a default)**
Collapsing a two-row header into one row (e.g., combining "Application Due Dates" + "New" into "Application Due Dates: New") requires composing new label text, which is an editorial decision rather than a mechanical fix, and often produces awkward or redundant labels. This could be offered as an optional manual tool, but shouldn't be applied automatically.

### What Needs to Be Done
- Confirm how table HTML is currently generated/stored by NOFO Builder's editor (on save, on import, or both) to determine where `scope`-injection logic should live.
- Build detection logic for the two-row spanning-header pattern (colspan row + single-column row beneath it).
- Implement automatic `scope="colgroup"` / `scope="col"` injection for detected patterns.
- Build detection + a non-blocking editor warning for header structures that don't match the expected pattern (mismatched colspans, irregular row lengths, >2 header row levels).
- Confirm scope of "auto-flatten" as a possible future manual tool, but exclude it from this ticket's automatic behavior.

### Acceptance Criteria
- [ ] Tables with a standard two-row spanning header pattern automatically receive correct `scope="colgroup"` and `scope="col"` attributes on save/import.
- [ ] Auto-injection does not alter visible table content, cell values, or existing colspan/rowspan structure.
- [ ] Tables with irregular/ambiguous header structures trigger an editor warning rather than being silently left as-is or incorrectly auto-corrected.
- [ ] Tables with a single header row are unaffected (no unnecessary attributes added, no false-positive warnings).
- [ ] Test cases cover: a valid two-row spanning header, a single-row header, a header with mismatched colspan totals, and a header nested more than two levels deep.

mikec-ai:
## Summary

- add explicit `scope="colgroup"` and `scope="col"` relationships to unambiguous two-row table headers
- apply the correction during Word import and when Builder subsections are created or saved
- leave single-row, mismatched, row-spanning, and deeper header structures unchanged rather than guessing

The non-blocking editor warning for ambiguous structures is split into follow-up #819 so this mechanical accessibility correction does not introduce an unreviewed warning UX.

## Reproduction and verification

The live example linked from #787 still reproduces the issue: its two-row header contains eight `<th>` elements and zero `scope` attributes.

Locally, the same pattern started with zero scoped headers. Saving the subsection added two `scope="colgroup"` attributes and four `scope="col"` attributes, while the rendered table remained visually unchanged.

## Tests

- `make test` — 1,682 tests passed
- Black, isort, pre-commit hooks, and `git diff --check` passed

Closes #787


mikec-ai:
Locally reproduced and verified against a two-row grouped header matching the live example from #787.

- Before save: 6 header cells, 0 `scope` attributes.
- After save: 2 group headers with `scope="colgroup"` and 4 leaf headers with `scope="col"`.
- The rendered table remained visually unchanged.

The screenshot shows the corrected markup persisted in NOFO Builder after saving the subsection.

NOFO Builder subsection editor showing colgroup and col scope attributes added to a two-row table header

benbrasso (changes requested):
**Pipeline ordering prevents this from firing on real imports**

`add_table_header_scopes()` runs early in `process_nofo_html()`, before the table has been split into `thead`/`tbody`. At that point, mammoth's raw output puts *every* row of a table with header formatting — grouped header rows and all the data rows below them — into a single `thead`, all as `th`, with no `tbody` (this repo's own `test_table_with_complex_html_in_ths` fixture confirms this raw shape).

Since `add_table_header_scopes` requires `thead` to contain **exactly 2** `tr`s, it bails out on any table that has data rows below the two header rows — which is the normal case. It would only fire on a degenerate table that's *just* the two header rows with zero data underneath.

Repro, using the functions as they are in this PR:

Input (2 header rows + 2 data rows, shaped like real mammoth output):
```html
<table><thead>
<tr><th colspan="2">Due dates</th><th colspan="2">Review cycle</th></tr>
<tr><th>New</th><th>Renewal</th><th>Review</th><th>Award</th></tr>
<tr><th>June 1</th><th>June 15</th><th>July 1</th><th>Aug 1</th></tr>
<tr><th>Dec 1</th><th>Dec 15</th><th>Jan 1</th><th>Feb 1</th></tr>
</thead></table>
```
1. `add_table_header_scopes(soup)` sees 4 rows in `thead`, not 2 → skips → `updated_tables == 0`.
2. Downstream, the existing `convert_table_with_all_ths_to_a_regular_table` keeps only row 0 as the header and demotes everything else (rows 1–3) to `td` in a new `tbody`. That includes the "New/Renewal/Review/Award" row, which was meant to be the second header row — it loses its `th` entirely, not just a `scope` attribute.

Final result: only one header row survives, no scopes get added, and the second header row's semantics are lost — for what should be exactly the unambiguous case this PR targets.

**Suggested fix**

Reordering the calls alone won't fix it, because by the time `convert_table_with_all_ths_to_a_regular_table` finishes, the second header row is already `td`. The two need to cooperate:

- Before demoting rows, check whether row 0 + row 1 of `thead` match the unambiguous grouped-header pattern (the same check `add_table_header_scopes` already does).
- If they match: only demote rows **2 onward** (not **1 onward**) to `tbody`/`td`, and apply `scope="colgroup"`/`scope="col"` to rows 0 and 1 as part of that same step.
- If they don't match: fall back to the current behavior unchanged.

Happy to put up a patch for this if useful — wanted to flag it here first since it changes `convert_table_with_all_ths_to_a_regular_table`, which already has its own test coverage.

---

mikec-ai:
Addressed the pipeline-ordering issue you found.

The grouped-header pattern is now detected inside the existing all-`th` table conversion. When the first two rows unambiguously form grouped headers, Builder keeps both in `thead`, applies `scope="colgroup"` and `scope="col"`, and moves only rows 3 onward into `tbody` as regular data cells. Ambiguous tables retain the existing one-header-row fallback.

I added the four-row Mammoth-shaped regression from your review and confirmed that the early scope pass skips it while the converter finishes it correctly.

Validation:
- 578 NOFO utility tests pass
- 14 focused table tests pass
- Black, isort, and pre-commit checks pass

@benbrasso ready for another look.

nofos/nofos/forms.py
```diff
@@ -296,6 +297,9 @@ def clean(self):
 
         return cleaned_data
 
+    def clean_body(self):
+        return add_table_header_scopes_to_html(self.cleaned_data.get("body", ""))
```
benbrasso, line 301: `_clean_fields` only catches `forms.ValidationError` from `clean_<field>`. Any other exception raised inside `add_table_header_scopes_to_html` (e.g. an edge case in the bs4 parsing on hand-edited HTML) will propagate uncaught and turn a normal save into a 500 instead of a graceful validation error — previously `clean_body` couldn't fail this way at all. Since adding scopes is a non-essential enhancement, it shouldn't be able to block or crash the save.

```suggestion
    def clean_body(self):
        body = self.cleaned_data.get("body", "")
        try:
            return add_table_header_scopes_to_html(body)
        except Exception:
            logging.getLogger(__name__).exception(
                "Failed to add table header scopes"
            )
            return body
```

(Same issue applies to `SubsectionCreateForm.clean_body` below — and this'll need `import logging` added at the top of the file.)

nofos/nofos/nofo.py
```diff
@@ -743,21 +794,72 @@ def convert_table_with_all_ths_to_a_regular_table(table):
     thead = table.find("thead")
```
benbrasso, line 794: `_get_unambiguous_grouped_header_rows` looks up `thead` with `table.find("thead", recursive=False)` (direct child only), but this uses `table.find("thead")` (recursive — can match a `thead` nested inside a nested table in a cell). If the recursive lookup here finds a deep `thead` with `len(rows) > 1`, this function enters the grouped-header branch, but the helper's direct-child lookup will return `None` for that same table, so it silently falls through to "first row only" instead of being scoped. Worth making the two consistent.

```suggestion
    thead = table.find("thead", recursive=False)
```

benbrasso (commented):
Reviewed the new table-header-scope logic. Left two inline suggestions for issues that could cause real problems (an uncaught-exception path that could 500 a normal subsection save, and a `thead` lookup depth mismatch between two functions that can silently skip scoping). Also left a top-level comment for @mikec-ai about a trickier nested-table edge case in `add_table_header_scopes_to_html` where I think the fix approach is worth a design call rather than a quick suggestion.

benbrasso:
@mikec-ai one more thing I wanted to flag separately since I think it deserves your call on the approach rather than a quick suggested-change:

In `add_table_header_scopes_to_html` (`nofos/nofos/nofo.py`), the regex used to pull table blocks out of the mixed markdown/HTML body is non-greedy:

```python
table_pattern = re.compile(r"<table\b[\s\S]*?</table>", re.IGNORECASE)
```

For a table nested inside another table's cell (`<table><tr><td><table>...</table></td></tr></table>`), this matches only up through the *inner* `</table>`. `add_scopes` then parses that truncated fragment with BeautifulSoup — which silently auto-closes it — and splices the reconstructed HTML back into the string, leaving the leftover `</td></tr></table>` dangling. The result is malformed/duplicated markup getting saved as the subsection body.

I don't think this is likely to hit through normal NOFO imports, but it's reachable any time a user hand-edits a subsection body to include a nested table in the Martor editor, so it seems worth closing before this ships.

A couple of directions that would fix it, happy to help however's useful:
- Track `<table`/`</table>` nesting depth while scanning the string to find the true end of each *top-level* table, instead of a single regex pass.
- Or skip the regex entirely: parse `value` once with BeautifulSoup, run the existing `add_table_header_scopes(soup)` over the whole tree, and reserialize — this reuses logic you already wrote and sidesteps string-slicing tables out entirely. The tradeoff is it would also reserialize (e.g. normalize quote style on) any table content outside the ones actually modified, which the current regex approach avoids — so this only works if that's an acceptable tradeoff for this codebase.

Let me know which way you'd rather go, or if you'd rather I take a pass at a patch.

benbrasso (changes requested):

benbrasso:
putting this to draft for the moment; some comments might have went beyond what we were wanting to quickly accomplish in this PR

PR #820, opened 2026-08-21
Make the Import NOFO action easier to find

### LINKED ISSUE #784: Import NOFO action is hard to discover for new users when NOFO list is long
## Scenario
After logging into NOFO Builder, users land on the In Progress NOFOs list. For new users, importing a new NOFO is often the very first and most important action they need to take, but the Import button sits at the very bottom of the list. When a workspace already has a long list of existing NOFOs, that button falls well outside the initial viewport, making it hard for first-time users to find. Experienced users generally know where to scroll to find it, but new users don't have that context.

## Background
The Import NOFO button is currently the only entry point for starting a new NOFO, and it's located at the bottom of the NOFO list page. As adoption grows, the number of NOFOs per workspace will only increase, making this discoverability problem worse over time.

## How to Reproduce
1. Log into NOFO Builder with an account that has a long list of existing NOFOs.
2. Land on the In Progress NOFOs page.
3. Observe that the Import NOFO action is not visible without scrolling to the bottom of the list.

## Investigation Notes / Possible Directions
- One option worth exploring is adding a persistent or sticky Import NOFO action near the top of the page, or in a fixed header/toolbar, so it's always visible regardless of list length.
- Another option is duplicating the Import action in two places — keeping it at the bottom for existing users' muscle memory, while adding a second entry point near the top of the page for first-time visibility.
- These are just starting points; the final direction should come out of design investigation, not be treated as prescribed.

## Acceptance Criteria
- [ ] The Import NOFO action is visible within the initial viewport regardless of how many NOFOs exist in the list.
- [ ] The solution accounts for both first-time users unfamiliar with the tool and existing users accustomed to the current button location.
- [ ] Design investigation and mockups are completed and reviewed before implementation.
- [ ] The chosen solution does not degrade the experience for users with short NOFO lists.

mikec-ai:
## What changed

- Added an **Import NOFO** button to the page header beside **Search NOFOs**, so it remains visible even when the NOFO list is long.
- Kept the existing import action at the bottom of the page for current users.
- Adjusted the header actions for narrow screens so they move below the page title instead of overlapping it.
- Added regression coverage confirming both import entry points remain available.

## Why

New users often need to import a NOFO first, but the existing action can sit far below the initial viewport when the workspace contains many records.

## Validation

- Focused Django view test passes.
- Black, isort, and repository pre-commit checks pass.
- Ben approved the additive header placement in #784.

## Screenshot

Import NOFO action in the page header beside Search NOFOs

Closes #784

### DIFF
--- a/nofos/bloom_nofos/static/styles.css
+++ b/nofos/bloom_nofos/static/styles.css
@@ -751,7 +751,20 @@ label.usa-label {
 }
 
 .nofo_index--header--view {
-  top: 10px;
+  top: 4px;
+  align-items: center;
+}
+
+@media all and (max-width: 39.99em) {
+  .nofo_index--header h1 {
+    width: 100%;
+  }
+
+  .nofo_index--header--view {
+    position: static;
+    justify-content: flex-start;
+    margin-top: 0.75rem;
+  }
 }
 
 .nofo_edit--audit-widget {
--- a/nofos/nofos/templates/nofos/nofo_index.html
+++ b/nofos/nofos/templates/nofos/nofo_index.html
@@ -20,6 +20,7 @@
 		</h1>
 		<div class="nofo_index--header--view font-sans-md">
 			<a class="search-nofos-link" href="{% url 'nofos:nofo_search' %}">Search NOFOs</a>
+			<a class="usa-button margin-0" href="{% url 'nofos:nofo_import' %}">Import NOFO</a>
 		</div>
 	</div>
 
--- a/nofos/nofos/tests_nofos/test_index.py
+++ b/nofos/nofos/tests_nofos/test_index.py
@@ -1,0 +1,27 @@
+from django.contrib.auth import get_user_model
+from django.test import TestCase
+from django.urls import reverse
+
+User = get_user_model()
+
+
+class NofoIndexViewTests(TestCase):
+    def setUp(self):
+        self.user = User.objects.create_user(
+            email="index-test@example.com",
+            password="testpass123",
+            group="bloom",
+            force_password_reset=False,
+        )
+        self.client.login(email="index-test@example.com", password="testpass123")
+
+    def test_import_action_appears_at_top_and_bottom_of_page(self):
+        response = self.client.get(reverse("nofos:nofo_index"))
+
+        self.assertEqual(response.status_code, 200)
+        self.assertContains(
+            response,
+            f'href="{reverse("nofos:nofo_import")}"',
+            count=2,
+        )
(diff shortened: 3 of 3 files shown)


benbrasso:
## Changes requested: header actions overlap the page title at medium widths

Great feature, and the two-entry-point approach (header + bottom section) is the right call. But I found a real visual regression when actually running this branch and checking it across viewport widths, not just at desktop and mobile phone sizes.

**What's happening:** `.nofo_index--header--view` (the box holding "Search NOFOs" + the new "Import NOFO" button) is absolutely positioned into the right ~35% of the header (`.nofo_index--header h1 { width: 65% }` / `.nofo_index--header--view { position: absolute; right: 0 }`). That split was tuned when the box only held one short link. Now that it holds two items, the box is wider, but the heading's reserved width wasn't adjusted to match.

The new mobile media query (`max-width: 39.99em`, ~640px) correctly stacks everything below that width. But **between roughly 640px and 768px, "Search NOFOs" visibly overlaps the tail of the h1 title.** Screenshots attached:
- 660px wide: "Search NOFOs" sits directly on top of the heading text.
- 700px wide: still touching/overlapping.
- 768px wide: clean, no overlap.

This isn't an edge case — it reproduces on the default "Bloom in progress NOFOs" view (the first screen most users land on) at very ordinary widths: a non-maximized laptop window, a tablet, a split-screen browser.

**Why the tests didn't catch it:** `test_import_action_appears_at_top_and_bottom_of_page` only checks that the import link's markup appears twice in the response HTML. Django tests don't render CSS, so this class of layout bug isn't something that test (or any test like it) can catch — worth knowing for future UI changes here, not a knock on this test.

**Suggested fix:** either widen the mobile breakpoint to cover this range (e.g. `39.99em` → something like `48em`/768px), or move away from the hardcoded `65%` / absolutely-positioned-box split toward a flex layout that shares space between the title and the action row dynamically — that would avoid re-tuning a magic percentage every time an item gets added to the header actions.

benbrasso:
Simplest fix: widen the mobile breakpoint so it covers the range where the overlap happens (roughly 640–768px), rather than re-tuning the desktop `65%`/absolute-position split every time an item is added to the header actions.

```suggestion
@media all and (max-width: 47.99em) {
```

Verified by running this branch locally and screenshotting every ~50–100px step from 375px to 1024px with this one-line change — clean at all of them, including right at the old breakpoint (767/768/769px) and the previously-broken widths (660px, 700px). Test suite (`test_import_action_appears_at_top_and_bottom_of_page`) still passes.

PR #834, opened 2026-08-25
feat: persist readability metric snapshots for NOFOs

### LINKED ISSUE #831: Record readability metric history for NOFOs
## Problem

The initial readability integration in #797 calculates metrics on demand and returns them only to the current browser session. Results are not persisted, so Builder cannot retain the latest successful measurement or later show how a NOFO changed over time. PR #822 made the feature flag runtime-configurable but did not change this storage boundary.

## Desired outcome

Persist successful readability calculations as durable, append-only snapshots linked to the NOFO. Builder should be able to identify the latest valid snapshot while retaining earlier measurements for future trend analysis.

This should establish a history-friendly data contract without requiring a historical chart or portfolio dashboard in the first implementation.

## High-level acceptance criteria

- A successful readability calculation can create a durable snapshot associated with the correct NOFO.
- A later calculation does not overwrite or destroy earlier snapshots.
- Each snapshot records when it was calculated, the NOFO revision it measured, all returned metric values and statuses, and enough measurement-contract provenance to interpret it later (for example package version, profile, and result schema version).
- Builder can retrieve the latest successful snapshot efficiently while preserving the existing permission boundary for the NOFO.
- Failed or unavailable calculations do not replace the last successful snapshot with incomplete metric data.
- The behavior for repeated calculations of the same NOFO revision and measurement contract is explicitly defined and covered by tests.
- Database migrations, model constraints/indexes, and focused persistence tests are included.
- Existing on-demand readability behavior continues to work while persistence is introduced.

## Design questions to resolve

- Should repeated successful calculations of the same NOFO revision and measurement contract create separate events, or be deduplicated?
- Should frequently queried metrics use normalized columns, a versioned result payload, or a hybrid that preserves the package contract while supporting efficient queries?
- Should snapshot creation happen in the current request path initially, or behind a service boundary that can later support save-time/background calculation?
- What retention and deletion behavior should apply when a NOFO is archived or deleted?

## Scope boundary

This issue covers durable readability snapshots and latest-result retrieval. Historical charts, portfolio dashboards, scheduled/background recalculation, backfilling existing NOFOs, and policy decisions based on score changes should be tracked separately unless needed to validate the persistence contract.

dfitchett:
Closes #831.

Builds on the on-demand integration from #797 and the runtime flag from #822. Readability results are now stored as durable, append-only snapshots instead of living only in the browser session, so Builder can retain the latest successful measurement and keep earlier ones for later trend work.

Snapshots are written **only on demand**: opening the readability panel starts a calculation, and the button allows retries/recalculation. Save-time and background triggers are deliberately out of scope.

## What changed

- Add a `NofoReadabilityScore` model and migration `0132`: one append-only snapshot per NOFO revision per measurement contract.
- Add `record_readability_snapshot()` as a service boundary in `readability.py`. The endpoint calls it instead of calling the package directly.
- Look up before measuring: an already-measured revision replays its stored snapshot without re-running the package.
- Add `latest_for()`, `current_for()`, and `complete()` for retrieval, backed by a partial index.
- Exclude the model from `easyaudit`, which would otherwise log a record of a record.
- Add 16 persistence tests.

The endpoint's response contract is unchanged: the package payload verbatim, with `Cache-Control: no-store`.

## The table

`nofos_noforeadabilityscore`

| Column | Type | Description |
| --- | --- | --- |
| `id` | `BigAutoField` | Primary key. Monotonic, so it also breaks ties between snapshots written in the same instant. |
| `nofo` | `FK(Nofo)`, `CASCADE` | The NOFO measured. Deleting a NOFO deletes its history; archiving is a soft delete and leaves snapshots intact. |
| `created_at` | `DateTimeField`, `auto_now_add` | When the calculation ran. |
| `created_by` | `FK(User)`, `SET_NULL` | Who ran it. Null once the user is deleted. |
| `nofo_revision` | `DateTimeField` | `Nofo.updated` at calculation time — the content revision these metrics describe, and the same value passed to the package as `revision`. **The load-bearing column:** a snapshot is current while this matches `Nofo.updated`. |
| `profile_reference` | `CharField(128)` | Metrics profile used, e.g. `hhs-nofo-fy27-html@0.4.0`. |
| `package_version` | `CharField(32)` | Installed `hhs-nofo-metrics` version, e.g. `0.5.2`. |
| `schema_version` | `CharField(32)` | `result["schema_version"]`, e.g. `1.1.0`. The result *contract* version, which moves independently of the package and profile. |
| `result_basis` | `CharField(64)` | `result["result_basis"]`, e.g. `structured_estimate`. How the measurement was derived. |
| `is_complete` | `BooleanField` | True when every metric the package returned had status `calculated`. False snapshots are kept but never serve as the latest successful measurement. |
| `result` | `JSONField` | The package's complete `AnalysisResult`, unflattened: `schema_version`, `result_basis`, `source`, `coverage`, `metrics`, `notes`. |
| `goals` | `JSONField` | Normalized `HHS_NOFO_METRIC_GOALS` in effect at calculation time, recorded so a snapshot stays interpretable. Live comparisons render from current config, not this copy. |

**Constraint** — `unique_readability_score_per_nofo_revision` on `(nofo, nofo_revision, profile_reference, package_version)`. This is what makes the table a cache rather than just a log.

**Index** — `readability_latest_ok_idx` on `(nofo, -created_at) WHERE is_complete`, for `latest_for()`.

### A note on `is_complete`

It is judged against the metrics the package actually returned, **not** against `GOAL_METRIC_IDS`. Those are different sets, and conflating them is a live bug hazard: `GOAL_METRIC_IDS` is the set Builder allows *goals* to be configured for, while v0.5.2 returns `characters_per_word` and does not return `sentences_per_paragraph`. Deriving completeness from the goals list marks every real snapshot incomplete, which makes `latest_for()` return `None` forever. `test_a_metric_builder_configures_goals_for_but_the_package_omits_is_not_missing_data` pins this.

## Why

The score is a pure function of the content revision and the measurement contract, so keying on `(nofo, nofo_revision, profile_reference, package_version)` makes a stale snapshot structurally impossible — there is no hook to forget. `Nofo.updated` is already maintained across section and subsection edits by `touch_updated()`, and was already passed to the package as `revision`, so the version key existed before this change.

Writing on calculation rather than on save also keeps a whole-document render out of the edit path. `touch_updated()` bypasses signals and imports use `bulk_create`, so a save-time trigger would need explicit call sites at every mutation point. Putting creation behind a service function means that trigger can be added later without touching the model.

## Design questions from the ticket

**Repeated calculations of the same revision and contract** are deduplicated. The unique constraint enforces it, and recalculating an unchanged NOFO returns the stored snapshot without re-running the package.

**Storage shape is hybrid.** The payload is stored whole, preserving the package contract including per-metric statuses and notes. Only revision, provenance, and `is_complete` are promoted to indexed columns. Metric values are not denormalized — charts are out of scope, and Postgres expression indexes can cover them when that work lands.

**Snapshot creation runs in the request path**, behind `record_readability_snapshot()` so save-time or background calculation can reuse it.

**Retention:** delete cascades, matching the existing `BaseNofo.successor` behavior. Archive leaves snapshots intact. Both are pinned by tests.

## Verify locally

Requires the `hhs-nofo-metrics` wheel from `pyproject.toml` to be installed in your active environment.

```bash
poetry install
poetry run python nofos/manage.py migrate
HHS_NOFO_METRICS_ENABLED=true poetry run python nofos/manage.py runserver
```

Save the inspector below as `nofos/inspect_snapshots.py` and run it from `nofos/` after each step:

```bash
poetry run python inspect_snapshots.py
```

| Step | Expected |
| --- | --- |
| 1. Open a populated NOFO, expand **Readability metrics — Beta** (this starts the calculation) | One row, marked `[CURRENT, complete, <- latest_for]`, with your user, the revision, the contract, and 6 goals recorded |
| 2. Click **Recalculate** without editing anything | Still one row and `created_at` is **unchanged** — the package did not re-run |
| 3. Edit a subsection, reload the edit page | `current_for()` reports `MISS` — the revision key tracked the body edit |
| 4. Recalculate | Two rows. The new one is `CURRENT`; the earlier one is retained at its own revision |
| 5. `poetry run python nofos/manage.py test nofos.tests_nofos.test_readability_metrics` | Focused endpoint, real-package, and persistence tests pass; the real-package test is skipped when the package is absent |

Step 2 is the caching claim proving itself: a fresh response with an unchanged `created_at`.

Failure paths (503 and 422 write no row; an incomplete result never becomes `latest_for`) are impractical to trigger by hand and are covered by tests instead.

### Actual output after steps 1–4

```
package version : 0.5.2
profile         : hhs-nofo-fy27-html@0.4.0
total snapshots : 2

Centers for Independent Living Competition—Kansas  (23876e97-...)
  current revision : 2026-08-25T21:45:57.243828+00:00
  current_for()    : HIT
  latest_for()     : 2026-08-25 21:46:03.020412+00:00

  #2  2026-08-25 21:46:03  [CURRENT, complete, <- latest_for]
      measured revision : 2026-08-25T21:45:57.243828+00:00
      contract          : hhs-nofo-fy27-html@0.4.0 / pkg 0.5.2 / schema 1.1.0
        characters_per_word              4.76
        flesch_kincaid_grade_level       2.66
        flesch_reading_ease              88.26
        passive_sentence_percentage      80.0
        word_count                       291
        words_per_sentence               6.9

  #1  2026-08-25 21:24:25  [INCOMPLETE]
      measured revision : 2025-08-06T19:52:10.596301+00:00
      contract          : hhs-nofo-fy27-html@0.4.0 / pkg 0.5.2 / schema 1.1.0
        ...
```

Snapshot #1 predates the `is_complete` fix described above, which is why it reads `INCOMPLETE` despite every metric being calculated.

<details>
<summary><code>nofos/inspect_snapshots.py</code></summary>

```python
"""Show the readability snapshots stored for each NOFO.

Save as nofos/inspect_snapshots.py and run it from the nofos/ directory:

    python inspect_snapshots.py
"""

import os
import sys

import django

sys.path.insert(0, os.getcwd())
os.environ.setdefault("DJANGO_SETTINGS_MODULE", "bloom_nofos.settings")
django.setup()

from nofos.models import Nofo, NofoReadabilityScore  # noqa: E402
from nofos.readability import (  # noqa: E402
    PROFILE_REFERENCE,
    get_metrics_package_version,
)

version = get_metrics_package_version()
print(f"\npackage version : {version or '(not installed)'}")
print(f"profile         : {PROFILE_REFERENCE}")
print(f"total snapshots : {NofoReadabilityScore.objects.count()}\n")

for nofo in Nofo.objects.filter(readability_scores__isnull=False).distinct():
    print("=" * 78)
    print(f"{nofo.short_name or nofo.title}  ({nofo.id})")
    print(f"  current revision : {nofo.updated.isoformat()}")

    current = NofoReadabilityScore.objects.current_for(nofo, PROFILE_REFERENCE, version)
    latest = NofoReadabilityScore.objects.latest_for(nofo)
    print(f"  current_for()    : {'HIT' if current else 'MISS (would recalculate)'}")
    print(f"  latest_for()     : {latest.created_at if latest else 'none complete'}")
    print()

    for score in nofo.readability_scores.all():
        marks = ["CURRENT"] if score.is_current else []
        marks.append("complete" if score.is_complete else "INCOMPLETE")
        if latest and score.pk == latest.pk:
            marks.append("<- latest_for")

        print(
            f"  #{score.pk}  {score.created_at:%Y-%m-%d %H:%M:%S}  [{', '.join(marks)}]"
        )
        print(f"      measured revision : {score.nofo_revision.isoformat()}")
        print(
            f"      contract          : {score.profile_reference} / "
            f"pkg {score.package_version or '(none)'} / schema {score.schema_version}"
        )
        print(f"      by                : {score.created_by or '(none)'}")
        print(f"      goals recorded    : {len(score.goals)} metrics")

        metrics = score.result.get("metrics", {})
        for metric_id, metric in sorted(metrics.items()):
            status = metric.get("status")
            shown = metric.get("value") if status == "calculated" else f"({status})"
            print(f"        {metric_id:<32} {shown}")
        print()

if not NofoReadabilityScore.objects.exists():
    print("No snapshots stored yet.\n")
```

</details>

## Copy and documentation aligned before merge

Updated the panel, Constance flag description, and readability documentation to describe saved, revision-scoped snapshots. Documented POST + CSRF, database-backed reuse, reimport/edit invalidation, and retention. Earlier snapshots are retained but **not displayed in this panel**; adding a history/latest-snapshot UI remains out of scope.

## Not included

- Historical charts, portfolio dashboards, and trend analysis.
- Backfilling snapshots for existing NOFOs.
- Save-time or background recalculation.
- Persisting `clearance_metrics` on the policy-language export view, which still calculates on render.

## Final local verification (2026-09-03)

Updated with current `main`, including pypdf 6.16.1. The full local suite passes: **1,843 tests**, using the pinned hhs-nofo-metrics 0.5.2. Migration consistency and pre-commit checks pass. The earlier note about a policy-language export test failing on main is superseded by this clean run.

Browser smoke testing on an isolated local database verified first calculation, deduplication on recalculation/reload, subsection-edit invalidation, and actual Word reimport. Snapshot counts progressed 1 → 1 → 2 → 3; earlier rows remained unchanged. The list-only Word fixture produced unavailable sentence-based metrics, retained as an incomplete snapshot without replacing the latest complete measurement.

No deployed feature flags were changed.

---

🤖 Generated with [Claude Code](https://claude.com/claude-code)


mikec-ai (changes requested):
I found two issues to address before merging:

1. **Blocking: re-import can silently replay stale metrics.** The cache treats Nofo.updated as the content identity, but overwrite_nofo() replaces sections and subsections through bulk operations without necessarily advancing that timestamp. I reproduced a complete body replacement where Nofo.updated stayed identical. A subsequent calculation would therefore return the prior snapshot without analyzing the revised document. Please explicitly advance the revision for re-import, add a regression test for this path, and guard against the revision changing while analysis is running.

2. **Use POST for snapshot creation.** The readability endpoint now creates durable database state but remains a GET. Please change the mutating request to POST with CSRF protection and update the browser call and tests accordingly.

The overall persistence design looks strong: append-only snapshots, full result and provenance retention, database-enforced deduplication, incomplete-result handling, permissions, and retention behavior are all sensible. I also verified the 27 focused tests, migration consistency, formatting checks, and current CI are green.

mikec-ai (approved):
Both requested changes are addressed.

- Re-import now explicitly advances `Nofo.updated`, with regression coverage proving a replaced body cannot replay the previous snapshot.
- Snapshot creation records the starting revision and declines to persist if the database revision changes during analysis.
- The mutating endpoint is now POST-only, the browser sends the page's CSRF token, and GET is rejected with 405.

Verification:

- 33 focused readability/re-import tests pass;
- migration consistency check passes;
- Black passes on all touched Python files;
- I synthetically merged the current `main` tip into the PR and reran the focused checks successfully;
- the full combined NOFO suite passes: 1,361 tests.

No remaining blocking findings from my review.

benbrasso:
@mikec-ai bump

mikec-ai:
@benbrasso @dfitchett — I brought this up to date with main (including pypdf 6.16.1), fixed the stale persistence/POST documentation and panel/flag copy, and updated the PR title for release notes. No history UI or other behavior was added.

Final local verification passed with hhs-nofo-metrics 0.5.2:

- Full suite: 1,843 tests passed; migration consistency and pre-commit checks passed.
- Actual browser smoke test on an isolated local database: opening the panel saved snapshot #1; recalculation reused the same row and timestamp; editing a subsection created #2; reimporting the repository’s lists.docx through the upload/confirmation flow created #3. Earlier snapshots stayed unchanged, and a reload reused #3.
- Word counts changed 27 → 38 → 159 as the content changed. The list-only reimport correctly showed unavailable sentence-based metrics, stored an incomplete snapshot, and retained #2 as the latest complete measurement.
- No browser console errors. No dev/production data or feature flags were changed.

Screenshot below shows the edited revision with saved-snapshot copy and calculated metrics. All six GitHub checks passed, including unit tests and Anchore/Trivy/Dockle scans. Squash-merged as `479bc76fc8d5f4863df28c4e46835a9af5fb9207`. Deployment and feature-flag activation remain separate.

readability-after-edit

mikec-ai (approved):
Final review of 5fe0625fd37155110bc918e771cf1e6370e5dbc4: no remaining blocking findings. Current main is incorporated, persistence/POST copy is aligned, all 1,843 local tests pass, and the actual browser calculation/edit/Word-reimport/reload loop is verified with database evidence. See the verification comment and screenshot. Merge remains contingent on green GitHub CI.

PR #842, opened 2026-08-27
Add "Add end notes" action to NOFO actions menu

benbrasso:
## Summary
- Adds an "Add Endnotes" item to the NOFO actions dropdown, shown under the same status rules as Find & Replace / Compare / Duplicate NOFO (visible while a NOFO is still editable; hidden once published or cancelled), and hidden once the NOFO already has an Endnotes section (matched by `html_id == "endnotes"`).
- Selecting it opens a dedicated page — not the generic "create subsection" page — that creates the NOFO's "Endnotes" section (fixed `name`/`html_id`, not user-editable) together with one initial subsection, prepopulated with placeholder endnote HTML (two example references) that the user can edit or just save as-is.
- New sections are always appended at the end of the NOFO, or slotted in directly above an existing "Modifications" section so Modifications stays last.
- This intentionally isn't a general "create section" flow — the section's name/html_id are fixed server-side, not accepted from the request, so it can only ever create this one specific section.

## Acceptance criteria / how to test
1. Open an existing NOFO for editing and click **NOFO actions**.
2. **"Add Endnotes" should only appear when:**
   - the NOFO's status allows editing (same rule as Find & Replace / Compare / Duplicate NOFO — hidden for `published` and `cancelled`), **and**
   - the NOFO doesn't already have a section with html_id `endnotes`.
3. Selecting "Add Endnotes" navigates to a page specific to this action (name/html_id are fixed, not editable).
4. The content editor on that page is prepopulated with a two-item placeholder endnotes list (HTML `<ol>` with two example references/backlinks).
5. Clicking **Save** with no changes creates the section successfully.
6. The new "Endnotes" section is added as the last section in the NOFO — or, if a "Modifications" section exists, directly above it, so Modifications stays last.
7. In the rendered NOFO view, since Endnotes doesn't have its own section page and its heading renders as `<h2>`, it shows up in the Table of Contents as a sub-page-style entry (indented, no icon) — consistent with how Modifications/References/Appendix/Glossary already behave — appearing right after whichever section precedes it (typically directly under "Contacts & support" in a standard NOFO).

## Test plan
- [x] Added/updated unit tests for the action-link visibility rules (`TestBuildNofoActionLinks`) and the new view (`test_section_add_end_notes.py`)
- [x] Full `nofos` app test suite passes (1320 tests)
- [x] `black`/`isort` clean on all changed files

### DIFF
--- a/nofos/nofos/forms.py
+++ b/nofos/nofos/forms.py
@@ -327,6 +327,18 @@ class SubsectionCreateForm(forms.ModelForm):
         return cleaned_data
 
 
+# Used only by NofoAddEndNotesSectionView. The Endnotes section's name and
+# html_id are fixed (not user-editable) so the ".section--endnotes" CSS hook
+# and the "Endnotes" import-detection convention keep working; the only
+# thing a user can change is the initial subsection's content.
+class EndNotesSectionCreateForm(forms.ModelForm):
+    body = MartorFormField(required=False, label="Subsection content")
+
+    class Meta:
+        model = Subsection
+        fields = ["body"]
+
+
 # Simple form for URL input
 class CheckNOFOLinkSingleForm(forms.Form):
     url = forms.URLField(
--- a/nofos/nofos/nofo.py
+++ b/nofos/nofos/nofo.py
@@ -1361,6 +1361,41 @@ def _update_link_statuses(all_links):
                 print(f"Error checking link {link['url']}: {e}")
 
 
+END_NOTES_SECTION_NAME = "Endnotes"
+END_NOTES_SECTION_HTML_ID = "endnotes"
+
+# Matches the "docx-style" endnote id/href convention this app already
+# recognizes elsewhere (see get_footnote_type in templatetags/utils), so the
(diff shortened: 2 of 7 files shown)


mikec-ai:
I ran a full review, reproduced the edge cases, and pushed a focused follow-up commit that:

- detects existing Endnotes by the canonical `html_id="endnotes"`;
- strictly blocks the direct action for published NOFOs, including those with modifications;
- checks group access before revealing Endnotes state;
- serializes and rechecks creation to prevent concurrent duplicates;
- safely shifts occupied section orders when inserting before Modifications; and
- adds adversarial regression tests for each case.

Verification on the updated branch: 23 focused tests, 1,326 NOFO tests, and the full 1,787-test application suite passed; formatting and commit hooks passed. I’m doing a fresh independent review of the updated merge result next.

mikec-ai (approved):
Verified the updated implementation independently after the follow-up fixes. Endnotes identity, strict published-state handling, authorization ordering, concurrent duplicate prevention, and collision-safe section ordering all check out. The focused, NOFO, and full application suites pass, and all GitHub checks are green.

PR #843, opened 2026-08-27
Fix checkbox glyph mismatch breaking application-checklist table layout

benbrasso:
## Problem

Some application-checklist tables were rendering with checkbox glyphs
floating outside their table cells in both the HTML view and the generated
PDF, instead of appearing as a properly aligned checkbox icon next to the
form name.

## Root cause

The codebase uses two different unicode characters to represent a
checkbox glyph in NOFO content:

- `◻` U+25FB (WHITE MEDIUM SQUARE)
- `☐` U+2610 (BALLOT BOX)

Several places in the code already treat these as interchangeable (e.g.
`has_checkbox()` and the `.docx` import normalization in `replace_chars()`
in `nofo.py`), but the actual icon-rendering filter,
`replace_unicode_with_icon`, only matched `◻`. When a subsection's stored
markdown contained a literal `☐` instead, the filter never found it, so:

- No SVG checkbox icon was inserted.
- None of the layout classes (`usa-icon__td`, `usa-icon__line`,
  `usa-icon__td--multi-block`, etc.) were applied to the cell.
- The raw `☐` character was left as plain text with no flex-layout
  wrapper, which is what caused it to render outside the table's normal
  column flow.

`.docx` uploads already get normalized to `◻` on import via
`replace_chars()`, but that normalization was only wired into the import
pipeline — not into the manual subsection editor — so a subsection edited
or pasted by hand could end up storing a raw `☐` that bypassed
normalization entirely.

## What we looked at

Before making the change, we audited every file in the codebase
referencing either character (7 files, all under `nofos/nofos/` —
templatetag code and its test suite) to check for unintended side
effects:

- `has_checkbox()` already treats `☐` and `◻` as equivalent for detection,
  with an existing test covering that.
- No existing test asserts that `☐` must pass through the icon-rendering
  filter *unrendered* — nothing currently locks in the old behavior.
- All other tests around this filter use `◻` consistently, so the fix is
  additive and doesn't change behavior for existing content.

## Solution

1. **`nofos/nofos/templatetags/replace_unicode_with_icon.py`**
   - Added `☐` (U+2610) to the `ICONS` list so it renders the same SVG
     checkbox icon as `◻`, with the same layout classes applied.
   - Updated `is_numbered_sublist()` to strip both checkbox characters
     before matching, so numbered-sublist detection stays correct
     regardless of which glyph is present.

2. **`nofos/nofos/forms.py`**
   - Added a `clean_body()` step to `SubsectionEditForm` and
     `SubsectionCreateForm` that runs the existing `replace_chars()`
     normalization (already used for `.docx` imports) on manually
     edited/pasted subsection content, so `☐` (and other checkbox-like
     characters `replace_chars()` already handles) gets normalized to `◻`
     at save time instead of being stored raw.

## Testing

- Verified via a standalone script exercising
  `replace_unicode_with_icon` directly: a table cell containing `☐`
  now produces the identical rendered HTML (icon + `usa-icon__td` class +
  flex wrapper div) as the existing `◻` case.
- Verified `is_numbered_sublist()` returns `True` for numbered items
  prefixed with either `◻` or `☐`.
- Confirmed no circular import is introduced by `forms.py` importing
  `replace_chars` from `nofo.py` (`nofo.py` does not import `forms.py`,
  directly or transitively).
- Reviewed the full existing test suite for this filter
  (`test_templatetags.py`, `test_nofo_markdown.py`,
  `test_import_transforms.py`, `test_nofo.py`) to confirm none of it
  depends on `☐` being left unrendered.
- Full Django test suite was not run in this environment (the project
  pins Python `~3.14`, which wasn't available in the sandbox); the above
  targeted verification was run directly against the modified modules.

## Acceptance criteria

- [ ] A subsection whose stored markdown contains a raw `☐` renders the
      checkbox as a properly aligned icon in both HTML and PDF preview,
      matching the existing `◻` rendering.
- [ ] Numbered sublist items prefixed with `☐` are still correctly
      detected as sublist items.
- [ ] Saving a subsection through the manual editor with a pasted/typed
      `☐` character stores it normalized as `◻` going forward.
- [ ] Existing `.docx` import behavior and all existing checklist-table
      rendering (using `◻`) is unchanged.
- [ ] Full test suite passes in CI (not runnable locally in this
      environment due to the Python 3.14 requirement).

### DIFF
--- a/nofos/nofos/forms.py
+++ b/nofos/nofos/forms.py
@@ -9,6 +9,7 @@ from .models import (
     Section,
     Subsection,
 )
+from .nofo import replace_chars
 from .utils import get_icon_path_choices, user_is_nih_group
 
 
@@ -286,6 +287,9 @@ class SubsectionEditForm(forms.ModelForm):
             "name": forms.TextInput(),
         }
 
+    def clean_body(self):
+        return replace_chars(self.cleaned_data.get("body") or "")
+
     def clean(self):
         cleaned_data = super().clean()
         name = cleaned_data.get("name")
@@ -306,6 +310,9 @@ class SubsectionCreateForm(forms.ModelForm):
         widgets = {
             "name": forms.TextInput(),
         }
+
+    def clean_body(self):
+        return replace_chars(self.cleaned_data.get("body") or "")
 
     def clean(self):
         cleaned_data = super().clean()
--- a/nofos/nofos/forms.py
+++ b/nofos/nofos/forms.py
@@ -9,8 +9,12 @@ from .models import (
     Section,
     Subsection,
 )
-from .nofo import replace_chars
 from .utils import get_icon_path_choices, user_is_nih_group
+
+
+def normalize_subsection_checkbox_glyphs(body):
+    """Normalize U+2610 to the canonical U+25FB checkbox glyph."""
+    return body.replace("☐", "◻")
 
 
 def create_object_model_form(model_class):
@@ -288,7 +292,7 @@ class SubsectionEditForm(forms.ModelForm):
         }
 
     def clean_body(self):
-        return replace_chars(self.cleaned_data.get("body") or "")
+        return normalize_subsection_checkbox_glyphs(self.cleaned_data.get("body") or "")
 
     def clean(self):
         cleaned_data = super().clean()
@@ -312,7 +316,7 @@ class SubsectionCreateForm(forms.ModelForm):
         }
 
     def clean_body(self):
-        return replace_chars(self.cleaned_data.get("body") or "")
+        return normalize_subsection_checkbox_glyphs(self.cleaned_data.get("body") or "")
 
     def clean(self):
         cleaned_data = super().clean()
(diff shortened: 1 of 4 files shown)


mikec-ai:
I ran a full review and reproduced silent changes to unrelated Markdown caused by applying the import-wide cleanup helper in the manual editor. I pushed a focused follow-up commit that:

- narrows manual-editor normalization to `☐` → canonical `◻` only;
- preserves `&nbsp;`, NBSP, diaeresis, DEL, and other unrelated content;
- removes the forms dependency on the import cleanup module; and
- adds regression tests for both forms, rendering equivalence/layout classes, and numbered sublists.

Verification on the updated branch: 20 focused tests and the full 1,776-test suite passed; formatting and commit hooks passed. I’m doing a fresh independent review of the updated merge result next. Draft #818 will need to compose its separate `clean_body()` transform when it is revisited.

mikec-ai (approved):
Verified the updated implementation independently after the follow-up fix. Manual-editor normalization is now narrowly scoped, unrelated content is preserved, both checkbox glyphs render equivalently with the expected layout classes, and numbered-sublist behavior remains correct. The full application suite and all GitHub checks pass. Draft #818 will need to compose its separate form transform when revisited, but that is not a blocker here.

PR #857, opened 2026-09-03
feat: warn editors about long callout boxes

### LINKED ISSUE #852: Warn users when a callout box subsection contains too much content
## Summary

Add a non-blocking warning in the subsection editor when a subsection is marked as a callout box (`Is callout box: Yes`) and its content is long enough that it's likely being misused — i.e., used for narrative/detailed content instead of short, high-priority information.

## Problem

We've seen a real NOFO where a "Required format" subsection had `Is callout box: Yes` and contained a full page of detailed formatting requirements (file formats, fonts, spacing, margins, etc.). Because callout boxes have no max-height or overflow handling in the PDF/print CSS, the rendered callout stretched across a full page of the PDF.

This isn't a one-off. Reviewing NOFO curation across the tool, this pattern shows up consistently in **ACF** NOFOs specifically — ACF authors appear to reach for the callout-box checkbox for long, detailed content more often than other OpDivs. That makes this a recurring content-authoring gap worth addressing proactively, not just a fix for a single reported case.

This also isn't just a layout bug — it's a content-authoring problem. Our own guidance (below, from the FAQ) says callout boxes are for short, high-priority content, not long narrative or policy text. When a callout is overloaded with information, it stops doing its job: applicants can no longer tell at a glance what's actually important, which undermines the reason we use callouts in the first place.

> **When should I use a callout box (or callout component)?**
>
> Callout boxes are best for short, high-priority content that applicants need to notice quickly. Common uses include:
> - Contact and support information
> - Reminders and post-award requirements
> - Action-oriented information that stands on its own
>
> Avoid using callout boxes for long narrative content or detailed policy explanations — those belong in the body of the NOFO. Avoid using callout boxes for table footnotes — if a note only clarifies how to read or apply a specific table, keep it as a footnote directly below the table rather than pulling it into a separate visual component. The goal is to surface information that applicants might otherwise miss, not to highlight everything.
>
> **Tip:** If you're unsure, ask yourself: would an applicant miss this if it weren't visually separated? If the answer is no, it probably doesn't need a callout.

This issue proposes an in-app, automatic check as a lighter-weight complement to that guidance — something that flags long callouts at the point of authoring, without requiring the designer to remember to check.

## Proposed behavior

When editing a subsection with `Is callout box` checked, show a warning if the content exceeds a length threshold (exact number TBD — see Open questions). The warning should:

- Be **non-blocking** — designers can still save/publish; this is guidance, not validation.
- Appear inline in the subsection editor, near the callout-box checkbox/body field, reusing the existing warning-banner style already used for `html_class` mismatches in `subsection_edit.html`.
- Use applicant-centered language — the point isn't "you wrote too many words," it's "this may not read as important to the person applying."

### Suggested warning copy

> **This callout box may be too long**
> Callout boxes work best for a single, high-priority takeaway — like contact info, a reminder, or one clear action. The more content you add, the harder it is for applicants to tell what actually matters. Consider trimming this to one key point and moving the rest into the body text.
>
> *Would an applicant miss this if it weren't visually separated? If not, it probably belongs in the body instead.*

Alternate, shorter version (if space is tight in the UI):

> ⚠️ **This callout box has a lot of content.** Callouts are meant to highlight one important thing, not a full section. Long callouts are less likely to stand out to applicants — consider moving detailed content into the body.

## Technical notes (from investigation)

- `Subsection.callout_box` is a `BooleanField` on `models.py` (`nofos/nofos/models.py:721`); the editable content is `Subsection.body`, a Markdown field (`MartorField`) edited through the Martor editor in `subsection_edit.html`.
- Callout rendering template: `nofos/bloom_nofos/templates/includes/callout_box.html`; styling in `nofos/bloom_nofos/static/theme-base.css:944-1030` (plus print rules further down). There is currently **no max-height/overflow clipping** on callout boxes — they grow to fit content, which is why long callouts can push into/overflow a PDF page. Page breaks are opt-in only (`.callout-box.page-break-before`).
- There's already a comparable word-count heuristic elsewhere in the codebase we can follow as precedent: `get_combined_wordcount_for_subsections` applies smaller font-size classes to floating/sidebar callouts at 99/79 combined words (`nofos/nofos/templatetags/get_floating_callout_boxes_from_section.py`), and table cells get a "large" class at a 120-word threshold (`nofos/nofos/templatetags/utils/__init__.py:88-138`). These aren't identical to what we need, but they suggest ~100 words is already treated informally as "getting long" for a callout-adjacent component in this app — a reasonable starting point for discussion, not a final answer.
- UI precedent for the warning itself: `subsection_edit.html` already renders a left-border ⚠️ warning banner for `html_class` mismatches — we should reuse that same component/style rather than invent a new one.
- This issue proposes a simple in-app word-count heuristic, not an LLM/AI call — no such integration exists in the codebase today (confirmed via search), and building one is out of scope here.
- The existing Readability Metrics panel (`nofos/nofos/readability.py`, `HHS_NOFO_METRIC_GOALS` setting) is document-level, not per-subsection, but its pattern — presentation-only "Within target" / "Needs improvement" language rather than pass/fail, with configurable thresholds — is a good tone/pattern to mirror here at the subsection level.

## Open questions

- [ ] What's the right threshold — word count, character count, or something closer to "estimated rendered height"? Word count is simplest to implement and matches the existing precedent in this codebase.
- [ ] What should the actual number be? (Needs a content/design decision — the example that prompted this was a full PDF page, so the threshold should be well below that.)
- [ ] Should the threshold be configurable (env var / settings, like `HHS_NOFO_METRIC_GOALS`) or hardcoded to start?
- [ ] Should this warning also surface anywhere else — e.g., the section's subsection-list table (`section_detail.html`) — so designers can spot long callouts without opening each one?

## Out of scope

- Blocking save/publish based on callout length.
- Any true AI/LLM-based content review (this would be a word-count heuristic only).
- Changing callout box CSS to add overflow/max-height clipping (a separate, complementary fix worth considering, but not this issue).

mikec-ai:
## Summary

Closes #852.

Add a non-blocking warning beside the callout checkbox in the subsection editor when the body exceeds a configurable estimated word count.

- Default threshold: **more than 100 words**. `CALLOUT_WORD_WARNING_THRESHOLD` can tune it per environment. This is a starting content-guidance heuristic, not an approved length limit.
- Use the same whitespace-separated Markdown source count as existing floating callouts. Markdown syntax can contribute to the estimate; this does not estimate rendered height or evaluate content quality.
- Update immediately when authors edit the body or toggle the checkbox, including unnamed subsections. Render the initial state on the server and preserve submitted values after validation errors.
- Reuse the existing warning style with a polite live region. No save/publish blocking, model migration, output/CSS changes, AI calls, or new application dependencies.

## Verification

- Local full Django suite: 1,856 tests passed.
- Eight dependency-free JavaScript unit tests passed; included in CI.
- Changed Python files pass Black/isort; `git diff --check` passes.
- Local browser: reproduced the missing warning on main with a synthetic 130-word callout; verified the new warning, mouse/keyboard toggles, live shortening, the 100/101-word boundary, unnamed subsections, and saving above the threshold.
- Before/after screenshots and the detailed verification record are in the PR comment.

## Review focus

Please review the warning copy and the initial 100-word threshold. The warning stays editor-only; adding list-page guidance or changing callout layout remains out of scope.

### DIFF
--- a/.github/workflows/django_ci.yml
+++ b/.github/workflows/django_ci.yml
@@ -38,3 +38,7 @@ jobs:
 
       - name: Run tests
         run: poetry run python manage.py test
+
+      - name: Run JavaScript unit tests
+        # Uses Node's built-in test runner (available on ubuntu-latest); no npm dependencies.
+        run: node --test ../tests/js/*.test.cjs
--- a/nofos/bloom_nofos/.env.example
+++ b/nofos/bloom_nofos/.env.example
@@ -11,6 +11,10 @@ DOCRAPTOR_API_KEY="YOUR_API_KEY_HERE"
 GRABZIT_APPLICATION_KEY="YOUR_APPLICATION_KEY_HERE"
 GRABZIT_APPLICATION_SECRET="YOUR_APPLICATION_SECRET_HERE"
 
+# Non-blocking subsection-editor guidance above this estimated word count.
+# Counts whitespace-separated Markdown source words, not rendered PDF height.
+CALLOUT_WORD_WARNING_THRESHOLD=100
+
 # Uses the hhs-nofo-metrics release pinned in pyproject.toml and poetry.lock.
 # This is only the starting value for the HHS_NOFO_METRICS_ENABLED constance
 # setting: once that has been saved in the admin, the saved value wins.
--- a/nofos/bloom_nofos/settings.py
+++ b/nofos/bloom_nofos/settings.py
@@ -571,6 +571,10 @@ DOCRAPTOR_API_KEY = env.get_value("DOCRAPTOR_API_KEY", default="")
 GRABZIT_APPLICATION_KEY = env.get_value("GRABZIT_APPLICATION_KEY", default="")
 GRABZIT_APPLICATION_SECRET = env.get_value("GRABZIT_APPLICATION_SECRET", default="")
 
+# Advisory only: estimated words in the subsection's Markdown source, using the
+# same whitespace count as floating callouts. This is not a publishing limit.
+CALLOUT_WORD_WARNING_THRESHOLD = env.int("CALLOUT_WORD_WARNING_THRESHOLD", default=100)
+
 # Provisional, source-native readability metrics integration. The package is
 # pinned, but each environment opts into the feature explicitly. This env var is
 # only the starting value for the HHS_NOFO_METRICS_ENABLED constance setting,
--- a/nofos/bloom_nofos/static/js/nofos/subsection_edit.js
+++ b/nofos/bloom_nofos/static/js/nofos/subsection_edit.js
@@ -13,3 +13,46 @@ document.addEventListener("DOMContentLoaded", () => {
     }
   });
 });
+
+// Keep this independent of the heading-copy button: unnamed subsections have
+// no copy button, but can still be callouts.
+document.addEventListener("DOMContentLoaded", () => {
+  const checkbox = document.getElementById("callout_box");
+  const warning = document.getElementById("callout-word-warning");
+  // Martor adds a generated suffix to the textarea ID.
+  const body = document.querySelector('.main-martor--container textarea[name="body"]');
+  if (!checkbox || !warning || !body) return;
+
+  const threshold = Number(warning.dataset.wordThreshold);
+  if (!Number.isFinite(threshold)) return;
+
+  let editor;
+  const update = () => {
+    // Advisory estimate, matching Python's body.split() and existing floating
+    // callouts. Markdown syntax may contribute to the count; this is not a
+    // rendered-height or precise readability measurement.
+    const text = editor ? editor.getValue() : body.value;
+    const wordCount = (text.match(/\S+/g) || []).length;
+    const hidden = !checkbox.checked || wordCount <= threshold;
+    if (warning.hidden !== hidden) warning.hidden = hidden;
+  };
+
+  checkbox.addEventListener("change", update);
+  body.addEventListener("input", update);
+
+  // Martor initializes Ace on jQuery ready, which can follow DOMContentLoaded.
+  // Until it is ready, keep the server-rendered warning and textarea fallback.
+  const attachEditor = (attempts = 0) => {
+    const element = document.querySelector(".main-martor--container .ace_editor");
+    if (window.ace && element) {
+      editor = window.ace.edit(element);
+      editor.on("change", update);
+      update();
+    } else if (attempts < 50) {
+      setTimeout(() => attachEditor(attempts + 1), 100);
+    }
+  };
+
+  update();
+  attachEditor();
+});
--- a/nofos/bloom_nofos/static/js/nofos/subsection_edit.js
+++ b/nofos/bloom_nofos/static/js/nofos/subsection_edit.js
@@ -26,8 +26,17 @@ document.addEventListener("DOMContentLoaded", () => {
   const threshold = Number(warning.dataset.wordThreshold);
   if (!Number.isFinite(threshold)) return;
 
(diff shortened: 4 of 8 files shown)


mikec-ai:
## Local verification and screenshots

Verified the change locally at commit `d7317722` using synthetic content, not an agency NOFO. The baseline is current main (`dc0bbf20`).

- **Before:** a 130-word "Required format" callout has no length guidance.
- **After:** the same callout shows the non-blocking warning below the checkbox.
- Mouse and keyboard toggles show/hide the warning without saving.
- Shortening the body hides it; exactly 100 words stays quiet and 101 words shows it.
- Unnamed subsections also support the warning, toggles, and live body edits.
- Saving the original 130-word callout succeeds and retains the content and callout setting.

**Automated checks:** all 1,856 Django tests and eight JavaScript unit tests passed locally, plus Black/isort and whitespace checks. The JavaScript tests are wired into CI. The real-browser pass caught and fixed Martor's generated textarea ID; the listener now uses the stable field name.

The count intentionally follows Builder's existing whitespace-based Markdown estimate. The 100-word default is adjustable guidance, not a publishing limit or rendered-height measurement. No dev/production deployment was made.

### Before

Before: long callout selected with no length guidance

### After

After: long callout shows non-blocking guidance below the checkbox

Both images were uploaded directly using GitHub CLI 2.99.0's `gh pr comment --attach` support. No screenshot files were added to the source branch.

mikec-ai:
@benbrasso ready for your review. The local browser verification and before/after screenshots are in the comment above; both images were uploaded directly through the new CLI attachment support. All 1,856 Django tests and eight JavaScript tests pass in GitHub, and all container/security checks are green. The first image-build attempt hit a Python package download timeout; rerunning the failed job passed without code changes.

Please take a look at the warning copy and the configurable 100-word starting threshold. This is editor-only, non-blocking guidance; saving, publishing, and PDF layout are unchanged. No deployment has been made.

benbrasso:
## Follow-up: Key facts/dates exemption + clearer warning copy

Two small changes on top of the original commit, based on review feedback:

**1. "Key facts" / "Key Facts" / "Key dates" / "Key Dates" subsections are now exempt from the warning, regardless of length.**

These subsections are structured fact lists (opportunity name, deadlines, assistance listing, etc.), not narrative content — they're expected to exceed the word-count guidance by design, so the warning no longer applies to them. The name match is exact (after trimming whitespace) and reuses the same four strings the app already recognizes for floating/sidebar callouts (`is_floating_callout_box` in `templatetags/utils/__init__.py`), so a name like "Key Fact" or "key facts" (different capitalization) still gets the warning as normal — only an exact match is exempt.

This is live in both places the warning is computed:
- Server-side, on page load (`views.py`)
- Client-side, live as you type — renaming a subsection into or out of one of these four names toggles the warning immediately, without saving (`subsection_edit.js`)

**2. Warning copy is more specific about where content should go.**

> Before: "...consider moving detailed content into the body."
>
> After: "...consider moving detailed content into the body of the section, before or after the callout."

The original phrasing left "the body" ambiguous — the updated copy says specifically where that content should go.

**Testing:**
- 5 new Django tests covering the exemption (all four exact names, a whitespace-padded exact match, and four near-misses that should still warn) — all passing alongside the existing 18 tests in this file.
- 3 new JavaScript tests covering the same cases client-side, including live renaming across the exemption boundary — all passing alongside the existing 8.
- Full test suite, Black, and isort all pass locally.

No changes to saving, publishing, or PDF layout — this remains editor-only, non-blocking guidance.

benbrasso (approved):
Approved, but added a commit

benbrasso:
@mikec-ai your merge

PR #860, opened 2026-09-03
fix: explain status-dependent NOFO actions

### LINKED ISSUE #806: Improve discoverability of status-dependent actions in NOFO Builder
**Summary**
Several actions in NOFO Builder — including deleting sections and subsections — are only available when a NOFO's status is set to **Draft**. There is currently no help text, inline messaging, or visual indicator on the main NOFO page that communicates this dependency to users. As a result, users are unable to locate or understand why certain actions are unavailable without external support.

**Problem**
When a NOFO's status is set to anything other than Draft (e.g., Active, In review), the delete action for sections and subsections is restricted. This is likely intentional to prevent accidental deletion of NOFOs that are in an active workflow stage. However, there is no feedback in the UI to explain why the action is unavailable or what the user needs to do to enable it.

Additionally, the delete action itself requires navigating away from the main NOFO page — either into the subsection editor or through the Configure section page — which adds friction even when the correct status is set.

**Proposed improvements to consider**
- Add inline help text or a tooltip on the main NOFO page that communicates which actions are restricted based on the current status, and what status is required to perform them
- When a restricted action is attempted, display a clear error or informational message explaining the status dependency (e.g., "To delete a section, change the NOFO status to Draft")
- Consider surfacing status-dependent action availability more prominently in the status field UI itself
- Consider adding an inline delete option on the main NOFO page to reduce the number of steps required to delete a section or subsection when the correct status is set

**Acceptance criteria**
- [ ] Users are informed — either proactively or at the point of attempting an action — that certain actions require the NOFO status to be set to Draft
- [ ] The messaging is clear enough that users can resolve the issue without external support
- [ ] Status-dependent restrictions are consistently communicated across all affected actions, not just deletion
- [ ] Any inline messaging or tooltips added do not interfere with the existing workflow for users who are already familiar with the status system
- [ ] The solution is discoverable without prior knowledge of the tool or external documentation

**User impact**
Medium impact. Users who are unfamiliar with the relationship between NOFO status and available actions are likely to encounter this as a blocker and require support to resolve it. This increases support burden and creates friction during the design workflow.

mikec-ai:
## Summary

Closes #806.

- Explain the Draft requirement for deleting a NOFO or its subsections beside the main status control, and list the statuses that allow re-importing when that action is restricted.
- Replace misleading subsection Delete links with explanatory unavailable text outside Draft, on both Configure section and the subsection editor. Link the guidance back to the main status control, with spacing for the sticky header.
- Keep Add subsection and full-width table controls consistent with the existing edit restrictions, including Published with modifications. Explain Published, Cancelled, and Archived restrictions without offering archived NOFOs an unavailable status control.
- Correct the Published-with-modifications message: editing is allowed; re-importing and deleting are not.

This changes presentation only. Existing permissions, status transitions, delete/re-import guards, confirmation screens, and return destinations remain unchanged. No new inline deletion, section-deletion endpoint, or automatic status changes. #818 and FY27 work are excluded.

## Verification

- Reproduced on main with a synthetic local Active NOFO: Configure section showed Delete; clicking it returned “Only subsections of draft NOFOs can be deleted.”
- Added nine regression tests covering all eight statuses, Published with modifications, archived variants, both subsection surfaces, status guidance, Draft confirmation/redirects, non-Draft GET/POST rejection, and group access.
- Before the fix, those tests exposed 22 failing subcases; after the fix, all nine pass.
- Full local Django suite: **1,858 tests passed**. Final focused rerun also passes.
- Browser checked the Active guidance, subsection editor, link back to the visible status control, and switching to Draft to reach the existing confirmation screen. No real NOFOs were used or deleted. Screenshots follow in a comment.
- Formatting hooks and final diff review completed.

No deployment or merge is included; human review follows CI.


mikec-ai:
Local browser verification for #806, using synthetic content and an isolated local database. **Not a dev or production test.**

**Before:** an Active NOFO showed a Delete link, which led to the existing 400 rejection.

<details>
<summary>Baseline screenshots</summary>

Before: Configure section offers Delete on an Active NOFO

Before: clicking Delete returns the Draft-only rejection

</details>

**After:** Configure section explains the Draft requirement, replaces Delete with unavailable text, and keeps Add subsection available. The guidance links directly to the main status control.

After: status guidance and unavailable Delete controls on Configure section

The status control is visible below the sticky header when following that link:

After: the Active status control explains the Draft requirement

The subsection editor gives the same guidance:

After: Other actions keeps Add subsection available and explains why Delete is unavailable

Also verified switching the synthetic NOFO to Draft still reaches the existing delete confirmation, then returned it to Active without deleting it. Automated GET/POST tests verify non-Draft deletion remains rejected and Draft deletion/redirects remain unchanged.

Full local suite: **1,858 tests passed**, including nine new regression tests. CI is running; review request will follow once checks are green.

mikec-ai:
@benbrasso — ready for your review. All CI checks are green, and the full local suite passed (1,858 tests).

This is the small, presentation-only fix for #806: explain the Draft requirement before users hit the existing restriction, keep subsection controls consistent with their existing status rules, and correct the Published-with-modifications re-import message. No permissions or workflow transitions changed.

Before/after screenshots and reproduction details are in [the local verification comment](https://github.com/HHS/simpler-grants-pdf-builder/pull/860#issuecomment-5529054899). I also checked Published, Published with modifications (including persisting the table setting), and Archived views in the local browser. The status link lands with the dropdown visible below the sticky header.

Left unmerged and undeployed for your review.

benbrasso (changes requested):
Thanks for tackling #806, Mike — this fixes a couple of real bugs, not just copy: the Delete links that were previously clickable-but-400'd on non-Draft NOFOs, and the incorrect claim that "published with modifications" NOFOs can be re-imported (they can't — `NofosImportOverwriteView.handle_nofo_create` blocks all published NOFOs regardless of modifications). A few things I'd like addressed before merging:

**1. Consistency: the subsection editor duplicates the guidance, and buries it**

`subsection_edit.html` currently renders both the `subsection_action_guidance.html` alert *and* a separate "Delete subsection unavailable — requires Draft status" line, both inside the "Other actions" accordion. Two problems: they say the same thing twice, and the accordion is a collapsed `<details>` element, so the alert is invisible until the user clicks to expand it — which works against the "proactive notification" goal from #806.

Can you move the `{% include "nofos/includes/subsection_action_guidance.html" %}` to the top of the page (visible on load, same placement as the section-detail alert), and drop the inactive "Delete subsection unavailable" text from inside the accordion since the alert already covers it?

**2. Consistency: `section_detail.html`'s "Delete unavailable" text is also redundant, and introduces a new pattern**

That page already shows the alert at the top; the per-row "Delete unavailable" label repeats it. "Add subsection" is already fully hidden (no "Add unavailable" placeholder) when it's not permitted — can Delete follow the same pattern (just omit the link) instead of introducing inactive-link-text, which doesn't exist anywhere else in the app?

**3. Copy: tighten up the status-guidance sentences in `subsection_action_guidance.html`**

Right now all three non-Draft branches end with "Review NOFO status and only change it when appropriate for your workflow" — it links to the right place, but doesn't say what to change the status *to*, and "when appropriate for your workflow" is vague filler. It also implicitly tells the user how to run their team's process, which even our own FAQ is careful not to do. Suggested replacements (make the link itself the action, state the actual constraint, keep it factual rather than prescriptive):

- **Cancelled:** "Subsections can't be edited, added, or deleted while this NOFO is Cancelled. Draft is the only status that allows deletion — [change the status to Draft](#nofo-status) if you need to."
- **Published, no modifications:** "This NOFO is Published. To edit or add subsections, add modifications from the main NOFO page. Draft is the only status that allows deletion — [change the status to Draft](#nofo-status) if you need to."
- **Editable, non-Draft (Active / Ready for QA / In review / Dep Sec / Paused / Published-with-modifications):** "You can still edit or add subsections. Draft is the only status that allows deletion — [change the status to Draft](#nofo-status) if you need to."

("Draft is the only status that allows deletion" also matches the wording in our internal FAQ, so the in-app copy and docs stay in sync instead of drifting apart.)

**Everything else checks out** — traced `can_edit_subsections`/`can_delete_subsections` against the real enforcement code (`PreventIfArchivedOrCancelledMixin`, `PreventIfPublishedMixin`, and the delete/reimport view checks) for all 8 statuses × archived/modifications combinations, and the logic is accurate for every case. No missing scenarios.

Once the above are addressed, this looks good to merge.

---

mikec-ai:
Addressed all three review items in `45fd3c37` and resolved the branch against current `main` while preserving the #857 long-callout behavior.

Changes:
- moved the status guidance above the subsection form so it is visible on load
- removed the redundant inactive Delete labels from both views
- updated all three non-Draft branches to name Draft directly and link the action text
- added regression coverage for placement, copy, and absence of the inactive labels

Verification:
- 33 focused Django tests passed
- 11 subsection-editor JavaScript tests passed
- full local suite: 1,898 tests passed, with no system-check issues
- browser-verified the Active section view and subsection editor; Add subsection remains available and Delete is omitted

Updated local browser evidence:

Active NOFO section view with concise Draft guidance and no redundant Delete unavailable labels

Status guidance visible at the top of the subsection editor on page load

Expanded Other actions area retains Add subsection without a redundant unavailable Delete label

mikec-ai:
@benbrasso — all requested changes are addressed on `45fd3c37`, the branch is current with `main`, and every GitHub check is green. Updated browser evidence and full verification details are in the preceding comment. Re-requesting your review.

benbrasso (approved):
Reviewed the latest commit (45fd3c37). All three requested changes are addressed: the subsection guidance is visible on load, the redundant unavailable Delete labels are removed, and the copy now clearly identifies and links Draft as the required status. The added regression coverage looks appropriate, and all CI checks are green. No remaining functional findings from me—this looks good to merge.

benbrasso:
One non-blocking process note: `45fd3c37` does not use a Conventional Commit prefix. Please squash-merge using the current PR title, or rewrite that commit message if using rebase merge.

Issue #844, opened 2026-08-28
Epic: Publish NOFOs from NOFO Builder to Grants.gov via the Simpler Grants.gov (SGG) API


benbrasso:
> **This is an epic** — a tracking issue for the spike, not a single unit of work. It stays open for the full effort; the checklist under **Sub-issues** below is the actual work breakdown. Each item is written to stand alone as its own issue title — convert each checkbox to a linked sub-issue (GitHub's "Convert to sub-issue" on a task-list item) rather than working the whole list inside this issue.

## Problem

Today, publishing a NOFO from NOFO Builder to Grants.gov is a fully manual process. NOFO Builder can generate a NOFO's PDF/HTML on demand and stores opportunity-level metadata (agency, subagency, opportunity number, etc. — see the field mapping below for specifics), but it has no connection to any downstream publishing system. The actual publishing workflow runs through Grant Solutions, eRA, or the Announcement Module, none of which currently talk to NOFO Builder.

Separately, Simpler Grants.gov (SGG) — the platform that will eventually replace Grants.gov — already exposes a set of publishing-related API endpoints (opportunity search, applications, user management, early grantor functionality). These endpoints are functional but currently staging-only and unused in production, because Smarter Grants Management (SGM), the system expected to be their primary consumer, hasn't been built yet. Until then, all real publishing flows through the legacy path: systems post to Grants.gov, which batch-syncs into SGG on a regular cadence.

This leaves a gap and an opportunity: there's a working publishing API sitting idle, and a NOFO Builder tool that agencies already use daily but that can't publish anywhere.

For context, the public-facing opportunity listings at https://simpler.grants.gov/search are themselves populated via that same batch job, syncing in from https://grants.gov/ — so today there is no code path, anywhere, that writes an opportunity into SGG directly. This spike would be the first.

## Goal

Connect NOFO Builder to the SGG API so that, when a NOFO is ready, a user can trigger publishing directly from NOFO Builder — with the API handling the actual post to Grants.gov (via SGG) and reporting back success or errors.

This is explicitly a proof of concept / spike, with two goals of roughly equal importance:

1. **Prove the workflow end-to-end.** Demonstrate that NOFO Builder can be wired directly into the publishing pipeline without a full architectural rebuild — adding a capability that doesn't exist today for either NOFO Builder or SGM, and showing that we can connect systems that already exist (NOFO Builder is already running in AWS and already used by agencies) rather than waiting on longer-running rebuild efforts.
[...]

PR #861, opened 2026-09-03
feat: add structured assistance listing (CFDA) number field to Nofo, flag-gated

### LINKED ISSUE #859: Add a structured assistance_listing_number field to the Nofo model
Part of #844.

Step 3 of the epic — add a structured field for a NOFO's CFDA/Assistance
Listing Number, which today is only ever captured as incidental free
text inside a NOFO's "Key Facts" content (if the source Word document
happens to include it at all), with no dedicated field and no import-time
recognition.

## Why now

HHS's FY27 NOFO guidance and Word template are expected to release in
mid-to-early September 2026 — imminent. Whatever we land on here needs
to be reflected in that template *before* it goes out broadly to HHS
grants management teams, not after — see "Action needed" below.

## What this adds

- `assistance_listing_number` field on `Nofo` (`CharField`, max 6
  characters — confirmed directly from Simpler Grants.gov's own write API
  schema, `opportunity_grantor_schemas.py`, which validates
  `Length(max=6)` on this exact field).
- **Import-time auto-suggestion**: scans an imported Word document for a
  line identifying the assistance listing number (matching the label
  variants seen in real documents today, e.g. `"Assistance listing:"`),
  extracts just the CFDA-shaped number, and pre-fills the field —
  mirroring how `Opportunity Number`, `OpDiv`, `Agency`, etc. are already
  auto-suggested today.
- **Edit UX**: its own field on the "Basic information" edit screen, with
  the same "find this value elsewhere in the document and offer to sync
  it" checkbox UI that editing `Number` or `Application deadline` already
  has today. No format enforcement beyond the 6-character cap, matching
  how `Number` itself is validated (SGG's own example value, `"12.ABC"`,
  isn't purely numeric either, so a stricter format check would be wrong).
- **Cover page display**, mirroring how `Number` already appears on the
  public PDF cover page — built, but hidden behind a feature flag (see
  below) pending a decision on whether it should actually show there.

## Feature flags (Constance, both default off)

- `HHS_NOFO_ASSISTANCE_LISTING_ENABLED` — master flag. While off: no
  import-time scanning happens, the field is hidden from the edit page,
  and the edit view itself 404s even if hit directly (mirrors the
  existing `HHS_NOFO_METRICS_ENABLED` defense-in-depth pattern). This
  lets the whole feature merge and deploy to `main` inertly, with nothing
  turned on until it's actually ready.
- `HHS_NOFO_ASSISTANCE_LISTING_ON_COVER_ENABLED` — narrower, second flag;
  only relevant once the field itself is enabled. Separate decision,
  separate switch.

## Action needed (outside this PR, and time-sensitive)

Whoever owns the FY27 Word template needs to confirm, before it ships:
1. The exact label wording for the assistance listing number, if it gets
   its own dedicated line (the more reliable option — matches the
   existing pattern for `Agency`/`OpDiv`/`Tagline`) vs. staying folded
   into Key Facts free text with inconsistent wording across templates.
2. Whether it should also appear on the Basic Information Key Facts box,
   only in a dedicated spot, or both (precedent exists for each — see the
   field itself vs. `Tagline`/`Number`).
3. Whether it should show on the NOFO's public cover page.

None of these block writing the code (the flags mean we can ship inertly
and adjust later), but they do block deciding when to flip the flags on,
and the template's own design should reflect whatever we land on.

## Out of scope

Anything related to actually publishing to SGG (matching this number
against SGG's pre-existing `AssistanceListing` record) — that's Step 6,
tracked separately under #844.

benbrasso:
## Summary
- Adds `assistance_listing_number` to `Nofo` (max 6 chars — confirmed
  directly from Simpler Grants.gov's own write API schema, which
  validates `Length(max=6)` on this exact field)
- Import-time auto-suggestion, mirroring the existing pattern for
  `Number`/`OpDiv`/etc.
- A dedicated "Basic information" edit screen with the same
  find-and-sync-elsewhere-in-the-document checkbox UX that editing
  `Number` or `Application deadline` already has
- Cover-page display, mirroring how `Number` already appears there
- All of the above gated behind two new Constance flags, both default
  off

Part of #844 (Step 3). Closes #859.

## Timeline / urgency

HHS's FY27 NOFO guidance and Word template are expected to release in
mid-to-early September 2026 — imminent. Whatever we land on for this
field needs to be reflected in that template before it goes out broadly
to HHS grants management teams. See "Open questions" below — this is a
real external dependency, not something this PR can resolve on its own.

## Why flag-gated

Today, the assistance listing number is only ever captured as incidental
free text inside a NOFO's "Key Facts" content, in at least three
different label spellings depending on the source document, with zero
extraction logic. This PR builds the full path (structured field, import
detection, edit UI, cover-page display) but ships it inert:

- **`HHS_NOFO_ASSISTANCE_LISTING_ENABLED`** (default off) — gates the
  field, import-time suggestion, and edit screen together. While off:
  no import-time scanning happens on any Word import, the field is
  hidden from the "Basic information" edit page, and the edit view
  itself 404s even if visited directly — same defense-in-depth pattern
  as the existing `HHS_NOFO_METRICS_ENABLED` flag.
- **`HHS_NOFO_ASSISTANCE_LISTING_ON_COVER_ENABLED`** (default off) — a
  second, narrower flag for cover-page display specifically, since
  that's a separate decision independent of the field itself being on.

This means merging and deploying today changes nothing for any user
until someone explicitly flips a flag in the Django admin (Constance,
no redeploy needed).

## Changes
- `nofos/nofos/models.py`: `assistance_listing_number` field
- `nofos/nofos/migrations/0133_nofo_assistance_listing_number.py`
- `nofos/nofos/nofo.py`: `suggest_nofo_assistance_listing_number`,
  wired into `suggest_all_nofo_fields`; matches known label variants,
  extracts just the CFDA-shaped code (discarding any trailing program
  title text), flag-gated
- `nofos/nofos/forms.py`: `NofoAssistanceListingNumberForm`
- `nofos/nofos/views.py`: `NofoEditAssistanceListingNumberView`
  (mirrors `NofoEditNumberView`'s sync-offer pattern, 404s while
  disabled); context flags added to `NofosEditView`/`NofosDetailView`
- `nofos/nofos/urls.py`: `nofo_edit_assistance_listing_number`
- `nofos/nofos/templates/nofos/nofo_edit_assistance_listing_number.html`
  (new); `nofo_edit.html` (new flag-gated row); `nofo_view.html`
  (flag-gated cover-page display)
- `nofos/bloom_nofos/settings.py`, `.env.example`: the two new
  Constance flags
- Tests: model field, import-suggestion (label variants, malformed
  input, flag on/off), edit view (GET/POST, sync-offer, flag on/off
  404), Basic Information row visibility, cover-page display — all new

## Open questions (for whoever owns the FY27 template, not blocking this PR)
1. Exact label wording, if it gets its own dedicated line (more
   reliable than scanning Key Facts free text — matches how
   `Agency`/`OpDiv`/`Tagline` already work).
2. Whether it should also still appear in Key Facts, only in a
   dedicated spot, or both.
3. Whether it should show on the NOFO's public cover page (this PR
   builds that capability but leaves it off by default).

## Test plan
- [x] `python manage.py test nofos users` — full suite (1470 tests),
      all passing
- [x] `black --check` / `isort --check-only` on changed files — clean
- [x] `makemigrations --check --dry-run` — migration matches model
      state exactly
- [ ] No UI to click through with flags off (by design) — with
      `HHS_NOFO_ASSISTANCE_LISTING_ENABLED=True` locally, confirm the
      Basic Information row appears, import auto-suggestion populates
      the field from a real NOFO doc, and the sync-offer checkbox works

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_012VnrZjGzrrJjhLMNPKwiXm


mikec-ai (changes requested):
I found three blocking edge cases in the import/flag behavior. The overall structure and mirrored edit flow look good, and the existing 12 feature tests plus CI are green, but these need coverage before enabling the field:

1. **A disabled feature can erase an existing value on re-import.** `suggest_nofo_assistance_listing_number()` returns `""` while the master flag is off, and `suggest_all_nofo_fields()` unconditionally assigns that result. The re-import path calls `suggest_all_nofo_fields()`, so turning the flag off does not remain inert: it silently clears a previously saved assistance listing number. The same happens with the flag on when the re-imported document lacks a recognized label. Please only replace the stored value when a valid suggestion is found (a new import already starts blank), and add regression tests for disabled and missing-label re-import/preservation.

2. **Malformed overlong codes are silently truncated into valid-looking values.** The regex is not bounded at the end, so `93.8840` becomes `93.884` and `12.ABCD` becomes `12.ABC`. Add a boundary such as `(?![A-Za-z0-9])` after the capture and tests for these inputs; delimiters used before a program title (space/colon/etc.) can still work.

3. **The narrower cover flag bypasses the master flag.** The context checks only `HHS_NOFO_ASSISTANCE_LISTING_ON_COVER_ENABLED`. Because Constance flags can be toggled independently, master off + cover on still exposes the stored field on the public cover, despite the master flag promising the field is hidden and the cover setting saying it is only relevant when the master flag is on. Compute this as master AND cover and test all combinations.

I stopped before the manual flagged-UI pass because these affect the core contract. Once updated, I’m happy to re-review and run the browser verification.

benbrasso:
Thanks for catching all three — fixed and pushed in 94849f9:

1. `suggest_all_nofo_fields()` now only overwrites
   `assistance_listing_number` when a real suggestion is found, instead
   of unconditionally assigning `""`. Added regression tests for both
   cases you named: flag disabled with a label present, and flag
   enabled but the re-imported document has no recognized label —
   both now leave an existing value untouched. A new NOFO with no
   label still correctly stays blank.
2. Bounded the regex with `(?![A-Za-z0-9])` right after the capture, so
   `93.8840` and `12.ABCD` are now rejected outright rather than
   truncated to `93.884`/`12.ABC`. Delimited trailing text (space,
   colon, etc. before a program title) still matches fine — added
   tests for both the overlong-rejection and still-works cases.
3. The cover-page context now computes `HHS_NOFO_ASSISTANCE_LISTING_ENABLED
   and HHS_NOFO_ASSISTANCE_LISTING_ON_COVER_ENABLED`, not the cover flag
   alone. Added tests for all four flag combinations — confirmed
   master-off-cover-on now correctly hides the value.

Full suite: 1479 tests passing (1470 + 9 new). Ready for the browser
verification pass whenever you get to it.

---

mikec-ai:
Browser verification passed on the current head (`c5944b02`) using a synthetic local NOFO.

Verified:
- the assistance listing value appears in Basic information and opens the dedicated editor
- saving `12.ABC` updates the structured field and the selected matching subsection
- the cover displays the saved value when both flags are enabled
- the master flag hides the Basic information row, blocks the direct editor URL, and suppresses the cover value
- the cover-only flag suppresses the cover value without disabling the editor
- a 7-character value is rejected and the saved value remains unchanged

I also re-checked the three requested fixes; they address the preservation, truncation, and flag-composition issues.

Local browser verification of the assistance listing editor and matching subsection update

mikec-ai (approved):
Re-reviewed the fixes and completed the browser verification pass on this head. The three requested changes are resolved; focused tests, the full 1,888-test local suite, and GitHub CI are green. Browser evidence and the verified flag combinations are documented in the preceding comment. No further changes requested.

Issue #865, opened 2026-09-05
Add usage & import-quality metrics dashboard to NOFO Builder


benbrasso:
## Summary
Add an internal, permission-gated metrics page to NOFO Builder that tracks usage and import-quality trends month over month, to give the team (and FY27 stakeholders) a quantifiable read on whether the tool is getting easier and faster to use as the user base grows.

## Background
Follow-up to #839, which investigated low-cost options for usage visibility ahead of a potential FY28 migration. That investigation's recommendation was to instrument the app itself rather than adopt an external tool — see the closing comment on #839 for the full reasoning.

## Metrics to track
| Metric | Definition |
|---|---|
| Total users | Cumulative NOFO Builder accounts to date, by sign-up date. |
| Monthly active users | Distinct users with ≥1 recorded action that month (import, edit, or print). |
| NOFOs created | New NOFO records first imported that month. |
| Import → first live PDF | Median calendar time from first import to first non-watermarked download (adaptive units: hours/days). |
| Blocking import errors | % of import attempts that fail outright, by error code. |
| Warnings per import | Average non-blocking formatting warnings per successful import. |

## Scope
- [ ] New `ImportAttempt` model + migration (user, filename, error_code, warning_count, nofo FK nullable, timestamp)
- [ ] Update `nofo.py` to always compute mammoth's warning count (not just under `WORD_IMPORT_STRICT_MODE`) — purely additive, no change to existing strict-mode blocking behavior
- [ ] New `nofos.view_builder_metrics` permission + "Metrics viewers" group
- [ ] Query layer for the six metrics above (validate via a management command first, consistent with existing `count_prints`-style commands)
- [ ] New metrics page, extending the app's existing `base.html` (inherits real header/nav/footer)
- [ ] Print/Save-as-PDF export via a print stylesheet
- [ ] Inline plain-language definition shown under each chart

## Out of scope for this issue
- Concurrent-usage tracking (see #839) — a cheap fallback (counting active Django sessions) can be added later if needed, not part of this build
- Any external tool (GA/DAP, Cabin, Metabase) — deferred per #839
- "Active working time" as a refinement of the time-to-PDF metric (isolating hands-on editing time from review-wait time) — worth revisiting only if the simpler elapsed-time metric proves too noisy

## Data workflow
[...]

PR #866, opened 2026-09-05
feat: add ImportAttempt tracking for import quality metrics

benbrasso:
## Summary
First step of #865 (usage & import-quality metrics dashboard). Adds an `ImportAttempt` model that records every NOFO import/reimport attempt — success or failure — as the data source for the metrics page's error-rate and warnings-per-import charts.

## What's in this PR
- **New `ImportAttempt` model + migration** (`nofos/migrations/0134_importattempt.py`): user, filename, `is_reimport`, `error_code` (blank on success), `warning_count`, and the resulting `nofo` (nullable, `SET_NULL` on delete so history survives a later deletion).
- **`nofo.py`**: mammoth's warning count is now always computed, not just when `WORD_IMPORT_STRICT_MODE` is on. This is additive only — the actual blocking condition (`if config.WORD_IMPORT_STRICT_MODE and warnings: raise ...`) is unchanged from today's behavior.
- **`views.py`**: wires `ImportAttempt` logging into every success/failure branch of `NofosImportNewView` and `NofosImportOverwriteView`, including the two-request reimport-confirmation flow (opportunity-number mismatch), threading `warning_count` through the session in that case.
- Deliberately **not** wired into Composer or Compare's import views — they create different document types (ContentGuide, CompareDocument), so a `track_import_metrics` flag (off by default on the shared base view, on only for the two NOFO views) keeps their data out of these metrics.
- Updated existing unit tests/mocks in `test_import.py`, `composer/tests/test_views.py`, and `compare/test_views.py` to match `parse_uploaded_file_as_html_string`'s new `(html, warning_count)` return shape.

## Out of scope for this PR
Everything else in #865's checklist (permission/group, query layer, the metrics page itself, PDF export) — tracked as follow-up PRs.

## Testing
Verified by code review and `black`/Python syntax checks. I wasn't able to run the full Django test suite (`manage.py test`) against a real Postgres instance in the environment this was written in, so please rely on CI here rather than take that as already green.


mikec-ai (approved):
Reviewed the import-tracking changes, including parser return-value callers, success/failure logging, reimport confirmation, and separation from Composer/Compare. No blocking findings.

Ran the import, reimport, metrics, Compare, and Composer suites against the combined stack at bf83488c: 246 tests passed using a fresh local SQLite test database. This exercises the same step-1 implementation included here. CI also passes. Reviewed the migration's nullable SET_NULL relationships and preservation of existing strict-mode blocking behavior.

PR #867, opened 2026-09-05
feat: add "Metrics viewers" permission and gated metrics page skeleton

benbrasso:
## Summary
Second step of #865 (usage & import-quality metrics dashboard). Adds a dedicated permission and group for viewing the future metrics page, plus a minimal gated skeleton view, without any real charts yet.

## Depends on #866
This branch is stacked on top of #866 (`ImportAttempt` tracking) — its migration needs step 1's migration to already exist for a clean, linear migration chain. Until #866 merges, this PR's diff will show both PRs' changes together; that's expected and will resolve itself once #866 lands. **Please merge #866 first.**

## What's in this PR
- **New `nofos.view_builder_metrics` permission**, declared on `Nofo.Meta.permissions`.
- **A data migration** that creates a "Metrics viewers" `Group` and attaches the new permission to it (`get_or_create`-based, safe to re-run).
- Superusers get access automatically through Django's normal permission behavior — no separate wiring needed for them.
- **`MetricsViewerRequiredMixin`** in `mixins.py`, mirroring the existing `SuperuserRequiredMixin` pattern, checking the new permission instead of `is_superuser`.
- **A minimal `BuilderMetricsView`** at `/nofos/metrics`, gated by the new mixin, extending the app's existing `base.html` (so it already inherits the real header/nav/footer). Content is a placeholder only.

## Out of scope for this PR
The actual usage/quality metrics — query layer, charts, PDF export — land in follow-up PRs once this permission scaffolding and #866's `ImportAttempt` data are both in place.

## Testing
Verified by code review and `black`/Python syntax checks in the environment this was written in — I wasn't able to run the full Django test suite (`manage.py test`) against a real Postgres instance there, so please rely on CI here rather than take that as already green.


mikec-ai (approved):
Reviewed this step's permission migration, gate, URL/view, and audit-test adjustments separately from the inherited step-1 diff. No blocking findings.

Verified the migration on a fresh local SQLite test database and exercised the gate: anonymous and unprivileged users are denied, a member of “Metrics viewers” receives 200, and an active superuser receives 200 without group membership. The combined-stack import/Compare/Composer/metrics suites pass (246 tests). Merge after #866 as documented.

PR #868, opened 2026-09-05
feat: add query layer for usage & quality metrics

benbrasso:
## Summary
Third step of #865 (usage & import-quality metrics dashboard). Adds the query functions behind all six metrics, plus a management command to validate them by hand before they're wired into an actual page.

## Depends on #866 and #867
Stacked on both prior steps — this uses the `ImportAttempt` model from #866. Until those merge, this PR's diff will show all three steps' changes together; that's expected. **Merge order: #866, then #867, then this one.**

## What's in this PR
- **`nofos/metrics.py`** — one function per metric, all built around an open-ended `months_from(start)` helper (no hardcoded date range; new months just appear as they happen):
  - `total_users_by_month` — cumulative `BloomUser` count as of each month's end.
  - `active_users_by_month` — distinct users with ≥1 `CRUDEvent` on a Nofo/Section/Subsection that month.
  - `nofos_created_by_month` — new `Nofo` rows per month.
  - `time_to_first_live_pdf_by_month` — median hours from import to first live PDF download. Mirrors the existing `count_prints.py` command's approach to parsing `CRUDEvent.changed_fields` (a plain JSON string field, not a queryable JSON column), but tracks the *first* live print per NOFO instead of the *last*, computed system-wide in one pass.
  - `import_error_rate_by_month` / `avg_warnings_by_month` — aggregates on the new `ImportAttempt` model.
- **`builder_metrics` management command** (`python manage.py builder_metrics --since 2026-09`) — prints all six as one tab-separated row per month, for manually sanity-checking the numbers once deployed.

## Out of scope for this PR
The actual metrics page/charts and PDF export — next and final steps.

## Testing
Verified by code review and `black`/Python syntax checks in the environment this was written in. No live Postgres instance was available there to actually run the command against real data — please run `python manage.py builder_metrics` after deploying this to confirm the numbers look sane before we build the page on top of it.


mikec-ai (approved):
Reviewed the query layer and management command separately from the inherited steps. No blocking findings in this step.

The existing metrics tests pass. Additional local checks verify distinct-user counting, exclusion of test prints, selection of the earliest live print rather than a later repeat, a 12-hour median from 4/12/20-hour examples, a 33.3% error rate from one failure in three attempts, and averaging warnings across successes only. These checks used a fresh SQLite test database against the combined stack. Production-data spot checks remain useful after deployment. Merge after #866 and #867.

PR #876, opened 2026-09-08
docs: document empty Word export for HRSA-27-072 WISH

benbrasso:
Documents the production report that **Export Word doc** for HRSA-27-072 WISH takes longer than usual and downloads an empty file. Includes reproduction links, expected behavior, confirmed structural evidence from both supplied downloads, and verification criteria.

Both DOCX files contain only section settings in the document body, with no paragraphs, tables, text, or drawings. Root cause is under investigation; this PR contains documentation only.

Validation: inspected both DOCX archives and document XML; `git diff --check` and commit hooks passed. No application code changed.


benbrasso:
This bug is now tracked in issue #877. Closing this documentation-only PR; investigation and any eventual fix should reference that issue.

PR #879, opened 2026-09-08
fix: prevent PDF preview from rendering the login page

### LINKED ISSUE #878: Preview PDF can render the Builder login page instead of the NOFO
## Report

Preview PDF intermittently returns a one-page PDF of the **Login — NOFO Builder** screen instead of the NOFO. The browser remains at the inline print endpoint; this is not simply the editor being logged out. Reported by Ben on September 8, 2026.

## Triage

On current main (`2d3b505c`), the authorized print POST sends a `document_url` to DocRaptor. DocRaptor then fetches the HTML without the editor's session. Builder's middleware permits that fetch only over HTTPS with an exact allowlisted `X-Forwarded-For` value.

Local reproduction of the real middleware confirms:

- Exact configured IP: HTML request is allowed.
- Missing/unlisted IP or a comma-separated forwarded chain: redirect to Builder login.

That redirect can be rendered as a successful PDF. This establishes a failure mechanism consistent with the report; **the exact production request/header that caused Ben's instance has not been verified**. No evidence currently points to a browser cache problem.

## Proposed fix and acceptance criteria

- Render NOFO HTML within the already-authorized print request and pass `document_content` to DocRaptor instead of requiring a separate anonymous HTML fetch.
- Set the document's absolute base URL so stylesheets, images, and relative links still resolve.
- Do not send session credentials, CSRF tokens, or account-specific controls in the HTML.
- Preserve authentication/group restrictions, preview/download behavior, test/live options, PDF/UA profile, error handling, and audit events.
- Add regression coverage and local browser evidence using a synthetic NOFO and a simulated renderer boundary. Keep a live DocRaptor/dev smoke test explicit as a separate verification step.

Do not fix this by broadening anonymous access or trusting arbitrary forwarded IPs.

mikec-ai:
## Summary

Fixes #878.

Preview PDF can successfully return a PDF of Builder's login page when DocRaptor's anonymous HTML fetch does not pass the IP allowlist. This renders the authorized NOFO HTML inside the print request and sends it directly to DocRaptor, removing that extra login-sensitive fetch.

- Reuse the existing document template and context, with a PDF-only child template that omits navigation and forms.
- Pass the absolute document base URL for stylesheets, images, and relative links.
- Exclude request/session/CSRF context; disable PDF-client debug payload logging.
- Preserve access restrictions, preview/download modes, test/live selection, print options, audit events, and error handling.

## Verification

- 1,909 Django tests passed; 11 JavaScript tests passed.
- New regression tests cover document payload/content, metadata, theme/cover, absence of request secrets, PDF options, permissions, and the renderer IP/login redirect boundary.
- Local real-browser reproduction: Preview PDF on baseline yields a login-page PDF using a simulated renderer; after the fix it yields the synthetic NOFO. Download also works and anonymous HTML access still redirects to login.
- Real DocRaptor test with synthetic HTML and public test key returned the correct five-page PDF with public Builder assets; all pages visually inspected.
- Independent Astra-low review completed with no actionable findings.

See [verification notes and before/after screenshots](documentation/verification/pdf-preview-878/README.md).

## Limits / rollout

The exact production request/IP/header behind Ben's report is not verified. The real-engine test used synthetic content, not a deployed dev build or private NOFO. Please smoke-test a representative NOFO in dev before production rollout. No authentication bypass or IP allowlist change is included.


mikec-ai:
@benbrasso I reproduced a concrete path to the login-page PDF and put the fix through separate Astra-low diagnosis, implementation, and independent review passes.

The fix sends the authorized NOFO HTML directly to DocRaptor instead of relying on a second anonymous HTML fetch. No auth or IP-allowlist relaxation.

**Verified:** 1,909 Django tests + 11 JavaScript tests; local browser Preview PDF and Download PDF; unchanged anonymous-access denial; and a real DocRaptor/Prince test using only synthetic content and the public test key. Independent review found no actionable issues.

**Before — local browser reproduction with a simulated renderer:**

Locally reproduced PDF of Builder login

**After — real DocRaptor output from the fixed synthetic payload:**

Synthetic NOFO cover rendered by DocRaptor

Authorized NOFO content rendered by DocRaptor

The watermarks, blank deadline, and placeholder cover are expected in this synthetic test. The original private Slack screenshot is not republished here. The precise production request/header remains unverified, and a representative dev smoke test remains before production rollout.

benbrasso:
What we lose

  1. A single, guaranteed-WYSIWYG rendering path. Before, DocRaptor fetched the exact same URL a browser
     would hit — full dispatch(), full context processors, byte-for-byte what's on screen. Now there
     are two code paths that render "the NOFO page": the real NofosDetailView.get() (used on screen) and
     this new hand-assembled path (NofosDetailView().get_context_data() + a manually-built context).
     They can drift. Concretely: Django's configured context processors are debug, request, auth
     (user/perms), messages, and template_context. The new code only replicates template_context by
     calling it directly. If someone later adds a context processor, or changes something
     NofosDetailView.dispatch() does before rendering, the on-screen page picks it up automatically —
     the PDF path won't, silently (Django templates fail silently on undefined vars, not with an error).
     - I checked whether this bites today: user is referenced in
       warning_successor.html/warning_archived.html for an is_superuser gate. It's now undefined instead
       of AnonymousUser, but both are falsy, so output is unchanged today. It's a latent risk, not a
       current bug.
  2. DocRaptor's own debug/request visibility. debug was flipped True → False specifically so the full
     document payload doesn't get logged. That's the right call, but it does mean less visibility into
     exactly what DocRaptor received if something renders wrong in production — you can't just re-click
     a URL to reproduce what they saw.
  3. No size ceiling handled by someone else. document_url let DocRaptor pull as much HTML as it wanted
     with no size constraint on our side. document_content puts the whole rendered page in the POST body
     Django sends to DocRaptor's API. For a very large NOFO (lots of sections/tables), this is a bigger
     request payload and more synchronous render time inside the request/response cycle of the print
     POST itself, versus that render work happening in a separate request before. Worth a quick check on
     DocRaptor's request-size limits and on whether the biggest real NOFOs render/post noticeably
     slower now.

  Why it probably wasn't built this way originally

  document_url is the simpler, more common DocRaptor integration pattern — point it at a public(ish) URL
  and let their infrastructure fetch it, no manual template assembly needed on your end. The catch
  nobody fully worked through at the time: DocRaptor's fetch is server-to-server, so it never carries
  the user's session cookie — it's always anonymous, regardless of who clicked print. The IP-allowlist
  bypass in NofosLoginRequiredMiddleware was the fix for that gap (let DocRaptor's known IP in without a
  login). It mostly worked, until the specific request in #878 didn't match the allowlist and DocRaptor
  got the login page instead. The document_content approach avoids needing that bypass at all, but
  requires the app to do its own template assembly — more upfront work, which is likely why it wasn't
  the first design.

  Unintended-consequence watch-list

  - The two-rendering-path drift risk above (biggest one, low-probability but easy to introduce
    accidentally in a future PR).
  - Confirm DocRaptor request-size/timeout limits against your largest real NOFOs.
  - The IP-allowlist bypass in middleware.py is untouched and, per the grep above, may now have zero 
    remaining callers — it existed only to let DocRaptor's anonymous fetch through, and nothing fetches
    anonymously anymore. That's not a regression from this PR, but it's now unused attack surface (a
    spoofable-header-based anonymous-access door on /nofos/{uuid}) worth a deliberate decision: keep it
    for a reason, or remove it.

  Stub comment for Mike

  ▎ Nice fix — sending the content directly instead of relying on the IP-allowlisted fetch is the right 
  ▎ call. Two follow-up thoughts, not blockers:
  ▎
  ▎ 1. I grepped the repo and this is the only place that talks to DocRaptor, so there's nothing else 
  ▎ that needs the same document_url → document_content change today. Just flagging in case another 
  ▎ export path gets added later that reaches for the old pattern.
  ▎ 2. Now that print/preview no longer needs DocRaptor to fetch /nofos/{uuid} anonymously, is the 
  ▎ IP-allowlist bypass in NofosLoginRequiredMiddleware (config.DOCRAPTOR_IPS / match_view_url) still 
  ▎ used for anything? If not, seems worth a follow-up to remove it — it's a header-based 
  ▎ (X-Forwarded-For) anonymous-access exception on the NOFO detail view, and this PR looks like it 
  ▎ removes its last legitimate reason to exist.
  ▎
  ▎ Separately: document_context is now built by hand (NofosDetailView().get_context_data() + a direct 
  ▎ template_context(None) call) instead of going through the normal request/context-processor pipeline.
  ▎ That's deliberate and correct for stripping session/CSRF/user data, but it does mean the 
  ▎ PDF-rendering path and the on-screen page are now two separate code paths that can drift if a 
  ▎ context processor or NofosDetailView changes later. Might be worth a comment there flagging that 
  ▎ risk for future maintainers.

benbrasso (approved):
Good, but let's have Derek review

dfitchett (commented):
Did we want to commit the documents in `/documentation`? It seems like those are just to support this implementation for this PR?

nofos/nofos/views.py
```diff
@@ -2284,15 +2287,30 @@ def post(self, request, pk):
...
+        # does not read its request argument. Preserve the base-page metadata.
+        document_context.update(template_context(None))
+        document_content = render_to_string("nofos/nofo_pdf.html", document_context)
+
         try:
             response = doc_api.create_doc(
                 {
                     "test": is_test_pdf,  # test documents are free but watermarked
-                    "document_url": nofo_url,
+                    "document_content": document_content,
```
dfitchett, line 2310: Oh this is sooo much better! 🥳

nofos/nofos/views.py
```diff
@@ -2284,15 +2287,30 @@ def post(self, request, pk):
...
                     "document_type": "pdf",
                     "javascript": False,
                     "pipeline": 11,
                     "prince_options": {
+                        "baseurl": nofo_url,  # resolve relative assets and links
```
dfitchett, line 2315: I'd be curious to see this option still allows the assets and links to render correctly... 🤔

dfitchett (commented):
This all looks good so far! I think it's worth a deploy to dev to verify the assets are fetched correctly still.

mikec-ai:
@dfitchett Addressed the documentation question in b1a60e99: removed the PR-only verification files from the branch. The existing comment links remain pinned to the earlier commit, preserving the evidence. Also added a maintenance note for the document-only rendering context. All 18 focused PDF tests pass. Dev deployment of this exact commit is running: https://github.com/HHS/simpler-grants-gov/actions/runs/34479368871 . I will verify preview/download and asset rendering before calling the dev check complete.

mikec-ai:
@dfitchett @benbrasso Update: the updated PR passes its unit tests, and the dev deployment passed lint/format/tests. The Builder Anchore scan now fails on Python 3.14.7 findings CVE-2026-17084 and CVE-2026-15806 (also reports low-severity CVE-2026-15310). These same findings are already being addressed in https://github.com/HHS/simpler-grants-gov/pull/12353 . Builder mirrors the Grype exceptions, so it will need the corresponding update once that decision is approved. I have not bypassed the scan or merged this PR. Authenticated dev browser access is ready, but the new commit has not yet deployed, so the requested dev PDF/asset verification remains pending.

dfitchett (approved):

dfitchett:
We can probably wait until the other PR is merged, and try to redeploy.

mikec-ai:
@dfitchett @benbrasso Dev verification is complete for 145c7d234fc55121a58276aa6d02c780e532648f. Synced the approved scan exceptions and artifact handoff from main; all PR checks pass, and the dev deployment passed tests, all scans, migrations, and rollout: https://github.com/HHS/simpler-grants-gov/actions/runs/34598528723 .

Using the existing Release 3.34.0 smoke-test NOFO, the authenticated Preview PDF rendered the document rather than the login page. The separate Download PDF action produced a six-page PDF. I rendered and inspected all six pages: NIH/HHS logos, section icons, embedded fonts, colors, page layout, and document content were present. The missing-Step-2 placeholders are existing synthetic-fixture content, not a new rendering failure. No NOFO content was edited.

All 1,909 local tests also passed. This completes the requested dev PDF/asset check; production was not deployed and no security check was bypassed.

Issue #880, opened 2026-09-09
Evaluate Word export options and establish reliable, documented operations


benbrasso:
## Problem

NOFO Builder needs a reliable, maintainable Word-export solution for continued use in the upcoming fiscal year, including the HHS Department Governance clearance-export workflow introduced in [[PR #821](https://github.com/HHS/simpler-grants-pdf-builder/pull/821)](https://github.com/HHS/simpler-grants-pdf-builder/pull/821).

The current GrabzIt integration has two operational problems:

- Its account and billing remain personally managed by a former team member.
- Development and production share account-level cookie state. Paul reported that exporting in development causes subsequent production exports to return blank Word documents until someone manually clears the stored cookies.

The current code supplies the user’s Builder session and CSRF cookies to GrabzIt, then asks it to fetch a protected export page. Paul described the blank-document failure as a login redirect: the returned login page lacks the expected document selector, so conversion produces an empty DOCX. The exact cause of GrabzIt choosing the wrong cookie remains unverified.

Before investing further in this integration, reassess whether GrabzIt is still the best fit. Another SaaS service, a self-hosted library, or a custom implementation may now meet our needs more effectively.

### Original conversion evaluation

Read the [[HTML-to-DOCX solution ADR](https://github.com/HHS/simpler-grants-pdf-builder/blob/main/documentation/adr/2026-01-12-html-to-docx-solution.md)](https://github.com/HHS/simpler-grants-pdf-builder/blob/main/documentation/adr/2026-01-12-html-to-docx-solution.md) before evaluating alternatives.

The January 2026 evaluation compared ConvertHub, ConvertAPI, GroupDocs, CloudConvert, and GrabzIt. GrabzIt performed best on formatting preservation and export/re-import behavior. The ADR also records limitations involving semantic Word styles, CSS classes, and footnotes.

Use these findings as a baseline and identify what has changed: available products, library capabilities, requirements, pricing, or implementation approaches.

### Clearance export and migration risk

Review [[PR #821 — Add feature-flagged Word export with HHS Department Governance policy-language tagging](https://github.com/HHS/simpler-grants-pdf-builder/pull/821)](https://github.com/HHS/simpler-grants-pdf-builder/pull/821) and subsequent changes to that workflow.

This work introduced a second Word-export path for OMB/Departmental clearance review, controlled by `HHS_NOFO_POLICY_EXPORT_ENABLED`. It identifies standard HHS Department Governance language against canonical text and:

[...]

PR #883, opened 2026-09-10
fix: use TLS for Word conversion provider requests

mikec-ai:
## What changed

- Explicitly enables TLS on the GrabzIt client before setting cookies or starting a Word conversion.
- Adds regression coverage for the provider call order.
- Verifies the pinned SDK selects an HTTPS connection without making a live provider request.

## Why

The Builder export page already requires HTTPS, but the conversion provider client also needs to be configured explicitly so its own API operations use TLS.

This is intentionally separate from the broader Word export investigation in #880. It does not change the current cookie-forwarding design or resolve the remaining operational questions in that spike.

## Validation

- 1,903 tests pass.
- Black and isort checks pass.
- No provider credentials or live conversion requests were used during verification.


dfitchett (approved):
Looks right to me! Nice test coverage!

benbrasso:
@mikec-ai this one's ready to merge again. I'll let you merge: I'm losing track a bit of our various workstreams related to exporting.

PR #889, opened 2026-09-10
fix: mirror upstream grype ignores for new Python stdlib CVEs

benbrasso:
## Summary

Anchore Scan has been failing on PRs here with no code change involved. Upstream HHS/simpler-grants-gov found three new CVEs in the base Python image, none of which are fixable in any GA Python release yet (fixed only in the Python 3.15 pre-release):

- CVE-2026-17084 — idna/stringprep codec
- CVE-2026-15806 — urllib.request.HTTPPasswordMgr
- CVE-2026-15310 — zipfile decompression

These were already ignored in HHS/simpler-grants-gov's own .grype.yml via HHS/simpler-grants-gov#12353 (tracked under HHS/simpler-grants-gov#8109), and this repo's .grype.yml explicitly says to keep the two files in sync. This PR catches it up.

## What changed

- `.grype.yml`: added the three CVE ignore entries above, matching the existing comment/format style.

## Test plan

- [x] Diff matches the equivalent entries in HHS/simpler-grants-gov's `.grype.yml`
- [ ] Anchore Scan check passes on this PR


mikec-ai (approved):
Looks good. The three added CVE exceptions match the approved and merged upstream change in HHS/simpler-grants-gov#12353. This keeps the Builder scan configuration aligned without disabling scanning or changing the failure threshold. All six CI checks pass, including Anchore.

Non-blocking follow-up: these exceptions suppress the findings rather than fix the vulnerabilities. Since the referenced tracking issue (#8109 in simpler-grants-gov) is closed, we should make sure someone tracks removing these exceptions when a supported Python release includes the fixes.

PR #890, opened 2026-09-11
fix: pass the built Docker image via artifact, not cache, in code scans

benbrasso:
## Summary

Anchore Scan (and, less often, Trivy/Dockle) has been intermittently failing with "docker load: /tmp/docker-image.tar: No such file or directory" — 9-second job failures with no real scan output, e.g. https://github.com/HHS/simpler-grants-pdf-builder/actions/runs/34479361265.

Root cause: the "Restore cached Docker image" step reports success even on a cache miss — it just silently restores nothing. Build Image saves the image under actions/cache keyed on `nofos-image-${{ github.sha }}-${{ github.run_id }}`, and the three scan jobs (Trivy/Anchore/Dockle) restore it independently in the same run. That key is unique, but the cache entry isn't guaranteed to survive: GitHub enforces a 10GB-per-repo cache limit, and unrelated cache writes elsewhere in the repo (other concurrent runs' image caches, poetry deps, trivy DBs) can evict this entry before a sibling job gets to it.

## What changed

- `.github/workflows/code_scans.yml`: replaced the Docker image handoff between the `build` job and the three scan jobs with `actions/upload-artifact` / `actions/download-artifact` instead of `actions/cache`. Artifacts are scoped to and guaranteed to last for the run — the right primitive for same-run job-to-job data, unlike cache (which is a best-effort, evictable, cross-run store).
- The Trivy DB cache (a genuine cross-run cache) is unchanged.

## Test plan

- [ ] Code Scans workflow (Build Image → Trivy/Anchore/Dockle) passes on this PR
- [ ] Re-run a few times to confirm the previous intermittent failure no longer reproduces


mikec-ai (approved):
Reviewed fa6f5e91. No blocking findings in the Docker image artifact handoff: producer/consumer names and paths align, each scan depends on the build, and the cross-run Trivy database cache remains unchanged. All three scan jobs passed during review. Workflow reruns were not exercised; require the remaining CI checks to pass before merge.

PR #898, opened 2026-09-14
fix: stabilize container security scans

benbrasso:
## Purpose

Keep container security scans reliable and green for all PRs by fixing the two current baseline failures:

- upgrade gzip, libpcre2-8-0, and libsqlite3-0 to Debian's available security-fixed versions instead of ignoring fixable CVEs
- replace erzz/dockle-action, whose runtime release download repeatedly returned a 92-byte error response, with Dockle's official version-pinned container action
- allow only the known non-secret runtime setting names Dockle flags, while preserving the existing warning threshold and accepted-file list
- avoid a secondary summary-step failure when no Dockle report exists

The three unfixed Python CVEs from HHS/simpler-grants-gov#12353 are already covered in this repository by #889.

## Validation

- Dockle scan passes in this PR after switching to the official action and allowing the known runtime setting names
- poetry run pre-commit run --files Dockerfile .github/workflows/code_scans.yml
- git diff --check


benbrasso:
@mikec-ai give me a quick gut check on this pr, thanks

PR #921, opened 2026-09-17
fix: exclude administrative metadata from readability metrics

mikec-ai:
## Summary
- Omit PDF author, subject and keywords only from metrics rendering; preserve normal Word export.
- Version the Builder measurement input independently from the metrics package/profile.
- Preserve historical snapshots while recalculating unchanged revisions under the corrected input contract.

## Validation
- 34 focused readability tests passed, including metadata invariance and historical snapshot/cache behavior.
- Migration forward/backward smoke verified with legacy data; v1/v2 coexistence and duplicate-v2 rejection verified.
- git diff --check passed.

## Scope and rollout
No formula, target, PDF extraction, or package dependency changes. Apply migration 0142 with the code. New calculations use word-export-v2; existing results remain historical. Once v1/v2 scores coexist, retain the forward schema during rollback rather than restoring the older uniqueness constraint, which cannot represent both records.


mikec-ai:
Review completed against current main: metadata omission is restricted to the metrics rendering flag; normal Word exports retain the fields. Snapshot identity includes an explicit Builder input contract so legacy measurements remain historical and cannot satisfy the corrected cache. Regression coverage includes unchanged metrics after metadata edits, export preservation, legacy/new snapshot coexistence and migration preservation. No formula/target/package changes. No blocking findings in this scoped change. Awaiting CI before merge. Rollback caveat is documented in the PR description.

PR #939, opened 2026-09-21
fix: suggest Word style repair for trailing heading outliers

### LINKED ISSUE #938: Make ambiguous-heading import errors recommend the likely Word fix
## Problem

A recent user could not import a Word document because one main-section heading used a different Word heading level from all the others.

In the reproduced document, Builder sees:

- seven main sections as Heading 2 (`Step 1` through `Step 6`, plus `Contacts and Support`)
- one final section, `Endnotes`, as Heading 1

The existing `IMPORT-AMBIGUOUS-HEADINGS` path correctly blocks the import and prevents silent content loss. After #913 / #914, the error page also shows the first Heading 2 and the first Heading 1 and tells the user to apply the same heading level to every main section.

That is accurate, but the user still has to infer the concrete repair. In this case the useful instruction is: **in Word, select `Endnotes` and change its style from Heading 1 to Heading 2, then save and import again.**

## Proposed improvement

When Builder can identify a likely isolated heading-level outlier with high confidence, make the ambiguous-headings error name the outlier and recommend the likely Word style change.

For this case, the page could say something like:

> `Endnotes` uses Heading 1, while the other seven main sections use Heading 2. In Word, select `Endnotes` and apply the Heading 2 style. Save the document, then import it again.

Keep the current neutral guidance for genuinely ambiguous documents where Builder cannot safely infer which level is intended.

## Investigation notes

- `resolve_section_heading_level()` currently considers non-empty, non-table `h1` and `h2` elements.
- It blocks when an `h2` appears before the first `h1`, carrying the first heading at each level into `AmbiguousHeadingHierarchyError`.
- It does not currently count headings or identify a minority/outlier level.
- Import preprocessing removes the known redundant `Before you begin` section before resolving the section level.
- There is no general rule that ignores an initial Heading 1 used as a document title. Any recommendation heuristic should therefore be conservative about leading Heading 1 content and should fall back to the current generic guidance when intent is unclear.

## Documentation requirement

When this work is implemented, update `documentation/IMPORT_ERROR_CODES.md` so the `IMPORT-AMBIGUOUS-HEADINGS` entry accurately describes:

- when Builder gives a specific outlier recommendation versus neutral guidance
- the message and recovery steps the user sees
- what support should tell someone who encounters either form of the error

Keep the documentation aligned with the user-facing catalog entry in `nofos/bloom_nofos/import_errors.py`. If the heading-resolution behavior changes, also update the matching rule in `documentation/IMPORT_RULES.md` in the same change, as required by the repository's import-rule policy.

## Acceptance criteria

- [ ] When one Heading 1 appears after multiple Heading 2 main sections, and Builder can confidently treat it as the outlier, the error page names that heading and tells the user to change it to Heading 2 in Word.
- [ ] Equivalent high-confidence cases with one Heading 2 among multiple Heading 1 main sections receive the corresponding recommendation, if supported by the resolver's valid-document rules.
- [ ] The message includes concrete Word actions: select the named heading, apply the recommended Heading style, save, and import again.
- [ ] Ambiguous cases keep neutral guidance and do not guess which heading level is correct.
- [ ] Tests cover the reproduced seven-Heading-2/one-trailing-Heading-1 case and a case where a leading Heading 1 could be a title.
- [ ] `documentation/IMPORT_ERROR_CODES.md` is updated to match the implemented trigger, user-facing message, recovery steps, and support guidance.
- [ ] `nofos/bloom_nofos/import_errors.py` remains aligned with the documentation, and `documentation/IMPORT_RULES.md` is updated if heading-resolution behavior changes.

## Related work

- #745 stopped mixed H1/H2 documents from silently importing only part of their content.
- #913 / #914 added specific, documented import-error copy and surfaced both conflicting headings on the error page.

## Out of scope

- Automatically changing the uploaded Word document.
- Special-casing the word `Endnotes`; the improvement should work for any high-confidence heading-level outlier.

mikec-ai:
## What changed

Previously, the mixed-heading import error named two headings but left the user to work out which Word style to change.

Now, when multiple Heading 2 headings are followed by a single final Heading 1, the error adds a **Likely Word fix**. It names that heading and suggests applying Heading 2 if these are all main sections. Other mixed structures retain neutral guidance.

This does not change which documents import successfully, automatically restyle a document, or special-case Endnotes. Valid Heading 1 sections with Heading 2 subsections are unchanged.

Closes #938.

## Before and after

Local previews rendered using Builder's real Django templates and styles, using the same seven-Heading-2 / final-Endnotes example. Not a deployed-environment smoke test. [Capture details](documentation/review-evidence/938/README.md).

| Before | After |
| --- | --- |
| Before: neutral heading guidance | After: specific likely Word fix |

## Verification

- 699 tests pass across `nofos.test_nofo`, `nofos.tests_nofos.test_import`, and `nofos.tests_nofos.test_import_error_catalog`.
- Covers seven Heading 2s followed by Endnotes or another title, uncertain mixed hierarchies, leading Heading 1 titles, ignored blank/table headings, safe escaping, and blocked imports saving no records.
- Independent agent review found no blocking correctness or security issues. Addressed its documentation request and added its suggested ignored-heading count test.
- Updated import-rule and error-code documentation.


benbrasso (approved):
Awesome

